PR #7, opened 2019-04-10
lvm-batch osd scenarios

shreekarSS:
ceph/ceph.py
     - updated multiple_lvm_scenarios() with 'batch' type in the retuning dict, updated dict 'devices_dict' to update the devices of node dynamically
     - moved osd_scenario_list list to lvm_utils utility
     - added lvm batch osd scenario which needed 'devices_prefix' as 'devices'
     - added a 'counter' variable which returns different scenario for each OSD nodes
     - updated check_health() with replacing self.ceph_demon_stat['osd'] to 'up_osds'
utility/lvm_utils.py
    - Added 2 new osd scenario with lvm-batch
    - Updated return type of all osd_scenario functions to a dict
conf/luminous/sanity-ansible-lvm.yaml
    - Added 2 new OSDs for lvm-batch scenarios

### DIFF
--- a/ceph/ceph.py
+++ b/ceph/ceph.py
@@ -200,13 +200,7 @@ class Ceph(object):
         nfs_hosts = []
         client_hosts = []
         iscsi_gw_hosts = []
-        osd_scenario_list = [
-            lvm_utils.osd_scenario1,
-            lvm_utils.osd_scenario1_dmcrypt,
-            lvm_utils.osd_scenario2,
-            lvm_utils.osd_scenario2_dmcrypt,
-            lvm_utils.osd_scenario3,
-            lvm_utils.osd_scenario3_dmcrypt]
+        counter = 0
 
         for node in self:  # type: CephNode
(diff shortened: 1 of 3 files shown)


ceph/ceph.py
```diff
@@ -550,8 +547,8 @@ def check_health(self, client=None, timeout=300):
         all_osds = int(match.group(1))
         up_osds = int(match.group(2))
         in_osds = int(match.group(3))
-        if self.ceph_demon_stat['osd'] != all_osds:
-            logger.error("Not all osd's are up. Actual: %s / Expected: %s" % (all_osds, self.ceph_demon_stat['osd']))
+        if up_osds != all_osds:
```
vasukulkarni: Is it ok for every suite?

shreekarSS: Yup, it will compare with parsed output of ceph status command, instead of OSD count from ceph_demon_stat.

PR #11, opened 2019-05-27
Added beta repos for 7.7 beta images

VasishtaShastry:


rakeshgm:
@VasishtaShastry I don't think we need beta repos into master. When RHEL 7.7 releases the dependencies will be merged. These repos are not needed for other RHEL versions. let's wait for the release and check. Till then let this change be in the rhel-7.7 branch

PR #30, opened 2019-09-19
update script location

XDsubhash:
* updated rgw script location from gitlab to github
* removed obsolete script


tests/sanity_rgw.py
```diff
@@ -18,7 +18,7 @@ def run(ceph_cluster, **kw):
     config = kw.get('config')
     log.info("Running rgw tests %s" % config.get('test-version', 'v2'))
     rgw_ceph_object = ceph_cluster.get_ceph_object('rgw')
-    git_url = 'http://gitlab.cee.redhat.com/ceph/ceph-qe-scripts.git'
+    git_url = 'https://github.com/red-hat-storage/cephci.git'
```
rakeshgm: This should be pointing to ceph-qe-scripts not cephci

XDsubhash: updated

PR #31, opened 2019-09-23
added new ansible container suite

XDsubhash:
* added new ansible container suite with mixed lvm configs

### DIFF
--- a/ceph/ceph.py
+++ b/ceph/ceph.py
@@ -1296,8 +1296,7 @@ class CephNode(object):
                         continue
                     self.exec_command(
                         cmd='sudo ping -I {interface} -c 3 {ceph_node}'.format(interface=eth_interface,
-                                                                               ceph_node=ceph_node.shortname),
-                        check_ec=False)
+                                                                               ceph_node=ceph_node.shortname))
                 logger.info(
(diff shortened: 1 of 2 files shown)


ceph/ceph.py
```diff
@@ -1296,8 +1296,7 @@ def search_ethernet_interface(self, ceph_node_list):
                         continue
                     self.exec_command(
                         cmd='sudo ping -I {interface} -c 3 {ceph_node}'.format(interface=eth_interface,
-                                                                               ceph_node=ceph_node.shortname),
-                        check_ec=False)
```
rakeshgm: @XDsubhash removing `check-ec=False`  might fail in RHEL-8. I added this ec=False cause of the same reason, Can you check once again

XDsubhash: its failing in container deployments, where there will be two interfaces (docker0, eth0) in the cluster,when ping over docker0 fails it still goes ahead and tries to use it as active interface(when docker0 will always be down) which causes this task to fail in config_roll_over,py TASK [ceph-validate : fail if docker0 is not active on ceph-1vp-1568814103136-node2-monmds],For rhel8 we can/have to check,as it uses podman.

XDsubhash: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1568814103136/config_roll_over_0.log

PR #47, opened 2019-11-13
Addition of SSL functional tests on both civetweb and beast

TejasC88:
test functionality of both rgw frontends with SSL.


rakeshgm:
@TejasC88 needs config changes since these options are not supported after the changes in ceph-qe-scripts.

TejasC88:
@rakeshgm I have changed the script name as per the latest changes.

PR #57, opened 2019-11-26
Added email attributes in html page.

sunilkumarn417:
Added below new attributes for email html page,
**_suite-name, test-suite-file, configuration-file, rhbuild, ceph-release-version.,_**

### DIFF
--- a/rhbuild.yaml
+++ b/rhbuild.yaml
@@ -1,0 +1,26 @@
+# Add more attributes w.r.t ceph version under specific releases.
+ceph:
+  '2':
+    name: jewel
+    composes:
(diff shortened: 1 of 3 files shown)


release-info.yaml
```diff
@@ -0,0 +1,8 @@
+# Add more attributes w.r.t ceph cluster version under specific version.
```
rakeshgm: @sunilkumarn417 the word `cluster` is not right here. these are ceph version names.

sunilkumarn417: updated code as per the suggestions, please review.

PR #65, opened 2019-12-31
added rhel 7 pkgs for rhcs4

XDsubhash:
- added repo location for rhcs4 rpms based on rhel7
- workaround for nfs deploy issue in interop suite


XDsubhash:
#66

PR #67, opened 2020-01-07
[WIP] Initial changes to move to podman

vasukulkarni:
Signed-off-by: Vasu Kulkarni <vasu@redhat.com>


rakeshgm:
@vasukulkarni if just replace `docker` to `podman` in CLIs. this will also affect Luminous which still uses docker

Issue #105, opened 2020-04-22
osd throwing assert, not coming up sometimes on rhel7.8 


XDsubhash:
https://ceph-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/rhceph-ansible-sanity-4.1-rhel7.8/39/consoleFull

https://bugzilla.redhat.com/show_bug.cgi?id=1822134#c18

XDsubhash:
probable udev/systemd race condition.

XDsubhash:
https://ceph-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/rhceph-ansible-sanity-4.1-rhel8.1/45/console

XDsubhash:
https://ceph-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/rhceph-ansible-sanity-4.1-rhel8.1/44/console

XDsubhash:
https://ceph-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/rhceph-ansible-sanity-4.1-rhel7.8/40/console

Ameenasuhani:
Setting  ceph_stable_release: nautilus is inducing this osd failure (https://github.com/red-hat-storage/cephci/blob/master/suites/nautilus/ansible/sanity_ceph_ansible.yaml#L16) upon excluding this the runs have passed

http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1614714802916/ceph_ansible_0.log
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1614716601570/ceph_ansible_0.log

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1934213

psathyan:
@Ameenasuhani do we want to keep this issue still open

psathyan:
The BZ is closed won't fix hence I don't see a point in keeping this issue open at our end.

PR #117, opened 2020-05-13
update to mixed lvm config osd scenario list

XDsubhash:
* comments out the osd scenario in #105  , as bz moved to 4.1z1 
* update tox flake8 check

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,5 +13,5 @@ commands =
 
 [flake8]
 max-line-length = 120
-ignore = E402, E741, W503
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -13,5 +13,5 @@ commands =
 
 [flake8]
 max-line-length = 120
-ignore = E402, E741, W503
+ignore = E402, E741, W503, F522
```
rakeshgm, line 16: good catch.

PR #156, opened 2020-07-14
Fixes upgrade suite

Ameenasuhani:
Upgrade suite failure is seen due to redundant ceph_test: true
Error logs: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1592492507135/ceph_ansible_install_rhcs_3.x_cdn_0.log


Ameenasuhani:
@XDsubhash According to the thread in gchat I didn't get clarity on why we should use ceph_test: true in this case. Because We don't enable  ceph_test: true for regular testing unless we need it explicitly

PR #163, opened 2020-07-16
adding the automation of BZ : 1829646 - Autotrimming of OSD map and MonDB when OSD(s) is in down state

pdhiran:

Added the test in polarion for the same.
The test cannot be run on a newly created cluster as monDB keeps in increasing for around 1 hour with new osd mappings. Hence placed the test at the end of rados suite tests so that cluster will be aged sufficiently.

Pass logs : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1594878423461/

### DIFF
--- a/tests/rados/test_bz_1829646.py
+++ b/tests/rados/test_bz_1829646.py
@@ -1,0 +1,231 @@
+import json
+import logging
+import random
+import time
+import re
+
+log = logging.getLogger(__name__)
+
+
+def run(ceph_cluster, **kw):
+    """
+    BZ https://bugzilla.redhat.com/show_bug.cgi?id=1829646 :
+
+    1. Check the mon memory usage to store the osd map in the DB and the OSD map epoch time.
+    2. Bring down ODS(s) and check the memory usage.
+    3. The OSD map should be trimmed even when the OSD is down.
+    4. The DB size should be reduced by removing the old mappings once the new mappings are added.
+    Args:
+        ceph_cluster (ceph.ceph.Ceph): ceph cluster
+
+    Note: the test cannot be run on a cluster that is just created as the DB size increases with
+          new mappings for around next hour or so.. Please run the test on cluster who's age is at least 1.5 hours.
+          By 1.5 hours, the MonDB will be updated with the mappings.
+    """
+    log.info("Running bz-1829646")
+    log.info(run.__doc__)
+    ceph_nodes = kw.get('ceph_nodes')
+    mons = []
+    osds = []
+    osd_list_dict = dict()
+    config = kw.get('config')
+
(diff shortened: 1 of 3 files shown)


pdhiran:
Addressing review comments from previous PR:

is bug present in luminous too? if so please add it to luminous sanity_rados_lvm suite too
-> The fix is not back ported to luminous. Test can be added in luminous suite once that is done.

there is function to bring down/uposd in rados_utils ..can we improve that to use it here?is it compatible for this bz?
-> The command seems to be slightly different and i'm using the same method to start/stop both OSD daemons and nodes. Will modify the method later.  That would need significant code change.

can we move this to rados_utils.if possible..looks like it'll be helpful to everyone.Even i had to use this ls-tree in one of the bz
-> moved the method to rados_utils as suggested.

tests/rados/test_bz_1829646.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    The method is used start/stop the given OSD daemon on the given node
+    :param osd_node: name of the OSD node (ceph.ceph.CephNode): ceph node
+    :param osd_number: ID of the ceph daemon to be stopped/started. If no ID is provided, all the oSD's on the node
+                        are stopped.
+    :param task: operation to be performed. either 'start' or 'stop'
+    :return: None
+    """
+
+    log.info(f"inside method to {task} the OSD(s) on the node : {osd_node.hostname}")
```
XDsubhash, line 211: from next time try to avoid "inside method" string

XDsubhash:
> Addressing review comments from previous PR:
> 
> is bug present in luminous too? if so please add it to luminous sanity_rados_lvm suite too
> -> The fix is not back ported to luminous. Test can be added in luminous suite once that is done.
the bz was found in 3.3 , we need to include in luminous too

tests/rados/test_bz_1829646.py
pdhiran: Should i change the string in the print here then? Sure, will keep it in mind and avoid that in future.

pdhiran:
> > Addressing review comments from previous PR:
> > is bug present in luminous too? if so please add it to luminous sanity_rados_lvm suite too
> > -> The fix is not back ported to luminous. Test can be added in luminous suite once that is done.
> 
> the bz was found in 3.3 , we need to include in luminous too

3.3 BZ : Bug 1798833 was in closed state with Resolution: --- → NEXTRELEASE. Hence had not included the test in luminous suite. We can add it to suite if the fix will be added to luminous.

XDsubhash:
as per dev , it wont be backported to 3.x.

XDsubhash:
@pdhiran Pl mark customer scenario checkbox in polarion and make sure polarion tc is approved and automated

PR #197, opened 2020-08-24
Fixed 'yum update metadata' issue w.r.t rhel-8 platform

sunilkumarn417:

In RHEL-8 this command fails due to below command, and as a fix disabled status verification.
However On RHEL-7, its doesn't throw error.

```
[cephuser@ceph-mgowri2-1598286215085-node8-installer ~]$ sudo yum update meta ; echo "exit status : $?"
Updating Subscription Management repositories.
Last metadata expiration check: 1:25:50 ago on Mon 24 Aug 2020 12:55:24 PM EDT.
No match for argument: meta
Error: No packages marked for upgrade.
exit status : 1
```

### DIFF
--- a/ceph/ceph.py
+++ b/ceph/ceph.py
@@ -1847,7 +1847,9 @@ class CephInstaller(CephObject):
                 raise CommandFailed(err)
 
         if kw.get('upgrade'):
-            self.exec_command(sudo=True, cmd='yum update metadata')
+            self.exec_command(sudo=True,
+                              cmd='yum update meta',
+                              check_ec=False)
             self.exec_command(sudo=True, cmd='yum update -y ansible ceph-ansible')
         else:
             self.exec_command(sudo=True, cmd='yum install -y ceph-ansible')
--- a/tests/ceph_installer/test_ceph_deploy.py
+++ b/tests/ceph_installer/test_ceph_deploy.py
@@ -74,7 +74,7 @@ def run(**kw):
             log.info("Using the cdn repo for the test")
         log.info("Updating metadata")
         if ceph.pkg_type == 'rpm':
-            ceph.exec_command(cmd='sudo yum update metadata')
+            ceph.exec_command(cmd='sudo yum update metadata', check_ec=False)
 
     ceph1.exec_command(cmd='mkdir cd')
     ceph1.exec_command(sudo=True, cmd='cd cd; yum install -y ceph-deploy')
--- a/tests/iscsi/update-kernel-iscsi.py
+++ b/tests/iscsi/update-kernel-iscsi.py
@@ -65,7 +65,7 @@ enabled=1
     kernel_repo.flush()
     o, e = client.exec_command(cmd='uname -a')
     log.info(o.read().decode())
-    client.exec_command(cmd='sudo yum update metadata')
+    client.exec_command(cmd='sudo yum update metadata', check_ec=False)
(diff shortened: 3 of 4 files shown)


ceph/ceph.py
```diff
@@ -1846,7 +1846,9 @@ def install_ceph_ansible(self, rhbuild, **kw):
                 raise CommandFailed(err)
 
         if kw.get('upgrade'):
-            self.exec_command(sudo=True, cmd='yum update metadata')
+            self.exec_command(sudo=True,
+                              cmd='yum update metadata',
```
vasukulkarni, line 1850: actually it is meta instead of metadata, for some reason it stayed on for so long :)

rakeshgm: We had a same problem in teuthology, we made added `metadata`. https://github.com/ceph/teuthology/blob/master/teuthology/task/internal/redhat.py#L101

rakeshgm: this too. https://github.com/red-hat-storage/cephci/blob/master/tests/misc_env/install_prereq.py#L75

sunilkumarn417: @vasukulkarni  @rakeshgm Just to be clear, is it meta or metadata?

rakeshgm: I would suggest to skip running this command if its rhel-8. both `meta` and `metadata` fails.

rakeshgm:
```sudo yum udpate meta 
Updating Subscription Management repositories.
No such command: udpate. Please use /bin/yum --help
It could be a YUM plugin command, try: "yum install 'dnf-command(udpate)'"
```

rakeshgm:
```sudo yum udpate metadata 
Updating Subscription Management repositories.
No such command: udpate. Please use /bin/yum --help
It could be a YUM plugin command, try: "yum install 'dnf-command(udpate)'"```

Comment (XDsubhash) [reply ceph/ceph.py]: yes i couldnt find any man page to 'yum update meta' cmd ..pl provide if it exists

File: ceph/ceph.py [hunk 2]
```diff
@@ -1846,7 +1846,9 @@ def install_ceph_ansible(self, rhbuild, **kw):
                 raise CommandFailed(err)
 
         if kw.get('upgrade'):
-            self.exec_command(sudo=True, cmd='yum update metadata')
+            self.exec_command(sudo=True,
+                              cmd='yum update meta',
```
XDsubhash, line 1850: @sunilkumarn417  can you add ec_check=False to  all references of 'yum update metadata' in the repo or remove it completely ..afaik its deprecated functionality and a very old way to update yum cache in rhel6..we dont need to manually update metadata in rhel7 i believe ..every time we use yum it searches for latest repos added and picks latest pkgs

XDsubhash: https://stackoverflow.com/questions/13328283/equivalent-in-yum-of-apt-get-update

ktdreyer:
Yeah, as far as I can tell, the equivalent with `dnf` in RHEL 8 is the `--refresh` CLI option. Like `yum install something --refresh` will tell DNF to fetch entirely new metadata before running the `install` command.

sunilkumarn417:
Hope this helps us to get  updated metadata, need your thoughts and inputs.
```
  yum clean metadata
         Eliminate all of the files which yum uses to determine the remote availability of packages. Using this option will force yum to download all the metadata the next time it is run.
```

rakeshgm (approved):

PR #207, opened 2020-09-09
mita: handle missing images

ktdreyer:
If OpenStack does not have an image that matches the image name we've configured, log a human-readable error message.

This makes it easier to understand what hard-coded image we've currently set in the configuration and how to resolve the problem.

### DIFF
--- a/mita/openstack.py
+++ b/mita/openstack.py
@@ -132,7 +132,16 @@ class CephVMNode(object):
                 self.vm_size)
             return
         vm_size = available_sizes[0]
-        image = [i for i in images if i.name == self.image_name][0]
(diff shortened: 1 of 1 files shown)


rakeshgm (approved):

sunilkumarn417 (approved):

sunilkumarn417:
I see an enhancement here, where we can add a provision to take custom image name from CLI. @ktdreyer , @rakeshgm what are your suggestions, does custom image provision help us in anyway?

XDsubhash:
@sunilkumarn417 there's already a cli option `--osp-image` which can help with it.can be improved if needed as per need

PR #208, opened 2020-09-09
inventory: update RHEL 7.7 image name

ktdreyer:
PSI OpenStack no longer hosts the `RHEL-7.7-20190723.1-Server-x86_64` image. Use the latest RHEL 7.7 released image name instead.


rakeshgm (approved):
good catch @ktdreyer

PR #222, opened 2020-09-24
prereq: improve log message for missing cdn_credentials

ktdreyer:
Log the exact file where a user should configure `cdn_credentials`.

Log the impact of lacking the credentials (we will not log into the `registry.redhat.io` registry).

### DIFF
--- a/tests/misc_env/install_prereq.py
+++ b/tests/misc_env/install_prereq.py
@@ -139,7 +139,8 @@ def registry_login(ceph, distro_ver):
         docker for RHEL 7.x and podman for RHEL 8.x'''
     cdn_cred = get_cephci_config().get('cdn_credentials')
     if not cdn_cred:
-        log.warn('cdn_credentials not found')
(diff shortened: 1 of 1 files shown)


rakeshgm (approved):

tests/misc_env/install_prereq.py
```diff
@@ -139,7 +139,8 @@ def registry_login(ceph, distro_ver):
         docker for RHEL 7.x and podman for RHEL 8.x'''
     cdn_cred = get_cephci_config().get('cdn_credentials')
     if not cdn_cred:
-        log.warn('cdn_credentials not found')
+        log.warn('no cdn_credentials in ~/.cephci.yaml.'
```
rakeshgm, line 142: since `warn` is deprecated in py3, why not use `warning`

ktdreyer: Sure. I will submit that in a separate PR after we merge this, since that's a separate logical change.

rakeshgm (approved):

PR #246, opened 2020-10-22
Implement cephadm bootstrap

VasishtaShastry:
ceph octopus and corresponding rhcs release uses cephadm
for bootstrapping a cluster.
This commit has
 - New test to handle cephadm bootstrap.
 - New suite and conf file to test cephadm using test_cephadm module

# Description

# Checklist:


tests/misc_env/install_prereq.py
```diff
@@ -58,6 +58,8 @@ def install_prereq(ceph, timeout=1800, skip_subscription=False, repo=False, rhbu
         # TODO enable only python3 rpms on both rhel7 &rhel8 once all component suites(rhcs3,4) are comptatible
         if distro_ver.startswith('8'):
             rpm_all_packages = ' '.join(rpm_packages.get('py3'))
+            if str(rhbuild).startswith('5'):
+                rpm_all_packages = rpm_all_packages + 'lvm2' + 'podman'
```
ktdreyer, line 62: If the `cephadm` package requires those two packages in order to function properly, but Yum does not automatically pull those in, that is a bug in my opinion.

mykaul:
What's the latest status of this? This is critically important for us to run in CI using cephadm.
@veera-raghava-reddy  - anything we can help with this?

tests/misc_env/install_prereq.py
VasishtaShastry: @ktdreyer Upstream doc says that  
> Dependencies are installed automatically by the bootstrap process below.
We will verify whether this behavior is happening and update the script accordingly and also file RFE if required.

ktdreyer: Thanks. I verified today that the latest cephadm RPM package has "Requires: lvm2" and "Recommends: podman". Maybe we should change that to "Requires: podman" now?

PR #279, opened 2020-12-24
added cephadm module support and test suite

sunilkumarn417:
- Added cephadm support and test suite.
- cephadm shell access support to run orchestration commands.

Test results: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1608799340416/result.html

### DIFF
--- a/ceph/ceph.py
+++ b/ceph/ceph.py
@@ -17,7 +17,6 @@ import random
 from utility.utils import custom_ceph_config
 from utility import lvm_utils
 
-
 logger = logging.getLogger(__name__)
 
 
@@ -1912,6 +1911,43 @@ class CephInstaller(CephObject):
         except CommandFailed as err:
             logger.error(err.args)
         return False
+
+    def read_cephadm_gen_pub_key(self):
+        """
+        Read cephadm generated public key
+        Arg:
+            Installer node
+        Returns:
+            Public Key string
+        """
+        ceph_pub_key, _ = self.exec_command(cmd="sudo cat /etc/ceph/ceph.pub")
+        return ceph_pub_key.read().decode().strip()
+
+    def distribute_cephadm_gen_pub_key(self, ceph_cluster, nodes=None):
+        """
+        Distribute cephadm generated public key to all nodes in the list.
+        Args:
+            ceph_cluster: ceph cluster object
+            nodes: node list to add ceph public key(default: None)
+        """
+        ceph_pub_key = self.read_cephadm_gen_pub_key()
+
+        if nodes is None:
+            nodes = ceph_cluster.get_nodes()
+
+        nodes = nodes if isinstance(nodes, list) else [nodes]
+
+        for each_node in nodes:
+            each_node.exec_command(cmd="sudo install -d -m 0700 /root/.ssh")
+            keys_file = each_node.write_file(
+                sudo=True,
+                file_name='/root/.ssh/authorized_keys', file_mode='a'
+            )
+            keys_file.write(ceph_pub_key)
+            keys_file.flush()
+            each_node.exec_command(
+                cmd="sudo chmod 0600 /root/.ssh/authorized_keys"
+            )
 
 
 class CephObjectFactory(object):
--- a/ceph/ceph_admin.py
+++ b/ceph/ceph_admin.py
@@ -1,0 +1,571 @@
+"""
+Ceph Administrator
+"""
+import re
+import json
+import logging
+
+from utility.utils import get_cephci_config
+from ceph.ceph import CommandFailed
+from time import sleep
+
+logger = logging.getLogger(__name__)
+
+
+class CephAdmin:
+
+    TIMEOUT = 300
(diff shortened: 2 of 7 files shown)


ceph/ceph.py
```diff
@@ -1913,6 +1916,655 @@ def enable_ceph_mgr_restful(self):
...
+            #
+            #     for obj, cmd in radosgw_cmds.items():
+            #         try:
+            #             logger.info("creating {}...".format(obj))
+            #             self.shell(
+            #                 remote=self.installer,
+            #                 args=cmd,
+            #             )
+            #         except CommandFailed as warn:
+            #             logger.warning(warn)
```
rakeshgm, line 2335: why do we need these commented code here.

ceph/ceph.py
```diff
@@ -1913,6 +1916,655 @@ def enable_ceph_mgr_restful(self):
...
+    #     """
+    #     try:
+    #         node.exec_command(
+    #             cmd='stat {disk} && sudo dd if={disk} of=/dev/null count=1'
+    #                 ' && ! mount | grep -q {disk}'.format(disk=disk)
+    #         )
+    #         return True
+    #     except:
+    #         logger.debug("get_scratch_devices: %s is in use" % disk)
+    #     return False
```
rakeshgm, line 2377: same as above, do we need this block of commented code.

rakeshgm:
most of the code looks good to me. would request to address the above comments and we are good to go.

ceph/ceph.py
sunilkumarn417: Initially added to check the disk. we can eliminate this part currently.

ceph/ceph.py
```diff
@@ -1913,6 +1916,606 @@ def enable_ceph_mgr_restful(self):
...
+                           for i in out if i.get('daemon_type') == daemon)
+
+            count = 0
+            for _id in ids:
+                if _id in daemons:
+                    status = daemons[_id]
+                    if status == 'running':
+                        count += 1
+
+            logger.info("{up_count}/{num} {daemon}s up....".format(num=len(ids),
```
psathyan, line 2012: Avoid using formatter with logging instead use `%`... read about it [doc](https://stackoverflow.com/questions/34619790/pylint-message-logging-format-interpolation)

ceph/ceph.py
```diff
@@ -1913,6 +1916,606 @@ def enable_ceph_mgr_restful(self):
...
+                    )
+                    j = json.loads(out)
+                    if len(j['mons']) == num_mons:
+                        break
+                    sleep(sleep_time)
+                    timeout -= sleep_time
+
+        elif op == "purge":
+            # todo: Handle mon purge
+            pass
```
psathyan, line 2210: IMO, we could raise `NotImplementedError`

ceph/ceph.py
```diff
@@ -1913,6 +1916,606 @@ def enable_ceph_mgr_restful(self):
...
+                file_name='/root/.ssh/authorized_keys', file_mode='a'
+            )
+            keys_file.write(ceph_pub_key)
+            keys_file.flush()
+            each_node.exec_command(
+                cmd="sudo chmod 0600 /root/.ssh/authorized_keys"
+            )
+
+
+class CephAdmin(object):
```
psathyan, line 1957: As we support Python3 only, we could leave it as `class CephAdmin`... you could read about it at [link](https://stackoverflow.com/questions/4015417/python-class-inherits-object)

psathyan: FYI, there is nothing wrong in doing `class CephAdmin(object)`

psathyan (commented):
IMHO, we should split `CephAdmin`  from `ceph.py`. This would ease our maintenance efforts and triage could be smoother. 

We would also be able to split them into smaller modules/objects based on the role. Allowing us to implement purge & create methods separately.

ceph/ceph.py
rakeshgm: though the advantage here is interpreting the strings when pylint is being used. but we have are not validating anything against pylint. if we have to formalise it we need to format the entire code base and emit these kinda messages. there are places we have `%` and there are places we have used `format` and there are places we have `f'{ }`

psathyan: Agreed. I am also not in favor of refactoring... at some point in the future we would need to move to a standard based on consensus. @sunilkumarn417 you can chose to ignore this comment.

psathyan (approved):
Nice...+1

sunilkumarn417:
latest result: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1609861771530/?C=M;O=A

```
All test logs located here: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1609861771530

TEST NAME                        TEST DESCRIPTION                                               DURATION                                  STATUS
install ceph pre-requisites      None                                                           0:05:35.819471                              Pass
cephadm deployment               cephadm cluster deployment                                     0:11:10.117310                              Pass
cephadm shell execution          cephadm commands execution                                     0:00:15.511207                              Pass
```

rakeshgm (approved):

PR #291, opened 2021-01-08
mita: refactor get methods

ktdreyer:
This pull request simplifies and standardizes the code in the `get_image()`, `_get_flavor()` and `get_volume()` methods.

Loop and immediately return the very first matching resource, rather than
continuing through the entire list comprehension and then indexing the 0'th element.

Raise a simple `ResourceNotFound` if we can't find the resource, and let every other possible exception raise naturally without hiding behind a generic `OpenStackDriverError`.

Make `get_volume()` raise the exact name of the volume we were trying to find so that we provide more information to the user and match the behavior of the other methods.

### DIFF
--- a/mita/openstack.py
+++ b/mita/openstack.py
@@ -137,15 +137,11 @@ class CephVMNode(object):
 
     def _get_image(self) -> NodeImage:
         """Return the glance image reference."""
-        try:
-            return [
-                i for i in self.driver_v2.list_images() if i.name == self.image_name
-            ][0]
-        except IndexError:
-            raise ResourceNotFound("Image {} not found".format(self.image_name))
-        except BaseException as be:  # noqa
-            logger.error(be)
-            raise OpenStackDriverError("Encountered an unknown exception.")
+        images = self.driver_v2.list_images()
+        for image in images:
+            if image.name == self.image_name:
+                return image
+        raise ResourceNotFound("Image {} not found".format(self.image_name))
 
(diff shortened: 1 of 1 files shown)


ktdreyer:
(by the way I think we should look into whether OpenStack supports getting any of these directly by name, and if so, whether libcloud supports that. It could speed up the application if we did not have to list every single resource and filter them client-side.)

psathyan:
I am unable to find OpenStack API documentation on using params. From the debug messages of `openstack client`, I understand it is possible to filter resources using _params_ i.e. `/v2/images?name=rhel-8-released`. 

Libcloud has methods to retrieve resources using `ID` but not name. I believe we could use `self.driver.image_connection.request('/images?name={name}')`...

psathyan:
> (by the way I think we should look into whether OpenStack supports getting any of these directly by name, and if so, whether libcloud supports that. It could speed up the application if we did not have to list every single resource and filter them client-side.)

There is no uniform way to retrieve a resource by name (across flavors, images, network, volumes, etc.). Each resource seems to have their own methods. #293 makes an attempt to address to it... kindly let me know your thoughts.

ktdreyer:
I've opened https://github.com/apache/libcloud/issues/1542 to track adding server-side name filtering to libcloud.

ktdreyer:
Since @psathyan has rewritten these methods to address this with server-side filtering instead, I'm closing this. Thanks!

Issue #289, opened 2021-01-07
merge mita CephVMNode "driver" with "driver_v2"


ktdreyer:
We have `self.driver` and `self.driver_v2` in [mita/openstack.py](https://github.com/red-hat-storage/cephci/blob/master/mita/openstack.py). The internal rhos-d environment is running RHOSP 16 now, so the server environment is fairly modern. It would be great to use one single driver everywhere, so we can simplify the code.

Here are the `CephVMNode` methods that are still using the older `self.driver`:

- [x] `get_private_ip()`
- [x] `get_volume()`
- [x] `destroy_node()`
- [x] `destroy_volume()`
- [x] `destroy_node`
- [x] `attach_floating_ip()`

psathyan:
As part of the first step, moving the resources to use `driver_v2`

- #295 
- #296 
- #297 
- #298 
- #299

PR #293, opened 2021-01-09
Improvements to CephVMNode

psathyan:
# Description

This PR improves the `CephVMNode` object by the changing the below

- Moves to api_version `2.2` hence using `OpenStack_2_NodeDriver` i.e. #289 
- Server side filtering of the resource instead of client side.
- Removes unused Exceptions
- Method to collect `CIDR` is modified and no longer uses subnets
- Network select process is now modified

VM and ansible run [logs](http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/issue_289/)

### DIFF
--- a/mita/openstack.py
+++ b/mita/openstack.py
@@ -152,7 +152,7 @@ class CephVMNode(object):
         Return True if the given network has an IPv4 Address to lease.
 
         Arguments:
-            net_id: String, The network UUID
+            net: String, The network UUID
 
         Returns:
             bool - True on success else False
@@ -284,7 +284,7 @@ class CephVMNode(object):
         """Create and attach the volumes."""
         for item in range(0, self.no_of_volumes):
             logger.info(
-                "Creating %gb of storage for %s", self.size_of_disk, self.node_name
+                "Creating %Gib of storage for %s", self.size_of_disk, self.node_name
             )
             try:
                 _vol = self.driver.create_volume(
@@ -306,7 +306,7 @@ class CephVMNode(object):
             logger.info("Successfully attached %s to %s", _vol.name, self.node_name)
 
     def _wait_until_volume_available(self, volume) -> bool:
-        """Wait until a StorageVolume's state is "available"."""
+        """Wait until a StorageVolume's is in available state."""
         tries = 0
         while True:
             sleep(3)
@@ -327,12 +327,12 @@ class CephVMNode(object):
         return False
 
     def get_private_ip(self) -> str:
-        """Retrieve the Private IP address"""
+        """Return the first private IP address."""
         _node = self.driver.ex_get_node_details(self.node)
         return _node.private_ips[0] if _node.private_ips else ""
 
     def get_volume(self, name) -> StorageVolume:
-        """Retrieve a StorageVolume instance using the provided name."""
+        """Return the StorageVolume instance for the provided name."""
         url = f"/volumes?name={name}"
         _object = self.driver.volumev2_connection.request(url).object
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -13,7 +13,7 @@
     author='Vasu Kulkarni',
     author_email='vasu@redhat.com',
     install_requires=[
-        'apache-libcloud',
+        'apache-libcloud==3.3.0',
```
ktdreyer, line 16: I'm thinking we should avoid pinning to an exact version, and we should require this as a minimal version instead (`libcloud>=3.3.0`). It will be easy to forget about this pinned version number and then we'll be stuck on an old version.

mita/openstack.py
```diff
@@ -220,7 +220,7 @@ def _get_network(self, network: Optional[str] = None) -> OpenStackNetwork:
                 logger.debug(be)
                 continue
 
-        raise ResourceNotFound(f"No suitable network resource found.")
+        raise ResourceNotFound("No suitable network resource found.")
```
ktdreyer, line 221: One commit edits this `ResourceNotFound` string to be an f-string, and then another commit in this same PR removes the `f`. Can you rebase your branch and edit the original commit so that no commits change this line? Then it will be easier to review each commit individually.

mita/openstack.py
```diff
@@ -111,80 +100,84 @@ def get_driver(self, **kw):
...
-            network: String, The network UUID
+            net: Reference to OpenStackNetwork object
 
         Returns:
             bool - True on success else False
 
         Note: This method returns True only if the given network has more than 3
         """
         try:
-            resp = self.driver_v2.network_connection.request(
```
ktdreyer, line 173: Heh, I did not realize we were already making some direct HTTP requests now, beyond what libcloud supports.

ktdreyer (commented):
These are great improvements. Thanks.

There are many changes here that are difficult for me to review all together. For example, some early commit make changes that other later commits revert.

Could you break this up into individual PRs so it is easier for me to understand exactly what is changing?

mita/openstack.py
```diff
@@ -111,80 +100,84 @@ def get_driver(self, **kw):
...
-        Get Apache Libcloud OpenStack driver for nova version 1.1.
+    def _get_image_by_name(self) -> NodeImage:
+        """Return the glance image reference."""
+        logger.info("Gathering %s image details.", self.image_name)
 
-        https://libcloud.readthedocs.io/en/latest/compute/drivers/
-        openstack.html#compute-1-0-api-version-older-installations
+        url = f"/v2/images?name={self.image_name}"
+        _object = self.driver.image_connection.request(url).object
+        images = self.driver._to_images(_object, ex_only_active=False)
```
ktdreyer, line 109: `_to_images()` -> `_to_image()` for simplicity. We can also drop `ex_only_active` as well.

ktdreyer: Whoops, I'm sorry, you're right this returns multiple images!

ktdreyer: But we do want to filter by active images, right? `ex_only_active=True`?

setup.py
psathyan: Exact pinning allows us to be reliable by preventing a fatal change breaking the existing implementation. I agree that it wouldn't break in new changes. Other than this, I don't see a reason why we shouldn't pick the latest versions. Let me go ahead and change to `>=`

mita/openstack.py
psathyan: I actually swept it under the rug... initially, the argument to the method was the network id hence didn't have the resource name. Subsequent changes forced me to get the Network object... sorry, I should have made it easier.

mita/openstack.py
psathyan: It should be `True`... it is a collateral from my experiments. Thanks... will fix it

psathyan:
Let me break this into smaller PR's... closing it for now.

mita/openstack.py
psathyan: Actually... it is a problem with libcloud... they look for upper case instead of lower case which prevents us from picking up the correct image.

PR #295, opened 2021-01-12
Improvements to get_image method

psathyan:

# Description

We form a list of suitable OpenStack image after retrieving the list of images from the server and then traversing it. In this PR, the identification of suitable images is modified to use the `name` param with the request to get a smaller suitable list of matched image resources.

In addition to the above change, the request call is made using `OpenStack_2_NodeDriver`. This requires us to move to `3.3.0` version of `apache-libcloud`. It is required to fix a problem with image retrieval. 

[Logs](http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/openstack/image_filter.log)

### DIFF
--- a/mita/openstack.py
+++ b/mita/openstack.py
@@ -135,17 +135,29 @@ class CephVMNode(object):
         """
         return self.driver_v2 if self.driver_v2 else self.get_driver(api_version="2.2")
 
-    def _get_image(self) -> NodeImage:
-        """Return the glance image reference."""
-        try:
-            return [
-                i for i in self.driver_v2.list_images() if i.name == self.image_name
-            ][0]
-        except IndexError:
-            raise ResourceNotFound("Image {} not found".format(self.image_name))
-        except BaseException as be:  # noqa
-            logger.error(be)
-            raise OpenStackDriverError("Encountered an unknown exception.")
+    def _get_image(self, name: Optional[str] = None) -> NodeImage:
+        """
+        Return a NodeImage instance using the provided name or self.image_name.
+
+        Args:
+            name: (Optional), the name of the image to be retrieved.
+
+        Return:
+            NodeImage instance that is referenced by the image name.
+
+        Raises:
+            ResourceNotFound - when the named image resource does not exist in the given
+                               OpenStack cloud.
+        """
+        name = self.image_name if name is None else name
+        url = f"/v2/images?name={name}"
+        object_ = self.driver_v2.image_connection.request(url).object
+        images = self.driver_v2._to_images(object_, ex_only_active=False)
+
+        if not images:
+            raise ResourceNotFound(f"Failed to retrieve image with name: {name}")
+
+        return images[0]
 
     def _get_flavor(self) -> NodeSize:
         """Return the flavor reference."""
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ setup(
     author='Vasu Kulkarni',
     author_email='vasu@redhat.com',
     install_requires=[
-        'apache-libcloud',
+        'apache-libcloud>=3.3.0',
         'docopt==0.6.2',
         'gevent==1.4.0',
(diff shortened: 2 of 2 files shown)


mita/openstack.py
```diff
@@ -135,17 +135,29 @@ def get_driver_v2(self):
...
+        Raises:
+            ResourceNotFound - when no resource was found.
+        """
+        name = self.image_name if name is None else name
+        url = f"/v2/images?name={name}"
+        logger.info("Gathering image details: %s", name)
+        object_ = self.driver_v2.image_connection.request(url).object
+        images = self.driver_v2._to_images(object_, ex_only_active=False)
+
+        if len(images) != 1:
```
ktdreyer, line 157: In the previous code, we did not care if we found more than one image with the same name. A few years ago I remember the OpenStack image maintainers accidentally uploaded two images with the same name, and obviously that is unlikely to happen often, but we can be robust against that. Let's preserve the existing behavior, so we'll simply pick the first one.

In other words, please change the condition to `if not images:` instead of checking `len() != 1`. That will also match the docstring above that says `ResourceNotFound - when no resource was found`.

Let's also change the exception message to simply be `ResourceNotFound(f"image {name}")`, so users know exactly what type of resource we failed to find.

setup.py
```diff
@@ -13,7 +13,7 @@
     author='Vasu Kulkarni',
     author_email='vasu@redhat.com',
     install_requires=[
-        'apache-libcloud',
+        'apache-libcloud>=3.3.0',
```
ktdreyer, line 16: The commit message makes no mention about what bug this version solves, or why we would want to set this *specific* number. I guess that it happens to be the latest version on PyPI, but that might be a coincidence... in general if we're hard-coding version numbers, let's give some clues to the future maintainers about why we're doing it and how safe it might be to bump the numbers in the future, etc.

ktdreyer (changes requested):
Great, thanks. This is cleaner and more performant.

mita/openstack.py
```diff
@@ -135,17 +135,29 @@ def get_driver_v2(self):
...
+
+        Return:
+            NodeImage instance that is referenced by the name.
+
+        Raises:
+            ResourceNotFound - when no resource was found.
+        """
+        name = self.image_name if name is None else name
+        url = f"/v2/images?name={name}"
+        logger.info("Gathering image details: %s", name)
```
ktdreyer, line 153: I think we should pull out these `logger.info` statements, to reduce the overall cephci log size

mita/openstack.py
psathyan: agree... changes have been carried out.

mita/openstack.py
```diff
@@ -135,17 +135,28 @@ def get_driver_v2(self):
...
+        Return the NodeImage instance using the image_name or the provided name.
+
+        Args:
+            name: The name of the image.
+
+        Return:
+            NodeImage instance that is referenced by the name.
+
+        Raises:
+            ResourceNotFound - when no resource was found.
```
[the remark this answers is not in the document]
psathyan: Thanks... it should have included `image`. I agree with the comment.

mita/openstack.py
```diff
@@ -135,17 +135,28 @@ def get_driver_v2(self):
...
-                i for i in self.driver_v2.list_images() if i.name == self.image_name
-            ][0]
-        except IndexError:
-            raise ResourceNotFound("Image {} not found".format(self.image_name))
-        except BaseException as be:  # noqa
-            logger.error(be)
-            raise OpenStackDriverError("Encountered an unknown exception.")
+    def _get_image(self, name: Optional[str] = None) -> NodeImage:
+        """
+        Return the NodeImage instance using the image_name or the provided name.
```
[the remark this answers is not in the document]
psathyan: Makes sense and it convey's the right precedence.

ktdreyer (approved):

PR #296, opened 2021-01-12
Improve get_vm_size method.

psathyan:

# Description

In this PR, the `_get_vm_size` method is improved so that it returns the flavor as soon it matches the provided name or configuration set for `vm-size` in the configuration file.

Unlike image resource, we are unable to use `params` to request server side filtration due to lack of support from OpenStack. Hence, we retrieve the entire list from the server and the traverse the list to find a suitable flavor.

[Logs](http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/openstack/flavor.log)

### DIFF
--- a/mita/openstack.py
+++ b/mita/openstack.py
@@ -147,15 +147,25 @@ class CephVMNode(object):
             logger.error(be)
             raise OpenStackDriverError("Encountered an unknown exception.")
 
-    def _get_flavor(self) -> NodeSize:
-        """Return the flavor reference."""
-        try:
-            return [f for f in self.driver_v2.list_sizes() if f.name == self.vm_size][0]
-        except IndexError:
-            raise ResourceNotFound("Flavor {} not found".format(self.vm_size))
-        except BaseException as be:  # noqa
-            logger.error(be)
-            raise OpenStackDriverError("Encountered an unknown exception.")
+    def _get_vm_size(self, name: Optional[str] = None) -> NodeSize:
+        """
+        Return the NodeSize instance retrieved based on the provided name or vm_size.
+
+        Args:
+            name: The OpenStack flavor (m1.small or m1.medium or m1.large) to be found.
+
+        Return:
+            Reference to NodeSize
+
+        Raises:
+            ResourceNotFound exception when none resources found.
+        """
+        name = self.vm_size if name is None else name
+        for flavor in self.driver_v2.list_sizes():
+            if flavor.name == name:
+                return flavor
+
+        raise ResourceNotFound("No flavor matched the name %s", name)
 
     def _has_free_ip_address(self, network: str) -> bool:
         """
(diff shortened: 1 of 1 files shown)


mita/openstack.py
```diff
@@ -147,15 +147,27 @@ def _get_image(self) -> NodeImage:
...
-    def _get_flavor(self) -> NodeSize:
-        """Return the flavor reference."""
-        try:
-            return [f for f in self.driver_v2.list_sizes() if f.name == self.vm_size][0]
-        except IndexError:
-            raise ResourceNotFound("Flavor {} not found".format(self.vm_size))
-        except BaseException as be:  # noqa
-            logger.error(be)
-            raise OpenStackDriverError("Encountered an unknown exception.")
+    def _get_flavor(self, name: Optional[str] = None) -> NodeSize:
```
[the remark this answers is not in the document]
sunilkumarn417: https://github.com/red-hat-storage/cephci/pull/291/files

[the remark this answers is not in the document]
psathyan: It is the same... in this PR, I am breaking out the changes of an earlier PR #293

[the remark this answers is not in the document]
psathyan: Unlike image, network or volume, we don't have the option of using `params` to pick the exact resource reference.

sunilkumarn417:
Looks like another PR(295) has same title, I think this PR title need to be changed w.r.t "_get_flavor".

mita/openstack.py
```diff
@@ -147,15 +147,27 @@ def _get_image(self) -> NodeImage:
...
+        """
+        Return the NodeSize instance retrieved based on the provided name or vm_size.
+
+        Args:
+            name:   The name of the flavor to be retrieved
```
ktdreyer, line 155: It would be great to provide an example here, like `m1.large` or whatever we expect

mita/openstack.py
```diff
@@ -147,15 +147,27 @@ def _get_image(self) -> NodeImage:
...
+
+        Return:
+            Reference to NodeSize
+
+        Raises:
+            ResourceNotFound exception when none resources found.
+        """
+        name = self.vm_size if name is None else name
+        logger.info("Gathering flavor details of %s", name)
```
ktdreyer, line 164: This logging is excessive, and the cephci logs are already massive at INFO level. Can we remove this `logger.info()` statement?

mita/openstack.py
```diff
@@ -147,15 +147,27 @@ def _get_image(self) -> NodeImage:
...
-        try:
-            return [f for f in self.driver_v2.list_sizes() if f.name == self.vm_size][0]
-        except IndexError:
-            raise ResourceNotFound("Flavor {} not found".format(self.vm_size))
-        except BaseException as be:  # noqa
-            logger.error(be)
-            raise OpenStackDriverError("Encountered an unknown exception.")
+    def _get_flavor(self, name: Optional[str] = None) -> NodeSize:
+        """
+        Return the NodeSize instance retrieved based on the provided name or vm_size.
```
ktdreyer, line 152: I recommend amending this to say `self.vm_size` so it's clear where "vm_size" comes from.

ktdreyer (changes requested):
Great, thank you

mita/openstack.py
psathyan: Done.

ktdreyer (approved):

PR #297, opened 2021-01-12
Improve get_network method

psathyan:

# Description
This PR modifies the following areas
- Identifying the available private network via filtration at server side
- Populating the value during gathering of network information

[Logs](http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/openstack/network.log)

### DIFF
--- a/mita/openstack.py
+++ b/mita/openstack.py
@@ -182,49 +182,78 @@ class CephVMNode(object):
 
         raise ResourceNotFound(f"Failed to retrieve vm size with name: {name}")
 
-    def _has_free_ip_address(self, network: str) -> bool:
+    def _get_network_by_name(self, name: str) -> OpenStackNetwork:
         """
-        Return True if the given network has an IPv4 Address to lease.
+        Retrieve the OpenStackNetwork instance using the provided name.
 
-        Arguments:
-            network: String, The network UUID
+        Args:
+            name:   the name of the network.
 
         Returns:
-            bool - True on success else False
-
-        Note: This method returns True only if the given network has more than 3
-        """
-        try:
-            resp = self.driver_v2.network_connection.request(
-                "/v2.0/network-ip-availabilities/{}".format(network),
-            )
-        except BaseException as be:
-            logger.info(be)
-            return False
-
-        if resp.status != 200:
-            logger.debug("Failed to gather the network details.")
-            return False
-
-        _total = resp.object.get("network_ip_availability", {}).get("total_ips")
-        _used = resp.object.get("network_ip_availability", {}).get("used_ips")
-
-        return (_total - _used) > 3
-
-    def _get_network(self, network: Optional[str] = None) -> OpenStackNetwork:
-        """
-        Return the provider network.
-
-        Arguments:
-            network: Network to be attached to the VM
-
-        Returns:
-            An OpenStackNetwork libcloud object
+            OpenStackNetwork instance referenced by the name.
 
         Raises:
-            GetIPError when there are no suitable networks
+            ResourceNotFound: when the named network resource does not exist in the
+                              given OpenStack cloud
         """
-        _networks = (
+        url = f"{self.driver_v2._networks_url_prefix}?name={name}"
+        object_ = self.driver_v2.network_connection.request(url).object
+        networks = self.driver_v2._to_networks(object_)
+
+        if not networks:
+            raise ResourceNotFound(f"No network resource with name {name} found.")
+
+        return networks[0]
+
+    def _has_free_ip_addresses(self, net: OpenStackNetwork) -> bool:
+        """
+        Return True if the given network has more than 3 free ip addresses.
+
+        This buffer of 3 free IPs is in place to avoid failures during node creation.
+        As in OpenStack, the private IP request for allocation occurs towards the end
+        of the workflow.
+
+        When a subnet with free IPs is identified then it's CIDR information is
+        assigned to self.subnet attribute on this object.
+
+        Arguments:
+            net:    The OpenStackNetwork instance to be checked for IP availability.
+
+        Returns:
+            True on success else False
+        """
+        url = f"/v2.0/network-ip-availabilities/{net.id}"
+        resp = self.driver_v2.network_connection.request(url)
+        subnets = resp.object["network_ip_availability"]["subnet_ip_availability"]
+
+        for subnet in subnets:
+            free_ips = subnet["total_ips"] - subnet["used_ips"]
+
+            if free_ips > 3:
+                self.subnet = subnet["cidr"]
+                return True
+
+        return False
+
+    def _get_network(self, name: Optional[str] = None) -> OpenStackNetwork:
+        """
+        Return the first available OpenStackNetwork with a free IP address to lease.
+
+        This method will search a preconfigured list of network names and return the
+        first one that has more than 3 IP addresses to lease. One can override the
+        preconfigured list by specifying a single network name.
+
+        Args:
+            name: (Optional), the network name to be retrieved in place of the default
+                              list of networks.
+
+        Returns:
+            OpenStackNetwork instance that has free IP addresses to lease.
+
+        Raises:
+            ResourceNotFound when there no suitable networks in the environment.
+        """
+        network_names = (
             [
                 "provider_net_cci_8",
                 "provider_net_cci_7",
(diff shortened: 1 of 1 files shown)


mita/openstack.py
```diff
@@ -157,47 +157,73 @@ def _get_flavor(self) -> NodeSize:
...
-        return (_total - _used) > 3
+        logger.info("%s has no IP addresses to lease.", net.name)
+        return False
 
     def _get_network(self, network: Optional[str] = None) -> OpenStackNetwork:
         """
-        Return the provider network.
+        Return the instance of OpenStackNetwork from the known list or given name.
+
+        A network object is returned only if it contains a free IP address to lease.
```
ktdreyer, line 217: Let's rewrite this text to be more specific:

```
Return the first available OpenStackNetwork with a free IP address to lease.

This method will search a preconfigured list of network names and return the
first one with a free IP address. You can override the list of networks by
specifying a single network's name.
```

mita/openstack.py
```diff
@@ -157,47 +157,73 @@ def _get_flavor(self) -> NodeSize:
...
 
         Arguments:
-            network: Network to be attached to the VM
+            network: The name of the network to be retrieved.
```
ktdreyer, line 220: Add: `If you do not specify a name, this method will search through a default list of names and pick the first one with a free IP.`

mita/openstack.py
```diff
@@ -157,47 +157,73 @@ def _get_flavor(self) -> NodeSize:
...
 
         Returns:
-            An OpenStackNetwork libcloud object
+            An OpenStackNetwork libcloud object.
 
         Raises:
-            GetIPError when there are no suitable networks
+            ResourceNotFound is raised when there are no suitable networks.
         """
         _networks = (
```
ktdreyer, line 228: This is not specifically related to your PR, and we can fix this later in order to keep the scope of this change manageable, but I'll just mention this here: I wonder why this variable is named `_networks` instead of simply `networks`, or more clearly, `network_names`. We could probably rename this (separate PR).

mita/openstack.py
```diff
@@ -211,28 +237,20 @@ def _get_network(self, network: Optional[str] = None) -> OpenStackNetwork:
...
+            # Treating an exception as a soft error as it is possible to find another
+            # suitable network from the list.
             try:
-                os_net = [n for n in _available_networks if n.name == net][0]
+                os_net = self._get_network_by_name(name=net)
 
-                if not self._has_free_ip_address(os_net.id):
-                    logger.debug("%s does not meet the requirements", net)
+                if not self._has_free_ip_address(net=os_net):
+                    logger.debug("%s has no free ip addresses.", os_net.name)
```
ktdreyer, line 247: To make the logs easier to read, let's take this `logger.debug()` statement out and only log the network names if we hit `ResourceNotFound` at the end.

mita/openstack.py
```diff
@@ -211,28 +237,20 @@ def _get_network(self, network: Optional[str] = None) -> OpenStackNetwork:
...
-                    if s.id in os_net.extra.get("subnets")
-                ][0]
-
                 return os_net
-            except TypeError:
-                logger.warning("%s is not accessible.", net)
             except BaseException as be:  # noqa
-                logger.warning("Something went wrong during network selection process.")
-                logger.info(be)
+                logger.debug(be)
```
ktdreyer, line 252: Instead of `logger.debug()`, It's appropriate to keep this at the WARNING level with `logger.warning()` here. Ideally we never hit this exception, but if we do, it could indicate a problem on the OpenStack server that we might report to the OpenStack administrators.

mita/openstack.py
```diff
@@ -211,28 +237,20 @@ def _get_network(self, network: Optional[str] = None) -> OpenStackNetwork:
...
+                continue
 
         raise ResourceNotFound("No suitable network resource found.")
```
ktdreyer, line 255: I think we should log the full list of names that we searched here instead, like:

`raise ResourceNotFound(f"No networks had free IP addresses: {_networks}")`

mita/openstack.py
```diff
@@ -211,28 +237,20 @@ def _get_network(self, network: Optional[str] = None) -> OpenStackNetwork:
             else [network]
         )
 
-        _available_networks = self.driver_v2.ex_list_networks()
-
         for net in _networks:
+            # Treating an exception as a soft error as it is possible to find another
+            # suitable network from the list.
```
ktdreyer, line 242: Great comment explaining *why* we want to `try`/`except` here. :+1:

mita/openstack.py
```diff
@@ -157,47 +157,73 @@ def _get_flavor(self) -> NodeSize:
...
             bool - True on success else False
 
         Note: This method returns True only if the given network has more than 3
         """
         try:
-            resp = self.driver_v2.network_connection.request(
-                "/v2.0/network-ip-availabilities/{}".format(network),
+            url = f"/v2.0/network-ip-availabilities/{net.id}"
+            resp = self.driver_v2.network_connection.request(url)
+            subnets = resp.object.get("network_ip_availability", {}).get(
```
ktdreyer, line 197: I think this is good that you're being careful with the data by using `.get()`, however it makes it a little harder to tell if the code breaks later.

As one example, `network_ip_availability` is always going to be present there, right? (I'm asking because I don't know)

For the fields that we *know* will always be present, I would rather see us raise the `KeyError`s here immediately, because otherwise we could cruise on through this code without logging any errors and then find no `_free_ips` for anything.

mita/openstack.py
```diff
@@ -157,47 +157,73 @@ def _get_flavor(self) -> NodeSize:
...
+
+        return networks[0]
+
+    def _has_free_ip_address(self, net: OpenStackNetwork) -> bool:
+        """
+        Return True if the given network has an ip address to lease.
 
         Arguments:
-            network: String, The network UUID
+            net:    The reference to OpenStackNetwork
```
ktdreyer, line 187: This is a great change. It makes a lot more sense to me to pass in a full `OpenStackNetwork` object than to pass the UUID. Thanks.

ktdreyer (changes requested):

mita/openstack.py
psathyan: Thank you.

mita/openstack.py
```diff
@@ -157,47 +157,73 @@ def _get_flavor(self) -> NodeSize:
...
-            return False
+            for subnet in subnets:
+                _free_ips = subnet.get("total_ips") - subnet.get("used_ips")
 
-        _total = resp.object.get("network_ip_availability", {}).get("total_ips")
-        _used = resp.object.get("network_ip_availability", {}).get("used_ips")
+                if _free_ips > 3:
+                    self.subnet = subnet.get("cidr")
+                    return True
+        except BaseException as be:  # noqa
```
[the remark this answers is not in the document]
psathyan: Thank you. Made the required changes.

mita/openstack.py
```diff
@@ -157,47 +157,73 @@ def _get_flavor(self) -> NodeSize:
...
+            logger.warning(be)
 
-        return (_total - _used) > 3
+        logger.info("%s has no IP addresses to lease.", net.name)
```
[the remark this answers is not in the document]
psathyan: Thanks... I agree the logging was bit noisy.

mita/openstack.py
psathyan: done.

mita/openstack.py
psathyan: done

mita/openstack.py
psathyan: Done

mita/openstack.py
psathyan: Done

mita/openstack.py
psathyan: Done...

mita/openstack.py
```diff
@@ -157,84 +157,105 @@ def _get_flavor(self) -> NodeSize:
...
+        return networks[0]
 
-    def _get_network(self, network: Optional[str] = None) -> OpenStackNetwork:
+    def _has_free_ip_address(self, net: OpenStackNetwork) -> bool:
         """
-        Return the provider network.
+        Return True if the given network has an ip address to lease.
+
+        This method determines if the provided network has more than 3 IP addresses to
+        lease. The buffer is in place to avoid failures during node creation. As in
```
ktdreyer, line 188: Thank you for explaining this. I never knew why this code worked like this.

Maybe we should rename this method with the plural form "addresses", like `_has_free_ip_addresses`.

mita/openstack.py
```diff
@@ -157,84 +157,105 @@ def _get_flavor(self) -> NodeSize:
...
+        OpenStack's create node method, the private IP request for allocation occurs
+        towards the end.
+
+        The CephVMNode attribute self.subnet takes the CIDR information of the OpenStack
```
ktdreyer, line 192: How about we rewrite this to: `If this method finds a subnet with free IPs, it assigns the network's subnet CIDR information to the self.subnet attribute on this object.`

mita/openstack.py
```diff
@@ -157,84 +157,105 @@ def _get_flavor(self) -> NodeSize:
...
+
+        Returns:
+            OpenStackNetwork instance that has free IP addresses to lease.
 
         Raises:
-            GetIPError when there are no suitable networks
+            ResourceNotFound when there no suitable networks in the environment.
         """
-        _networks = (
+        network_names = (
```
ktdreyer, line 232: :heart: the rename

mita/openstack.py
psathyan: Thank you... I wasn't sure how to frame it hence `CephVMNode`.

ktdreyer (approved):

PR #298, opened 2021-01-12
Simplify the _create_vm_node method

psathyan:

# Description

It is fatal if any off the steps of `_create_vm_node` fails. Removing the `try/except` allows us to bubble up the error to `resurrect_traceback`. 

Anotther change is the reduction of the wait time.

[Logs](http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/openstack/create_node.log)

### DIFF
--- a/mita/openstack.py
+++ b/mita/openstack.py
@@ -2,7 +2,6 @@
 import datetime
 import logging
 import socket
-from ssl import SSLError
 from time import sleep
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


mita/openstack.py
```diff
@@ -238,50 +237,36 @@ def _get_network(self, network: Optional[str] = None) -> OpenStackNetwork:
...
-        raise NodeErrorState("Failed to create the instance {}".format(self.node_name))
+        if self.node is None:
+            raise NodeErrorState(f"Failed to create {self.node_name}")
 
     def _wait_until_vm_state_running(self):
         """Wait till the VM moves to running state."""
         timeout = datetime.timedelta(seconds=600)
         start_time = datetime.datetime.now()
-        while True:
-            logger.info("Waiting for %s state to be running ", self.node_name)
```
ktdreyer, line 270: I kinda liked this message, but it was noisy. I wonder if it would be more useful to log the total time that we've waited so far, like "Waited {timedelta} for {self.node_name} to start running" or something. What do you think?

ktdreyer (commented):
A great improvement. Just one comment about the "Waiting" log message.

mita/openstack.py
psathyan: Finding the time taken to move to `running` state is straight forward. I agree it is a useful data point if monitor...

ktdreyer: Oh, I was thinking more like logging it while we're waiting, before `node.state == "running"`, so the user has some visual indication of progress. But this implementation is fine too.

ktdreyer (dismissed):

ktdreyer (approved):

PR #299, opened 2021-01-12
Improve retrieval of volume details

psathyan:

# Description

In this PR, improvements to volume create & attach methods have been carried out. 

[Logs](http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/openstack/volume.log)

### DIFF
--- a/mita/openstack.py
+++ b/mita/openstack.py
@@ -7,7 +7,7 @@ from time import sleep
 from typing import Optional
 
 from libcloud.compute.base import NodeImage, NodeSize
-from libcloud.compute.drivers.openstack import OpenStackNetwork
+from libcloud.compute.drivers.openstack import OpenStackNetwork, StorageVolume
 from libcloud.compute.providers import get_driver
 from libcloud.compute.types import Provider
 
@@ -54,7 +54,7 @@ class OpenStackDriverError(Exception):
     pass
 
 
-class VolumeAttachmentError(Exception):
+class VolumeOpError(Exception):
     pass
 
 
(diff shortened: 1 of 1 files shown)


mita/openstack.py
```diff
@@ -310,63 +310,46 @@ def _wait_until_ip_is_known(self):
...
-            )
-            try:
-                _vol = self.driver_v2.create_volume(
-                    self.size_of_disk, "{}-vol-{}".format(self.node_name, item)
-                )
-                self.volumes.append(_vol)
-            except BaseException as be:  # noqa
-                logger.debug("Failed to create volume.")
-                logger.debug(be)
+            vol_name = f"{self.node_name}-vol-{item}"
```
ktdreyer, line 313: It's much clearer when you've moved `vol_name` to its own assignment. Thanks.

mita/openstack.py
```diff
@@ -310,63 +310,46 @@ def _wait_until_ip_is_known(self):
...
-        ok_states = ["creating"]  # it's ok to wait if the volume is in this
+    def _wait_until_volume_available(self, volume: StorageVolume) -> bool:
+        """Wait until the state of the StorageVolume is available."""
         tries = 0
-        if maybe_in_use:
-            ok_states.append("in_use")
-        logger.info("Volume: %s is in state: %s", volume.name, volume.state)
-
-        while volume.state in ok_states:
+        logger.debug("%s waiting for volume to become available.", volume.name)
```
ktdreyer, line 335: Let's rewrite this string as `Waiting for %s volume to become available`

mita/openstack.py
```diff
@@ -310,63 +310,46 @@ def _wait_until_ip_is_known(self):
...
+                return True
+
+            if "error" in volume.state.lower():
+                logger.error("%s state is %s", volume.name, volume.state)
                 break
-            if volume.state == "notfound":
-                logger.error("no volume was found for: %s", volume.name)
+
+            if tries > 10:
+                logger.error("Maximum amount of retries reached.")
```
ktdreyer, line 349: Let's log the volume name here, `logger.error(f'Max retries for {volume.name} reached')`

mita/openstack.py
```diff
@@ -376,14 +359,16 @@ def get_private_ip(self):
...
-        volumes = driver.list_volumes()
-        try:
-            return [v for v in volumes if v.name == name][0]
-        except IndexError:
-            raise RuntimeError("Unable to get volume")
+    def get_volume(self, name: str) -> StorageVolume:
+        """Retrieve the StorageVolume instance using the given name."""
+        url = f"/volumes?name={name}"
+        object_ = self.driver_v2.volumev2_connection.request(url).object
+
```
ktdreyer, line 366: We have `object_.get("volumes", [])` here, and then later on we do `object_.get("volumes")[0]`...

Let's simply assign `volumes = object_.get("volumes", [])` once, and then use that `volumes` variable in the rest of the code.

mita/openstack.py
```diff
@@ -376,14 +359,16 @@ def get_private_ip(self):
...
+        if len(object_.get("volumes", [])) != 1:
```
ktdreyer, line 367: The old code did not care about multiple volumes with the same name, but here we're checking that `len()` is exactly one. Let's rewrite this to `if not volumes:`

mita/openstack.py
```diff
@@ -376,14 +359,16 @@ def get_private_ip(self):
...
+            raise ResourceNotFound(f"Exact match failed for volume with name: {name}.")
+
+        volume_info = object_.get("volumes")[0]
```
ktdreyer, line 370: After rewriting (per above) to use a new `volumes` variable, we can edit this line to simply be `volume_info = volumes[0]`

mita/openstack.py
```diff
@@ -394,6 +379,8 @@ def create_node(self):
         if self.no_of_volumes:
             self._create_attach_volumes()
 
+        logger.info("Successfully created VM node with name %s.", self.node_name)
```
ktdreyer, line 382: Let's take this out of this PR since it's logically separate

ktdreyer (changes requested):

mita/openstack.py
psathyan: Yes it would make it more readable.

mita/openstack.py
psathyan: done

mita/openstack.py
psathyan: changed.

mita/openstack.py
```diff
@@ -309,64 +309,50 @@ def _wait_until_ip_is_known(self):
 
     def _create_attach_volumes(self):
         """Create and attach the volumes."""
+        logger.info(
```
ktdreyer, line 312: Fewer, clearer log statements :heart:

ktdreyer (approved):

mita/openstack.py
psathyan: thank you

PR #314, opened 2021-01-20
CEPHADM: added ISCSI support

sunilkumarn417:
- Added iscsi support.

**Test results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1611113270349/

### DIFF
--- a/ceph/ceph_admin.py
+++ b/ceph/ceph_admin.py
@@ -212,8 +212,8 @@ class CephAdmin:
         self.ceph_osds(self.cluster.get_nodes(role="osd"))
         self.ceph_mdss(self.cluster.get_nodes(role="mds"))
         self.ceph_rgws(self.cluster.get_nodes(role="rgw"))
+        self.ceph_iscsi(self.cluster.get_nodes(role="iscsi"))
         self.ceph_clients(self.cluster.get_nodes(role="client"))
-        # self.ceph_iscsi(self.cluster.get_nodes(role="iscsi"))
 
         # monitoring tools
         daemons = ["prometheus", "node-exporter", "alertmanager", "grafana"]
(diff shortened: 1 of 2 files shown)


ceph/ceph_admin.py
```diff
@@ -548,8 +548,55 @@ def ceph_iscsi(self, nodes, **kwargs):
             nodes: iscsi node list
             kwargs: key-value arguments
```
psathyan, line 549: Please document what is supported and what are the required keys. If it is for future use, please document it. IMO, future use... can be avoided.

ceph/ceph_admin.py
```diff
@@ -548,8 +548,55 @@ def ceph_iscsi(self, nodes, **kwargs):
...
+                    ]
+                )
+
+            # check daemon existence
+            assert self.check_exist(
+                daemon='iscsi',
+                ids=daemons,
+                timeout=900,
+            )
+        elif op == "purge":
```
psathyan, line 597: Could we avoid the `elif` statement instead opt to use `else`?

ceph/ceph_admin.py
```diff
@@ -548,8 +548,55 @@ def ceph_iscsi(self, nodes, **kwargs):
...
+            daemons = []
+            for node in nodes:
+                logger.info("adding iscsi on %s-%s" % (node.hostname,
+                                                       node.ip_address))
+                iscsis.append("{host}=iscsi.{host}".format(host=node.hostname))
+                daemons.append("iscsi.{host}".format(host=node.hostname))
+
+            if iscsis:
+                pool_name = "iscsi"
+
```
psathyan, line 565: Ignore if it is unreasonable...
```
script = f"""
ceph osd pool create {pool_name} 3 3 replicated
ceph osd pool application enable {pool_name} rbd
ceph orch apply iscsi {pool_name} user password --placement '{len(iscsis)};{";".join(iscsis)}'
"""[1:-1]

for cmd in script.splitlines(): self.shell(remote=self.installer, args=cmd)
```

ceph/ceph_admin.py
sunilkumarn417: yeah its been for future-use.

ceph/ceph_admin.py
sunilkumarn417: we may have update/modify in future to update services, we can take this in future if we don't have further actions on services.

ceph/ceph_admin.py
sunilkumarn417: yes, its for future use, in case user deploy services with existing pool and may have further CLI statements.

ceph/ceph_admin.py
```diff
@@ -546,10 +582,76 @@ def ceph_iscsi(self, nodes, **kwargs):
...
-            kwargs: key-value arguments
+            kwargs: key-value arguments (future-use)
+
+        kwargs:
+            op: "create|purge" # (default: create, required: true)
         """
-        # todo: Handle iscsi
-        raise NotImplementedError
+        op = kwargs.get("op", "create").lower()
+
```
psathyan, line 591: If the parameter `nodes` is empty, we should avoid going down the workflow.

psathyan (commented):
Thank you... the formatting is consistent

ceph/ceph_admin.py
```diff
@@ -12,6 +12,17 @@
 logger = logging.getLogger(__name__)
 
 
+def nodes_is_empty(func):
+    def func_wrapper(*args, **kwargs):
+        if not kwargs.get("nodes"):
+            logger.warning("Nodes not found")
```
psathyan, line 18: IMO, we shouldn't warn as it is acceptable to have a configuration without some of the roles.

ceph/ceph_admin.py
```diff
@@ -12,6 +12,17 @@
...
+            return
+        logger.info("Nodes found: %s", kwargs.get("nodes"))
```
psathyan, line 20: This would be noisy... let's avoid it.

ceph/ceph_admin.py
psathyan: Nice work... the decorator idea is good.

ceph/ceph_admin.py
```diff
@@ -497,59 +545,132 @@ def ceph_monitoring(self, daemon, nodes, **kwargs):
...
+        nodes = nodes if isinstance(nodes, list) else [nodes]
+        if op == "create":
+            iscsis = []
+            daemons = []
+            for node in nodes:
+                logger.info("adding iscsi on %s-%s" % (node.hostname, node.ip_address))
+                iscsis.append("{host}=iscsi.{host}".format(host=node.hostname))
+                daemons.append("iscsi.{host}".format(host=node.hostname))
+
+            if iscsis:
```
psathyan, line 620: Since the list cannot be empty... with the `deco`, I think this check can be avoided.

psathyan (approved):
Looks great!!!

sunilkumarn417:
Latest test result : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1611147137285

PR #324, opened 2021-01-27
ansible suite failure fix

rakeshgm:
fixing the ansible installation 

Signed-off-by: rakeshgm <rakeshgm@redhat.com>


veera-raghava-reddy:
@rakeshgm Should we have check to define variable if the setup is MS?

rakeshgm:
> @rakeshgm Should we have check to define variable if the setup is MS?

yes @veera-raghava-reddy we do have a variable called `rgw_multisite` and we have the same check in this fix

veera-raghava-reddy (approved):
We can merge this PR.

PR #327, opened 2021-01-28
Adding S3Test suite support for RHCS 5

psathyan:
# Description

In this PR, support for s3test execution in a RHCS 5 environment is being added. Due to `cephadm`, the following changes are done

- the way conf file is modified
- service/daemon restart
- Node selection changed due to the interface.

Additional changes
- Introducing new SUT for reducing VM footprint.

**Logs**
[RHCS 5](http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/s3_tests/5/)
[RHCS 4.x](http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/s3_tests/4/)

### DIFF
--- a/conf/inventory/rhel-8.3-server-x86_64-medlarge.yaml
+++ b/conf/inventory/rhel-8.3-server-x86_64-medlarge.yaml
@@ -1,0 +1,38 @@
+version_id: 8.3
+id: rhel
+instance:
+  create:
+    image-name: rhel-8.3-server-x86_64-released
+    vm-size: ci.m1.medlarge
+
+  setup: |
+    #cloud-config
+
+    ssh_pwauth: True
+
+    groups:
+        - cephuser
+
+    users:
+      - name: cephuser
+        primary-group: cephuser
+        sudo: ALL=(ALL) NOPASSWD:ALL
+        shell: /bin/bash
+        ssh-authorized-keys:
+           - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDI0tHxJQ7n+uMiLpsoR6CKAVd0xatgVQuqp/gmnGpZU0kE54a29vPNnEt7/aLitbfyhc57rrbHOT09H3ov74GZKkoVBSbMJUSsK3drbN+58wcuk+HK0htRewmwCfcfi9AkrVbyw6pbPXW/pbjxnxLep52fKmpJJnImZ5eHRV5le9OSAcLA1LHYR4y9R3IOrTp7jgpE205UxZi5OopAx7gkyTsmfydvmq4MjaSwbVOJ7aW/Fdt5FVxNJP3Zl/OrvDoo/1WovoRIDbVQH8JFpLikMSnCqtBVIHDeW6imAKl6dpn9Gf4FxD94+OcurhXo2p0pvSzC4Strg4d2Sxqh4wph jenkins-build@jenkins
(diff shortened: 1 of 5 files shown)


tests/rgw/test_s3.py
```diff
@@ -18,101 +44,247 @@ def run(**kw):
...
+    Args:
+        nodes:  List of the nodes participating in the cluster
+        role:   The role of the node to be retrieved.
     """
-    log.info("Removing existing s3-tests directory if it exists")
-    client_node.exec_command(cmd="if test -d s3-tests; then rm -r s3-tests; fi")
+    for node in nodes:
+        if node.role == role:
+            return node
+
```
rakeshgm, line 165: I think we could make use this https://github.com/red-hat-storage/cephci/blob/be1fc08a2856a46f2bdb6ad5877faf0ad4b8250b/ceph/ceph.py#L118 or this https://github.com/red-hat-storage/cephci/blob/be1fc08a2856a46f2bdb6ad5877faf0ad4b8250b/ceph/ceph.py#L102

suites/pacific/rgw/sanity_rgw.yaml
```diff
@@ -0,0 +1,36 @@
...
+          - 'ceph -s'
+      desc: Verify the health of the storage cluster.
+      destroy-cluster: false
+      module: test_cephadm.py
+      name: verify cluster health
+
+  - test:
+      abort-on-fail: true
+      config:
+        branch: master
```
rakeshgm, line 32: do you want to use `master` or`pacific`. I think its best we use `pacific` branch here. cause `master` always points to latest release. Because of this we may have to update this value in the future, which we may forget :)

tests/rgw/test_s3.py
```diff
@@ -18,101 +44,247 @@ def run(**kw):
...
 
-    log.info("Cloning s3-tests repository")
-    branch = config.get("branch", "ceph-luminous")
+def clone_s3_tests(node: CephNode, branch="master") -> None:
```
rakeshgm, line 167: we already take the branch value from the `config` do we still want to add the default value as `master` here again

conf/inventory/rhel-8.3-server-x86_64-medlarge.yaml
```diff
@@ -0,0 +1,42 @@
+version_id: 8.3
+id: rhel
+instance:
+  create:
+    image-name: rhel-8.3-server-x86_64-released
+    vm-size: ci.m1.medlarge
```
rakeshgm, line 6: any particular reason  for this `ci.m1.medlarge` size

rakeshgm:
the doc strings are nice :)

suites/pacific/rgw/sanity_rgw.yaml
psathyan: Ideally, it should be `ceph-pacific`. At the time of pushing the PR, the branch was not created hence I left it as `master`. Now that there is a `ceph-pacific` branch, I'll modify it.

tests/rgw/test_s3.py
psathyan: As I noticed that the branch was not created, I wanted a mechanism to pick the latest... hence the default value. This allows us to avoid specifying the branch name incase of dev runs...

conf/inventory/rhel-8.3-server-x86_64-medlarge.yaml
psathyan: This allows us to reduce the VM footprint but the hard truth is that the memory footprint has not reduced to the expected extended. For sanity, I believe we can keep the footprint small.

tests/rgw/test_s3.py
psathyan: Thanks... it makes sense to reuse

conf/inventory/rhel-8.3-server-x86_64-medlarge.yaml
rakeshgm: so can you remove this inventory file, we can use the existing rhel-8.3 inventory file since other tests will also added in this suite soon.

psathyan: @rakeshgm I don't remember if I got you convinced on using the newer inventory file during our side discussions.

rakeshgm (approved):

PR #331, opened 2021-01-30
Adding new cli option to clean stale volumes

veera-raghava-reddy:
# Description
python3 -m venv .;source ./bin/activate;pip install -r requirements.txt;ls -lt
python run.py --osp-cred osp/osp-cred-ci-2.yaml --clean-volumes=yes --log-level INFO

http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1611999407374/startup.log

### DIFF
--- a/ceph/utils.py
+++ b/ceph/utils.py
@@ -165,6 +165,21 @@ def volume_cleanup(volume, osp_cred):
                     return 1
 
 
+def cleanup_ceph_vols(osp_cred):
+    '''
+    Added this function to cleanup stale volumes with status deleting, available, error, unknown, reserved
+    '''
+    vol_states = [ "deleting","available", "error", "unknown", "reserved" ]
+    driver = get_openstack_driver(osp_cred)
+    with parallel() as p:
+        for volume in driver.list_volumes():
+            if volume.state in vol_states:
+                log.info("volume with id:%s, name:%s has status:%s", volume.id, volume.name, volume.state)
+                p.spawn(volume_cleanup, volume, osp_cred)
+                sleep(1)
+    sleep(30)
+
+
 def keep_alive(ceph_nodes):
     for node in ceph_nodes:
         node.exec_command(cmd='uptime', check_ec=False)
--- a/run.py
+++ b/run.py
@@ -23,8 +23,16 @@ from ceph.clients import WinNode
 from ceph.utils import create_ceph_nodes, cleanup_ceph_nodes
 from utility.polarion import post_to_polarion
 from utility.retry import retry
-from utility.utils import timestamp, create_run_dir, create_unique_test_name, create_report_portal_session, \
-    configure_logger, close_and_remove_filehandlers, get_latest_container, email_results
+from utility.utils import (
+    timestamp,
+    create_run_dir,
+    create_unique_test_name,
+    create_report_portal_session,
+    configure_logger,
+    close_and_remove_filehandlers,
+    get_latest_container,
+    email_results,
+)
 from utility.xunit import create_xunit_results
 
 doc = """
@@ -109,7 +117,7 @@ log = logging.getLogger(__name__)
 root = logging.getLogger()
 root.setLevel(logging.INFO)
 
-formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
 
 ch = logging.StreamHandler(sys.stdout)
 ch.setLevel(logging.ERROR)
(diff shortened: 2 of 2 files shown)


veera-raghava-reddy:
[http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1611999407374/startup.log](url)

ceph/utils.py
```diff
@@ -165,6 +165,33 @@ def volume_cleanup(volume, osp_cred):
                     return 1
 
 
+# Added this function to cleanup stale volumes with status deleting, available, error, unknown, reserved
```
sunilkumarn417, line 168: can you please this as doc string within definition.
```
def cleanup_ceph_vols(...):
'''
Added this function to cleanup stale volumes with status deleting, available, error, unknown, reserved
'''

File: ceph/utils.py [hunk 2]
```diff
@@ -165,6 +165,33 @@ def volume_cleanup(volume, osp_cred):
...
+def cleanup_ceph_vols(osp_cred, pattern=None, timeout=300):
```
sunilkumarn417, line 169: I think pattern param is not used, are we deleting the volumes by name pattern?

sunilkumarn417: it would be good if we use name pattern, so it would be safe delete.

ceph/utils.py
```diff
@@ -165,6 +165,33 @@ def volume_cleanup(volume, osp_cred):
...
+    driver = get_openstack_driver(osp_cred)
+    timeout = datetime.timedelta(seconds=timeout)
+    for volume in driver.list_volumes():
```
sunilkumarn417, line 172: IMO, this won't be required, since we have another loop where it iterates volumes again. there we can have log with vol status.

ceph/utils.py
```diff
@@ -165,6 +165,33 @@ def volume_cleanup(volume, osp_cred):
...
+        log.info("volume %s status is %s", volume.name, volume.state)
+    with parallel() as p:
+        for volume in driver.list_volumes():
+            if volume.state in "deleting":
+                log.info("volume %s %s status is %s", volume.id, volume.name, volume.state)
+                p.spawn(volume_cleanup, volume, osp_cred)
+            if volume.state in "available":
```
sunilkumarn417, line 179: do we need to consider this state `available` for deleting?

ceph/utils.py
```diff
@@ -165,6 +165,33 @@ def volume_cleanup(volume, osp_cred):
...
 
+# Added this function to cleanup stale volumes with status deleting, available, error, unknown, reserved
+def cleanup_ceph_vols(osp_cred, pattern=None, timeout=300):
+    driver = get_openstack_driver(osp_cred)
+    timeout = datetime.timedelta(seconds=timeout)
+    for volume in driver.list_volumes():
+        log.info("volume %s status is %s", volume.name, volume.state)
+    with parallel() as p:
+        for volume in driver.list_volumes():
+            if volume.state in "deleting":
```
sunilkumarn417, line 176: IMO, it would be nice if vol state list is available and any vol state which falls could be considered for deletion.
```
vol_states = [ "deleting", "unknown", . , .. ,]
if vol.state in vol_states:
   # consider it for deletion or any required action.
   p.spawn(<def>, args)
```

ceph/utils.py
```diff
@@ -165,6 +165,33 @@ def volume_cleanup(volume, osp_cred):
...
+                log.info("volume %s %s status is %s", volume.id, volume.name, volume.state)
```
sunilkumarn417, line 177: this works, but IMO declaring a tuple list for more than one would be nice.

sunilkumarn417 (commented):
Hi @veera-raghava-reddy,

Thanks for the PR. 

I have left few comments on the code changes,

ceph/utils.py
veera-raghava-reddy: Updated the comment to docstring.

ceph/utils.py
veera-raghava-reddy: When VMs are deleted from PSI webportal, the attached volumes are moved to "available". So need to delete those Volumes

ceph/utils.py
veera-raghava-reddy: Update the states to use tuple.

ceph/utils.py
veera-raghava-reddy: Update to tuple

ceph/utils.py
veera-raghava-reddy: Not using name pattern, to clean volumes which are not attached to any known VM instances

veera-raghava-reddy:
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1612167890769/startup.log

run.py
```diff
@@ -61,6 +61,8 @@
         [--xunit-results]
   run.py --cleanup=name [--osp-cred <file>]
         [--log-level <LEVEL>]
+  run.py --clean-volumes=yes [--osp-cred <file>]
```
psathyan, line 64: IMO, the clean up of orphaned volumes can be a separate script. This script uses a generic scope to perform action whereas `run` uses a restrictive scope. That is, either it is performed against a `pattern` or `reuse`.

psathyan (changes requested):
If the intent is to run the script on a periodic bases then I would suggest having it as a separate script instead of tying it with `run.py`.

run.py
veera-raghava-reddy: Add option to run.py to keep single script to run different options.

ceph/utils.py
```diff
@@ -165,6 +165,22 @@ def volume_cleanup(volume, osp_cred):
...
+    vol_states = [ "deleting","available", "error", "unknown", "reserved" ]
+    driver = get_openstack_driver(osp_cred)
+    timeout = datetime.timedelta(seconds=timeout)
+    with parallel() as p:
+        for volume in driver.list_volumes():
+            if volume.state in vol_states:
+                log.info("volume with id:%s, name:%s has status:%s", volume.id, volume.name, volume.state)
+                p.spawn(volume_cleanup, volume, osp_cred)
+                sleep(1)
+    sleep(30)
```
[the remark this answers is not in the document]
veera-raghava-reddy: Added Sleep to avoid exiting script before multiple threads are able to trigger and clean the Volumes.

ceph/utils.py
```diff
@@ -165,6 +165,22 @@ def volume_cleanup(volume, osp_cred):
                     return 1
 
 
+def cleanup_ceph_vols(osp_cred, pattern=None, timeout=300):
+    '''
+    Added this function to cleanup stale volumes with status deleting, available, error, unknown, reserved
+    '''
+    vol_states = [ "deleting","available", "error", "unknown", "reserved" ]
+    driver = get_openstack_driver(osp_cred)
+    timeout = datetime.timedelta(seconds=timeout)
```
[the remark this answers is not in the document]
veera-raghava-reddy: removed the unused variable declarations.

run.py
veera-raghava-reddy: We can enhance to restrict to run with specific pattern or user-id. Also as a prevention Openstack does not delete in-use volumes, so risk of deleting unintended vols is minimal.

veera-raghava-reddy:
Will add clean vol as a different file instead of adding to run.py

PR #337, opened 2021-02-04
Add support for isort

psathyan:
# Description

`isort` helps us to sort and group the import modules automatically. This is inline with our automation guidelines. In this PR, we and the package and ensure it's followed at gate.

### DIFF
--- a/tests/ceph_installer/test_ceph_deploy.py
+++ b/tests/ceph_installer/test_ceph_deploy.py
@@ -1,8 +1,7 @@
 import logging
 import time
 
-from ceph.utils import keep_alive, setup_deb_repos
-from ceph.utils import setup_repos, create_ceph_conf
+from ceph.utils import create_ceph_conf, keep_alive, setup_deb_repos, setup_repos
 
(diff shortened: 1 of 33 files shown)


tests/ceph_installer/test_ceph_deploy.py
```diff
@@ -1,8 +1,7 @@
 import logging
 import time
 
-from ceph.utils import keep_alive, setup_deb_repos
-from ceph.utils import setup_repos, create_ceph_conf
+from ceph.utils import create_ceph_conf, keep_alive, setup_deb_repos, setup_repos
```
rakeshgm, line 4: I allays favoured these kinda imports [like this](https://github.com/red-hat-storage/cephci/pull/337/files#diff-d6af0459a37d985953d7040c14f53feb3b9cc9e58b543aa3c2b80256d276c5e0R29-R31). can we also get these here and other places.  I am assuming this is auto formatted.  what do you think ?

rakeshgm (approved):

PR #344, opened 2021-02-09
README: improve versions in cluster config section

ktdreyer:
Reformat the versions information in the Cluster Configuration section of the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Cluster configuration files are kept in a directory under `conf` for each ceph v
 | "Jewel"                | RH Ceph Storage 2       | `conf/jewel`    |
 | "Luminous              | RH Ceph Storage 3       | `conf/luminous` |
 | "Nautilus              | RH Ceph Storage 4       | `conf/nautilus` |
+| "Pacific"              | RH Ceph Storage 5       | `conf/pacific`  |
 
 The conf files describes the test bed configuration.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,11 +38,14 @@ your own OpenStack credentials.
...
-For luminous, configs are under `conf/luminous`.
-For nautilus, configs are under `conf/nautilus`
 
-The conf files describes the test bed configuration.
+| Ceph upstream codename | RH Ceph Storage Release | Location        |
+| ---------------------- | ----------------------- | --------------- |
+| "Jewel"                | RH Ceph Storage 2       | `conf/jewel`    |
+| "Luminous              | RH Ceph Storage 3       | `conf/luminous` |
+| "Nautilus              | RH Ceph Storage 4       | `conf/nautilus` |
+| "Pacific"              | RH Ceph Storage 5       | `conf/pacific`  |
```
psathyan, line 47: Thank you, I was thinking if it would be a good idea to have one to one mapping with RHCS versions.

rakeshgm: @psathyan. this looks good, I think it would be little difficult to map with downstream releases cause, we will have dot releases for example. `4.1`, `4.2` etc. and both of these belong to nautilus. and we are not creating dirs based on specific releases..

rakeshgm (approved):

psathyan (approved):

PR #348, opened 2021-02-09
conf: install basic packages in single yum transaction

ktdreyer:
Reduce the multiple Yum invocations to a single command so that this runs more quickly and efficiently.


psathyan (approved):
Looks good to me. Thanks

PR #351, opened 2021-02-12
use dictionary for repos

rakeshgm:
use of dictionary for repos avoids if-else statement

[Here is the log](http://magna002.ceph.redhat.com/ceph-qe-logs/rakesh/cephci_pr_logs/cephci-run-1613112266232/)

### DIFF
--- a/tests/misc_env/install_prereq.py
+++ b/tests/misc_env/install_prereq.py
@@ -159,16 +159,13 @@ def enable_rhel_rpms(ceph, distro_ver):
     Args:
         distro_ver: distro version details
     """
-    repos_7x = ["rhel-7-server-rpms", "rhel-7-server-extras-rpms"]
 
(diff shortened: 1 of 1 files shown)


psathyan (approved):
Nice one.

tests/misc_env/install_prereq.py
```diff
@@ -159,16 +159,13 @@ def enable_rhel_rpms(ceph, distro_ver):
...
+        "8": ["rhel-8-for-x86_64-appstream-rpms", "rhel-8-for-x86_64-baseos-rpms"],
+    }
 
-    repos = None
-    if not distro_ver.startswith("8"):
-        repos = repos_7x
-    else:
-        repos = repos_8x
-    for repo in repos:
+    for repo in repos.get(distro_ver[0]):
```
ktdreyer, line 168: I think we should raise `KeyError` or `NotImplementedError` here instead of calling `get()`. We want to fail hard if the caller provides something other than "7" or "8" right?

PR #365, opened 2021-02-16
Clean_volumes - update size of available VMs

veera-raghava-reddy:
update clean_ceph_vols.py to display count and capacity consumed by volumes in state "available"
python clean_ceph_vols.py 
102 Volumes in available state using storage capacity 1710 GB

Also added check to delete volumes whose name start with "ceph-" to exclude deletion of non-ci volumes.

### DIFF
--- a/clean_ceph_vols.py
+++ b/clean_ceph_vols.py
@@ -7,7 +7,6 @@ from gevent import monkey, sleep
 
 monkey.patch_all()
 import sys
-from datetime import datetime
 
 import yaml
 from docopt import docopt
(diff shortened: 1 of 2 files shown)


clean_ceph_vols.py
```diff
@@ -91,8 +87,9 @@ def cleanup_ceph_vols(osp_cred):
     with parallel() as p:
         for volume in driver.list_volumes():
             if volume.state in vol_states:
-                p.spawn(volume_cleanup, volume, driver)
-                sleep(1)
+                if volume.name[:5] in "ceph-":
```
psathyan, line 90: IMO, we could do `if volume.name.startswith('ceph-'):`

veera-raghava-reddy: modified the search cmd

clean_ceph_vols.py
```diff
@@ -64,8 +63,6 @@ def volume_cleanup(volume, driver):
     """
     print("Removing volume", volume.name)
     errors = {}
-    timeout = datetime.timedelta(seconds=200)
-    starttime = datetime.datetime.now()
     while True:
```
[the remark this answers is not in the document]
veera-raghava-reddy: Remove loop.

clean_ceph_vols.py
```diff
@@ -64,22 +63,17 @@ def volume_cleanup(volume, driver):
...
-                if datetime.datetime.now() - starttime > timeout:
-                    for vol, err in errors.items():
-                        print("Error destroying", vol, ":", err)
-                    return 1
+    try:
+        volobj = driver.ex_get_volume(volume.id)
+        driver.detach_volume(volobj)
+        driver.destroy_volume(volobj)
+    except BaseHTTPError as e:
+        print(e)
```
psathyan, line 71: As this method is performing for a single volume... do we still need to store the value in the `errors` dict? I believe it would be sufficient to have line no. 71 and remove ln 72 to 76.

clean_ceph_vols.py
```diff
@@ -64,22 +63,17 @@ def volume_cleanup(volume, driver):
...
+        errors.update({volume.name: e.message})
+        if errors:
+            for vol, err in errors.items():
+                print("Error destroying", vol, ":", err)
+            return 1
```
psathyan, line 76: It would be good if we have consistent return types.

clean_ceph_vols.py
veera-raghava-reddy: I saw that coming :-), while committing. Will update it.

clean_ceph_vols.py
veera-raghava-reddy: Noted. Will change accordingly.

psathyan (approved):
Looks good to me

PR #374, opened 2021-02-18
New Improved ceph admin modules with support of flexible orchestration options

### LINKED ISSUE #359: remove "chmod 777 /etc/ceph" for cephadm
[ceph/ceph_admin/bootstrap_mixin.py](https://github.com/red-hat-storage/cephci/blob/4513932a193de58909851a00cf19fd028e9f7224/ceph/ceph_admin/bootstrap_mixin.py) always runs `chmod 777 /etc/ceph`

Do we need to make that directory world-writable for some reason?

If we could remove this, it would make the test closer to the documented procedures for users, and it would reduce test execution time.

sunilkumarn417:
New design which adds different deployment provision and enhances flexibility.
- added provision for apply label and different placement option.
- supports cephadm orchestration command execution with more CLI arguments.
- updated sanity and deployment test suite as per the new design.
- added remove service provision.
- support to get json/yaml and other supported format CLI response.

This design brings flexibility to QE by supporting staged deployment. Starting with a small number of nodes and then
expanding the cluster. Below illustrates an example of it's strength.

```
tests:
  ...
  - test:
      name: cluster deployment
      desc: Execute the cluster deployment workflow.
      module: test_deployment.py
      polarion-id:
      config:
        steps:
          - config:
              command: bootstrap
              service: cephadm
              args:
                mon-ip: node1
          - config:
              command: add_hosts
              service: host
              args:
                nodes: [node1, node2, node3]
                attach_ip_address: true
                labels: apply-all-labels
          - config:
              command: apply
              service: osd
              args:
                all-available-devices: true
      destroy-cluster: false
      abort-on-fail: true
  - test:
      desc: Perform base line operations
      ...
  ...
  - test:
      name: adding 2 more hosts
      desc: Adding node[4,5]
      module: test_development.py
      config:
        steps:
          - config:
               command: add_hosts
                service: host
                args:
                  nodes: [node3, node4]
                  attach_ip_address: true
                  labels: apply-all-labels
          -  config:
              command: apply
              service: osd
              args:
                 all-available-devices: true
                 unmanaged: true
      destroy-cluster: false
      abort-on-fail: true
  - test
  ...
```

**Test results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1613639908753/result.html

Closes #359


rakeshgm:
phew,  this is a big one :)

sunilkumarn417:
Latest results: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1613649237345

- Fixed remote_file issue.

psathyan:
> phew, this is a big one :)

Yes, we couldn't break into small pieces as the required was to stand up a minimal cluster for the FGs.

ceph/ceph_admin/apply.py
```diff
@@ -1,85 +1,122 @@
...
-from ceph.ceph import ResourcesNotFoundError
+from ceph.utils import get_nodes_by_id
 
-logger = logging.getLogger(__name__)
+from .common import config_dict_to_string
+from .typing_ import ServiceProtocol
 
 
-class ServiceRemoveFailure(Exception):
+class ServiceApplyFailure(Exception):
```
psathyan, line 18: IMO, the name of the exception class could be `OrchApplyFailed`.

ceph/ceph_admin/apply.py
```diff
@@ -1,85 +1,122 @@
 """
+Represent the ceph orch CLI action 'apply'.
+
 Module to deploy ceph role service(s) using orchestration command
 "ceph orch apply <role> [options] --placement '<placements>' "
 
-this module inherited where service deployed using "apply" operation.
-and also validation using orchestration process list response
+This module inherited where service deployed using "apply" operation and also
```
psathyan, line 7: I believe we could reword here.
```
This is a mixin object and can be applied to the supported service classes.
```

ceph/ceph_admin/apply.py
```diff
@@ -1,85 +1,122 @@
...
+class ServiceApplyFailure(Exception):
     pass
 
 
-class Apply:
-
-    apply_cmd = ["ceph", "orch", "apply"]
-    remove_cmd = ["ceph", "orch", "rm"]
+class ApplyMixin:
+    """Add apply support to the base class."""
```
psathyan, line 23: I believe it should be `Add apply command support to child object.`

ceph/ceph_admin/bootstrap.py
```diff
@@ -0,0 +1,91 @@
...
+        )
+        cmd += custom_image_args.format(
+            user=cdn_cred.get("username"),
+            password=cdn_cred.get("password"),
+        )
+
+        # To be generic, the mon-ip contains the global node name. Here, we replace the
+        # name with the IP address. The replacement allows us to be inline with the
+        # CLI option.
+        mon_node = args.pop("mon-ip")
```
psathyan, line 70: Should we allow it be optional and take the `installer-node ip address` if the value is not provided?

ceph/ceph_admin/host.py
```diff
@@ -0,0 +1,386 @@
...
+            label_set = set(ceph_node.role.role_list)
+            _labels = list(label_set)
+
+        cmd.extend(["host", "add", ceph_node.shortname])
+
+        if attach_address:
+            cmd.append(ceph_node.ip_address)
+
+        if _labels:
+            # To fill mandate <address> argument, In case attach_address is False
```
psathyan, line 82: Shall we raise a usability issue in the upstream. Having positional arguments as empty does not make sense.

ceph/ceph_admin/host.py
```diff
@@ -0,0 +1,386 @@
...
+            if not attach_address:
+                cmd.append("''")
+
+            cmd += _labels
+
+        logger.info(
```
psathyan, line 88: Do we need this log statement? There is a log statement in `exec_command` already.

ceph/ceph_admin/iscsi.py
```diff
@@ -1,77 +1,50 @@
...
-        )
+        pos_args = config.pop("pos_args")
 
-        Apply.apply(
-            self,
-            role=self.__ROLE,
-            command=cmd,
-            placements=host_placement,
-        )
+        if isinstance(pos_args[-1], list):
```
psathyan, line 45: Do we need to make this dynamic? That is pull the `CephNode` for the node names and then use the associated IP address.

Maybe we can take it up as an RFE.

ceph/ceph_admin/shell.py
```diff
@@ -0,0 +1,44 @@
...
+        out, err = self.installer.exec_command(
+            sudo=True,
+            cmd=" ".join(cmd),
+            timeout=timeout,
+            check_ec=check_status,
+        )
+
+        out = out.read().decode().strip()
+        err = err.read().decode().strip()
+        LOG.debug(out)
```
psathyan, line 43: Do we need this statement? I believe `exec_command` already adds a statement at `info` level

psathyan (commented):
Thanks for doing all the heavy lifting.

Minor comments.

ceph/ceph_admin/bootstrap.py
sunilkumarn417: However currently node cannot be other than installer node.

ceph/ceph_admin/host.py
sunilkumarn417: we can keep this, just to make sure about config provided

ceph/ceph_admin/shell.py
sunilkumarn417: I think this is needed, to validate CLI response(for debug purpose).
However currently all runs going with info log-level, which doesn't include debug log records.

ceph/ceph_admin/apply.py
sunilkumarn417: added

ceph/ceph_admin/apply.py
sunilkumarn417: added `OrchApplyServiceFailure`.

psathyan (approved):
Looks good to me

PR #383, opened 2021-02-23
Cephadm: Add NFS support

sunilkumarn417:

- Added NFS support.

**Test results**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1614095651953/result.html

### DIFF
--- a/tests/cephadm/test_nfs.py
+++ b/tests/cephadm/test_nfs.py
@@ -1,0 +1,32 @@
+import logging
+
+from ceph.ceph_admin.common import fetch_method
+from ceph.ceph_admin.nfs import NFS
+
+log = logging.getLogger(__name__)
+
+
+def run(ceph_cluster, **kw):
+    """
+    Ceph-admin module to manage NFS service
+
+    Args:
+        ceph_cluster (ceph.ceph.Ceph): Ceph cluster object
+        kw: test data
+
+    check ceph.ceph_admin.nfs for test config
+    """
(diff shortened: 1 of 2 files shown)


psathyan (approved):
At some point of time, we may want to think about restricting the scope of `cephadm.yaml` test suite and create a new one. Otherwise we might end up having a big test suite and have difficulty maintaining it.

sunilkumarn417:
> At some point of time, we may want to think about restricting the scope of `cephadm.yaml` test suite and create a new one. Otherwise we might end up having a big test suite and have difficulty maintaining it.
I Strongly agree on this and we should address which tests to be part of sanity and feature test-suite.

tests/cephadm/test_nfs.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    Ceph-admin module to manage NFS service
+
+    Args:
+        ceph_cluster (ceph.ceph.Ceph): Ceph cluster object
+        kw: test data
+
+    check ceph.ceph_admin.nfs for test config
+    """
+    log.info("Running Ceph-admin Rados-GW( RGW ) test")
```
ktdreyer, line 19: This is a mistaken copy-and-paste from `tests/cephadm/test_rgw.py`, mind updating this?

sunilkumarn417: Addressed in https://github.com/red-hat-storage/cephci/pull/385

PR #385, opened 2021-02-24
Cephadm/Bootstrap: initial-dashboard user/password support

sunilkumarn417:
- Added initial-dashboard-user, initial-dashboard-password bootstrap support options.
- Added dashboard login validations by calling API call.
- Moved out validations outside bootstrap class.

**Results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1614174384742/Cephadm_Bootstrap_0.log

### DIFF
--- a/ceph/ceph_admin/bootstrap.py
+++ b/ceph/ceph_admin/bootstrap.py
@@ -1,8 +1,11 @@
 """Module that allows QE to interface with cephadm bootstrap CLI."""
 import json
 import logging
+import re
 import tempfile
 from typing import Dict
+
+import requests
 
 from ceph.ceph import ResourceNotFoundError
 from utility.utils import get_cephci_config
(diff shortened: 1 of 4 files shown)


ceph/ceph_admin/bootstrap.py
```diff
@@ -52,22 +55,124 @@ def construct_registry(cls, registry: str, json_file: bool = False):
...
+            host = host.replace(host_, node.ip_address)
+            break
+    else:
+        return False
+
+    data = json.dumps({"password": passwd_, "username": user_})
+    url_ = f"{host}api/auth"
+
+    session = requests.Session()
+    session.headers = {}
```
ktdreyer, line 117: Can we collapse the assignment and `.update()` call here?

```suggestion
    session.headers = {
            "accept": "application/vnd.ceph.api.v1.0+json",
            "content-type": "application/json",
        }
```

ceph/ceph_admin/bootstrap.py
```diff
@@ -52,22 +55,124 @@ def construct_registry(cls, registry: str, json_file: bool = False):
...
+    session.headers.update(
+        {
+            "accept": "application/vnd.ceph.api.v1.0+json",
+            "content-type": "application/json",
         }
-        return verify_map.get(key)
+    )
+
+    resp = session.post(url_, data=data, verify=False)
+    return resp.ok
```
ktdreyer, line 126: What do you think about raising here if `resp.ok` is False? Then we'll get more information about the exact error message from the server. Like

```suggestion
    resp.raise_for_status()
    return resp.ok
```

ceph/ceph_admin/bootstrap.py
```diff
@@ -52,22 +55,124 @@ def construct_registry(cls, registry: str, json_file: bool = False):
...
+    Returns:
+        boolean
+    """
+    _, _, db_log_part = out.partition("Ceph Dashboard is now available at:")
+
+    if not db_log_part:
+        return False
+
+    def parse(string):
+        str_ = r"%s:\s(.*)" % string
```
ktdreyer, line 93:
```suggestion
        str_ = r"%s:\s(.+)" % string
```

`*` captures zero or more characters, and `+` captures one or more characters, and we always expect to find one or more here.

ceph/ceph_admin/bootstrap.py
```diff
@@ -52,22 +55,124 @@ def construct_registry(cls, registry: str, json_file: bool = False):
...
+        out: bootstrap response log
+        user: dashboard user
+        password: dashbard password
+
+    Returns:
+        boolean
+    """
+    _, _, db_log_part = out.partition("Ceph Dashboard is now available at:")
+
+    if not db_log_part:
```
ktdreyer, line 89: Should we `raise` here, or `log.error()` here?

ceph/ceph_admin/bootstrap.py
sunilkumarn417: AssertionError raised at called line of code, So just returning Boolean value

ceph/ceph_admin/bootstrap.py
sunilkumarn417: AssertionError raised based on boolean at called place, So just returning Boolean value.

ceph/ceph_admin/bootstrap.py
sunilkumarn417: done

ceph/ceph_admin/bootstrap.py
sunilkumarn417: done

ceph/ceph_admin/bootstrap.py
sunilkumarn417: added error level log record.

ceph/ceph_admin/bootstrap.py
```diff
@@ -52,22 +55,122 @@ def construct_registry(cls, registry: str, json_file: bool = False):
...
+    def parse(string):
+        str_ = r"%s:\s(.+)" % string
+        return re.search(str_, db_log_part).group(1)
+
+    host = parse("URL")
+    user_ = parse("User")
+    passwd_ = parse("Password")
+
+    if user:
+        assert user == user_, f"user {user} did not match"
```
ktdreyer, line 102: Thanks for adding these explanation strings to the assertions!

ceph/ceph_admin/bootstrap.py
ktdreyer: I mean we don't get to see any detailed information from the server here if `resp` was not `ok`. It's just `False`, and we don't get more information about the HTTP response code or anything else.

sunilkumarn417: added log record which displays response.text and response status code.

ceph/ceph_admin/bootstrap.py
```diff
@@ -52,22 +55,123 @@ def construct_registry(cls, registry: str, json_file: bool = False):
...
+    if password:
+        assert password == passwd_, f"password {password} did not match"
+
+    host_ = re.search(r"https?:[/]{2}(.*):8443", host).group(1)
```
psathyan, line 106: Could we avoid the port number? In the scenario where the port number is customized or modified, this would fail.

ceph/ceph_admin/bootstrap.py
```diff
@@ -52,22 +55,123 @@ def construct_registry(cls, registry: str, json_file: bool = False):
...
+    for node in cls.cluster.get_nodes():
+        if host_ in node.shortname:
```
psathyan, line 108: What is the probability of this check failing when we use `--allow-fqdn-hostname`?

ceph/ceph_admin/bootstrap.py
```diff
@@ -52,22 +55,123 @@ def construct_registry(cls, registry: str, json_file: bool = False):
...
+            host = host.replace(host_, node.ip_address)
+            break
+    else:
+        return False
+
+    data = json.dumps({"password": passwd_, "username": user_})
+    url_ = f"{host}api/auth"
```
psathyan, line 115: Could we ensure that the url has a trailing `/`?

ceph/ceph_admin/bootstrap.py
```diff
@@ -143,29 +249,17 @@ def bootstrap(self: CephAdmProtocol, config: Dict) -> None:
...
         # Verification of arguments
-
         args = self.config.get("args", {})
         for key, value in args.items():
-            func = self.get_method(key)
+            func = fetch_method(key)
             if not func:
                 continue
-            if not func(value):
+            if not func(self, value, out + err):
```
psathyan, line 264: Would it possible to add a note on why we are concatenating `out + err`

ceph/ceph_admin/bootstrap.py
sunilkumarn417: done

ceph/ceph_admin/bootstrap.py
sunilkumarn417: done

ceph/ceph_admin/bootstrap.py
sunilkumarn417: with current CephVMNode, there is no issue. 
However It will be problem if,
- hostname and FQDN are different.
- hostname is not substring of FQDN.

psathyan (approved):
Looks good to me

PR #392, opened 2021-03-01
cephadm: validation for apply service with label

sunilkumarn417:
- validation for apply service with label.
- updated inventory configuration.
- Added validations for skip-monitoring-stack option.
- added check cluster state to know ceph state after after action, just for sanity purpose.

**Results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1614613199281/Cephadm_Bootstrap_0.log
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1614592538795/

### DIFF
--- a/ceph/ceph_admin/__init__.py
+++ b/ceph/ceph_admin/__init__.py
@@ -124,3 +124,14 @@ class CephAdmin(BootstrapMixin, ShellMixin):
             self.installer.exec_command(sudo=True, cmd="yum update -y cephadm")
 
         self.installer.exec_command(cmd="rpm -qa | grep cephadm")
+
+    def get_cluster_state(self, commands):
+        """
+        Display cluster state by executing commands provided
+        Just used for sanity.
+        """
+        for cmd in commands:
+            out, err = self.shell(args=[cmd])
+            logger.info(out)
+            if err:
+                logger.error(err)
--- a/ceph/ceph_admin/orch.py
+++ b/ceph/ceph_admin/orch.py
@@ -24,6 +24,21 @@ class Orch(LSMixin, PSMixin, RemoveMixin, CephCLI):
 
     direct_calls = ["ls", "ps"]
 
+    def get_hosts_by_label(self, label: str):
+        """
+        Fetch host object by label attached to it.
+        Args:
+            label: name of the label
+        Returns:
+            hosts
+        """
+        out, _ = self.shell(args=["ceph", "orch", "host", "ls", "--format=json"])
+        hosts = list()
+        for node in loads(out):
+            if label in node.get("labels"):
+                hosts.append(node)
+        return hosts
+
     def check_service_exists(
         self, service_name: str, ids: List[str], timeout: int = 300, interval: int = 5
     ) -> bool:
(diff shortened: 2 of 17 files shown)


ceph/ceph_admin/orch.py
```diff
@@ -24,6 +24,21 @@ class Orch(LSMixin, PSMixin, RemoveMixin, CephCLI):
...
+            label: name of the label
+        Returns:
+            hosts
+        """
+        out, _ = self.shell(args=["ceph", "orch", "host", "ls", "--format=json"])
+        hosts = list()
+        for node in loads(out):
+            if label in node.get("labels"):
+                hosts.append(node)
+        return hosts
```
psathyan, line 40: `return [node for node in loads(out) if label in node.get("labels")]`

psathyan (commented):
Minor one...

ceph/ceph_admin/orch.py
sunilkumarn417: it would result in NameError.

sunilkumarn417: it might be pdb mode, let me see in interactive mode.

sunilkumarn417:
**Latest Results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1614682572322/result.html

tests/ceph_installer/test_cephadm.py
```diff
@@ -76,11 +76,11 @@ def run(ceph_cluster, **kw):
     log.info("Executing %s %s" % (service, command))
 
     if command in CephAdmin.direct_calls:
-        cephadm = CephAdmin(cluster=ceph_cluster, **config)
-        method = fetch_method(cephadm, command)
+        instance = CephAdmin(cluster=ceph_cluster, **config)
```
psathyan, line 79: I liked `cephadm`... this change is also okay

suites/pacific/cephadm/cephadm.yaml
```diff
@@ -22,6 +22,11 @@ tests:
           initial-dashboard-user: admin123
           initial-dashboard-password: admin123
           fsid: f64f341c-655d-11eb-8778-fa163e914bcc
+        check_cluster_state:                    # to view ceph status only
+          - "ceph status"
```
psathyan, line 26: Instead of `check_cluster_state`, should be change it to `get_cluster_details`? If they are the set of standard commands that we would be executing... would it be a good idea to support `bool`?

We can support both `bool` & `list` if it makes sense.

sunilkumarn417: I think it would good if we keep it list of CLI commands, because we may need mapping in case of Boolean and conditional check at execution.

ceph/ceph_admin/__init__.py
```diff
@@ -124,3 +124,13 @@ def install(self, **kwargs: Dict) -> None:
...
+
+    def get_cluster_state(self, commands):
+        """
+        Display cluster state by executing commands provided
+        Just used for sanity.
+        """
+        for cmd in commands:
+            out, err = self.shell(args=[cmd])
+            logger.info(out)
+            logger.error(err)
```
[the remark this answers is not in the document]
sunilkumarn417: sure.

suites/pacific/cephadm/cephadm.yaml
sunilkumarn417: `get_cluster_details` looks good.

psathyan (approved):
Looks good to me

PR #394, opened 2021-03-02
Fixing broken base_url links in rhbuild

psathyan:
# Description

Some of the `base_url` links are broken in `rhbuild.yaml`, in this PR, moving them to use the released version links that have a longer shelf life.

Also, aligning '5.0' to use the OS version as suffix 

### DIFF
--- a/rhbuild.yaml
+++ b/rhbuild.yaml
@@ -4,16 +4,16 @@ ceph:
     name: jewel
     composes:
       'latest':
-        base_url: 'http://download.eng.bos.redhat.com/composes/auto/ceph-2-rhel-7/latest-RHCEPH-2-RHEL-7/'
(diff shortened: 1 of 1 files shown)


rakeshgm:
we need the links that you are removing, we don't want to test release versions, we need these links for testing unreleased versions. we still are going to get 4.x builds.

rakeshgm:
or better create another YAML file for GA versions.

rakeshgm:
> Thanks, I'll revert the modifications done in the `rhbuild.yaml` file. In case of upgrades and long running tests, I believe we would need to have General Availability (GA) links along with development build pointers. Instead of creating a new file, I would like to introduce `ga_url` in `rhbuild.yaml` along with the CLI option `use_ga_url`. This would allow us to use the released version for N-1 along with the latest build of N release for our upgrade testing.
> 
> Could you let me know your thoughts on the approach?

yeah. using `ga_url` sounds good.

psathyan:
Closing this PR as the requirement needs more thought/design

PR #399, opened 2021-03-03
adds --hotfix-repo support to cephadm tests

### LINKED ISSUE #369: --hotfix-repo does not work for cephadm
cephci ignores the `--hotfix-repo` option when installing the `cephadm` RPM on nodes. This makes it difficult to test cephadm fixes with the `--hotfix-repo` argument.

andrewschoen:
Fixes #369

### DIFF
--- a/ceph/ceph_admin/__init__.py
+++ b/ceph/ceph_admin/__init__.py
@@ -89,7 +89,7 @@ class CephAdmin(BootstrapMixin, ShellMixin):
     def set_tool_repo(self):
         """Add the given repo on every node part of the cluster."""
         if not self.config.get("hotfix_repo"):
-            base_url = self.config.get("base_url")
+            base_url = self.config["base_url"]
             if not base_url.endswith("/"):
                 base_url += "/"
             cmd = f"yum-config-manager --add-repo {base_url}compose/Tools/x86_64/os/"
--- a/ceph/ceph_admin/__init__.py
+++ b/ceph/ceph_admin/__init__.py
@@ -88,12 +88,13 @@ class CephAdmin(BootstrapMixin, ShellMixin):
 
     def set_tool_repo(self):
         """Add the given repo on every node part of the cluster."""
-        base_url = self.config.get("base_url")
-        if not base_url.endswith("/"):
-            base_url += "/"
-        cmd = f"yum-config-manager --add-repo {base_url}compose/Tools/x86_64/os/"
-        for node in self.cluster.get_nodes():
-            node.exec_command(sudo=True, cmd=cmd)
+        if not self.config.get("hotfix_repo"):
+            base_url = self.config.get("base_url")
+            if not base_url.endswith("/"):
+                base_url += "/"
+            cmd = f"yum-config-manager --add-repo {base_url}compose/Tools/x86_64/os/"
+            for node in self.cluster.get_nodes():
+                node.exec_command(sudo=True, cmd=cmd)
 
     def install(self, **kwargs: Dict) -> None:
         """
--- a/tests/misc_env/install_prereq.py
+++ b/tests/misc_env/install_prereq.py
@@ -36,17 +36,34 @@ def run(**kw):
     ceph_nodes = kw.get("ceph_nodes")
     # skip subscription manager if testing beta RHEL
     config = kw.get("config")
+    ceph_cluster = kw.get("ceph_cluster")
     skip_subscription = config.get("skip_subscription", False)
     enable_eus = config.get("enable_eus", False)
     repo = config.get("add-repo", False)
+    hotfix_repo = config.get("hotfix_repo", False)
     rhbuild = config.get("rhbuild")
+    ubuntu_repo = config.get("ubuntu_repo", None)
+    base_url = config.get("base_url", None)
+    installer_url = config.get("installer_url", None)
 
     with parallel() as p:
         for ceph in ceph_nodes:
             p.spawn(
-                install_prereq, ceph, 1800, skip_subscription, repo, rhbuild, enable_eus
+                install_prereq,
+                ceph,
+                1800,
+                skip_subscription,
+                repo,
+                rhbuild,
+                enable_eus,
             )
             time.sleep(20)
+
+    ceph_cluster.ansible_config = dict(dedicated_devices=[])
+    ceph_cluster.setup_packages(
+        base_url, hotfix_repo, installer_url, ubuntu_repo, rhbuild
+    )
+
     return 0
 
 


ceph/ceph_admin/__init__.py
```diff
@@ -88,12 +88,13 @@ def distribute_cephadm_gen_pub_key(self, nodes=None):
...
     def set_tool_repo(self):
         """Add the given repo on every node part of the cluster."""
-        base_url = self.config.get("base_url")
-        if not base_url.endswith("/"):
-            base_url += "/"
-        cmd = f"yum-config-manager --add-repo {base_url}compose/Tools/x86_64/os/"
-        for node in self.cluster.get_nodes():
-            node.exec_command(sudo=True, cmd=cmd)
+        if not self.config.get("hotfix_repo"):
+            base_url = self.config.get("base_url")
```
psathyan, line 92: Would you be able to modify this to `base_url = self.config['base_url']`? `base_url` is a required parameter for building out the cluster... the change would help in failing fast.

tests/misc_env/install_prereq.py
```diff
@@ -114,6 +125,19 @@ def setup_addition_repo(ceph, repo):
     ceph.exec_command(sudo=True, cmd="yum update metadata", check_ec=False)
 
 
+def setup_hotfix_repo(ceph, hotfix_repo):
```
psathyan, line 128: Could you add `docstrings` for this method?

ceph/ceph_admin/__init__.py
```diff
@@ -88,12 +88,13 @@ def distribute_cephadm_gen_pub_key(self, nodes=None):
 
     def set_tool_repo(self):
         """Add the given repo on every node part of the cluster."""
-        base_url = self.config.get("base_url")
-        if not base_url.endswith("/"):
-            base_url += "/"
-        cmd = f"yum-config-manager --add-repo {base_url}compose/Tools/x86_64/os/"
-        for node in self.cluster.get_nodes():
-            node.exec_command(sudo=True, cmd=cmd)
+        if not self.config.get("hotfix_repo"):
```
psathyan, line 91: Is there a possibility of cephadm package not being part of the hotfix-repo? If yes, I believe we should be adding the base url and hot fix repo.

psathyan: After discussing with Rakesh, I understand that `hotfix_repo` would ideally have all the required packages.

tests/misc_env/install_prereq.py
andrewschoen: Sure thing, I've added a docstring.

ceph/ceph_admin/__init__.py
andrewschoen: Ok, I've made that change.

psathyan:
Thanks... I have a concern (it could be misplaced) that the changes done as part of `install_prereq` would be a duplicated when we use `test_ansible.py` module... in this module there is a step that performs a similar operation. Ideally, we should have called the same method `ceph_cluster.setup_packages(...)` in CephAdm.

@sunilkumarn417 @rakeshgm could you way your thoughts? I am not sure why it's done there in the first place.

PS: I don't think it would break... also `enable_eus` is also done here

andrewschoen:
> Thanks... I have a concern (it could be misplaced) that the changes done as part of `install_prereq` would be a duplicated when we use `test_ansible.py` module... in this module there is a step that performs a similar operation. Ideally, we should have called the same method `ceph_cluster.setup_packages(...)` in CephAdm.

Yeah, I noticed that `ceph_cluster.setup_packages` was doing a similar operation. It wasn't already being used for the cephadm tests though, so I figured that was purposeful. That method is doing quite a bit more than what I'm doing in `install_prereq`. You're right that in the ceph-ansible tests the hotfix repo would be added twice with this change though. If anything maybe the call to `ceph_cluster.setup_packages` should only exist in `install_prereq` if we wanted to consolidate everything.

rakeshgm:
I think `Ceph` class is not directly exposed in run.py but the object is in [L182)](https://github.com/red-hat-storage/cephci/blob/master/run.py#L182) and ceph objects are used in test cases. `setup_packages` is also used in other test cases like upgrades and roll-over tests. we could try to  do call in `install_prereq` since ceph object is ready to be used.

rakeshgm:
I think Ideally setting up `hotfix_repo` should be in `install_prereq` instead of being called in test case.And I think setup_packages was being called in `test_ansible`  because probably we were being very optimistic that `ceph-ansible` might be the only installer and now _cometh_ ceph-adm :)

andrewschoen:
> I think Ideally setting up `hotfix_repo` should be in `install_prereq` instead of being called in test case.And I think setup_packages was being called in `test_ansible` because probably we were being very optimistic that `ceph-ansible` might be the only installer and now _cometh_ ceph-adm :)

I agree that this setup should happen in `install_prereq`. In looking at this further there seems to be quite a bit of duplication between `ceph.Ceph` and `install_prereq`. I'm wondering if all the code in `ceph.Ceph` that handles package setup should eventually be deprecated and moved to `install_prereq`. That seems like a large refactor though.

What specifically would you like to see changed for this PR? I can try to call `setup_packages` here and remove `setup_hotfix_repo`, but honestly I think eventually moving away from using `ceph.Ceph` for package setup is a good idea.

andrewschoen:
> I think calling setup_packages here should be fine for now. I think Depricating ceph.Ceph or moving to install prereq might too disruptive.

Ok, I'll make that change and do some tests. I agree that it could be potentially very disruptive. I just wanted to point out that I'm already seeing quite a bit of duplication between setup_packages and install_prereq.

ceph/ceph_admin/__init__.py
ktdreyer: I think we should always enable `Tools` along with the hotfix repo for the following reasons:
- Customers will already have `rhceph-5-tools-for-rhel-8-x86_64-rpms` enabled on their node(s), and we might as well test in a way that matches what a customer will have.
- It's true that `cephadm` has no non-RHEL dependencies today, so it's probably ok to skip adding `Tools` in that particular case, but we're also planning to ship `ceph-ansible` in `Tools`, and `ceph-ansible` does require some dependencies that are not in RHEL (the deps are only in `Tools`).

ktdreyer:
Adding a hotfix repo twice on RH Ceph Storage 3 and 4 _should_ introduce no behavior change. The repo content will be identical, and DNF should handle it ok.

Long term, eventually we could refactor that so RH Ceph Storage 3, 4, and 5 all use the same codepaths to be faster and simpler.

ceph/ceph_admin/__init__.py
andrewschoen: The issue I ran into was that when I didn't do this the version of `cephadm` from the `Tools` repo was used, not the one from `hotfix_repo`. Perhaps I did something wrong though.

ktdreyer: On Monday I rebased the downstream packages from `cephadm-16.1.0-100.el8cp` to `cephadm-16.1.0-486.el8cp`. 
Maybe `16.1.0-486.el8cp` in `Tools` was higher than the hotfix repo's cephadm Version-Release?

andrewschoen: That's exactly what happened, the build I'm testing with is a `16.1.0-100` build. Isn't the point of ``--hotfix-repo`` to be able to test with ceph versions that are potentially lower than the latest available build or release though? I think for our use case we'd always want ``--hotfix-repo`` to take precedence. This is how it works when testing rpms and using ``--hotfix-repo``, iirc.

andrewschoen:
@rakeshgm I was able to get `setup_packages` to work here. It is pretty tightly coupled with ansible config though, which is why I imagine `setup_packages` is called in the tasks and not here initially. I had to fake a bit of the ansible config to get it to run. Another other possible concern here is that inside `setup_package` it loops over all the nodes, so it can't be run inside of `install_prereq` directly.

If you like this approach I can get the commits cleaned up.

sunilkumarn417:
> Thanks... I have a concern (it could be misplaced) that the changes done as part of `install_prereq` would be a duplicated when we use `test_ansible.py` module... in this module there is a step that performs a similar operation. Ideally, we should have called the same method `ceph_cluster.setup_packages(...)` in CephAdm.
> 
> @sunilkumarn417 @rakeshgm could you weigh in your thoughts? I am not sure why it's done there in the first place.
> 
> PS: I don't think it would break... also `enable_eus` is also done here. Also, the impact is with 4.x versions and not with 5.x

From the little knowledge I have about Hot-Fix, I do see below below things for application of hot-fix.
- At first, when we want to enable and what are the things we need to consider in for application of hot-fix. considering this point I think we may have below questions,
     - Having hot-fix repo enabled before ceph deployment (need to check things what should get overridden for example., base_url), which might be needed at install test step itself.
     - Second one, apply hotfix post deployment/install or wherever it might be needed at certain test step.
     - What if hot-fix has changes only in Tools(for 5.0 cephadm), but not with ceph images and vice-versa.

IMO, this could be handled in ceph building modules like test_ansible.py, test_cephadm.py rather than at `pre-req` task. I think this also ensures `ceph.ceph` more generic and not tightly coupled ansible or cephadm.

rakeshgm:
> @rakeshgm I was able to get `setup_packages` to work here. It is pretty tightly coupled with ansible config though, which is why I imagine `setup_packages` is called in the tasks and not here initially. I had to fake a bit of the ansible config to get it to run. Another other possible concern here is that inside `setup_package` it loops over all the nodes, so it can't be run inside of `install_prereq` directly.
> 
> If you like this approach I can get the commits cleaned up.

@andrewschoen well the change looks correct, I guess some of parameters are unavoidable. I am fine with the change for now.

andrewschoen:
@rakeshgm ok, I've got the commits cleaned up. Is there anything else needed here? Thanks!

rakeshgm (approved):

psathyan (approved):

PR #400, opened 2021-03-04
[Interop] Fixing rados_utils.py: Fix for osd wait method

pdhiran:
Corrected the method that waits for 120 seconds for the OSD to go down.

### DIFF
--- a/ceph/rados_utils.py
+++ b/ceph/rados_utils.py
@@ -208,22 +208,26 @@ class RadosHelper:
             if osd_id == osd["osd"]:
                 return osd["up"]
 
-    def check_osd_up(self, osd_id):
+    def wait_until_osd_state(self, osd_id, down=False) -> bool:
         """
-        Checks if the given osd is up and if the osd is down waits for 120 seconds for the same to come up
+        Checks the state of given osd and waits for 120 seconds for the same state to be achieved
         Args:
             osd_id: ID of the osd to be checked
+            down: If True, Specifies that the OSD should be down
 
-        Returns: 1 if up, 0 if down
+        Returns: True -> Pass, False -> Fail
 
         """
         end_time = datetime.datetime.now() + datetime.timedelta(seconds=120)
         while end_time > datetime.datetime.now():
-            status = self.is_up(osd_id)
-            if status:
-                return 1
             time.sleep(5)
-        return 0
+            osd_is_up = self.is_up(osd_id)
+            if osd_is_up and not down:
+                return True
+            if down and not osd_is_up:
+                return True
+        print("Desired OSD state not achieved after 120 seconds")
+        return False
 
     def revive_osd(self, osd_node, osd_service):
         """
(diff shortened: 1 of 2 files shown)


pdhiran:
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1614891674199/
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1614891674199/result.html

ceph/rados_utils.py
```diff
@@ -221,9 +221,11 @@ def check_osd_up(self, osd_id):
         while end_time > datetime.datetime.now():
             status = self.is_up(osd_id)
             if status:
-                return 1
-            time.sleep(5)
-        return 0
+                time.sleep(5)
```
psathyan, line 224: The method returns true when the OSD is up... but this change returns true when it is down. We could fix the return code to True (or 0) when OSD is up. 

```
while end_time > datetime.datetime.now():
  if self.is_up(osd_id): return True
  time.sleep(5)

return False
```

psathyan: For interop, can we look at the code where the check is for down and negate at that location.

pdhiran: The method returns true when OSD is up. ( after waiting for 120 seconds ), and returns false as soon as OSD goes down. This is expected of the method. Do we want to change the functionality?

psathyan: My bad... sorry for the misdirect. I got confused with 1 being `False`. Do we want to wait for `120` seconds to return the OSD is up?

pdhiran: We wait for 120 seconds to return when OSD is down. We stay in the while loop until the said OSD is still up .

ceph/rados_utils.py
```diff
@@ -219,10 +219,27 @@ def check_osd_up(self, osd_id):
...
+
+        Returns: 1 if up, 0 if down
+
+        """
+        end_time = datetime.datetime.now() + datetime.timedelta(seconds=120)
+        while end_time > datetime.datetime.now():
+            time.sleep(5)
+            if self.is_up(osd_id):
+                print("OSD is up")
+                return 1
```
psathyan, line 242: IMHO, we can achieve this with a single method `def wait_until(self, osd, up=True) -> bool:` as the difference is the state of OSD daemon.  For example
```
def wait_until_osd_state(self, osd, down=True) -> bool:
  end_time = datetime.datetime.now() + datetime.timedelta(seconds=120)
  while end_time > datetime.datetime.now():
    time.sleep(5)
     osd_is_up = self.is_up(osd_id):
     if osd_is_up and not down: return True
     if down and not osd_is_up: return True

  return False
```

pdhiran: Added a single method.

pdhiran:
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1614927542498/

http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1614927542498/result.html

psathyan (approved):
Looks good to me

PR #407, opened 2021-03-10
Adding pipeline script for tier-0 job

psathyan:
# Description

A new approach of executing the test suites based on their classifications from our central Jenkins server. In this approach, the job configurations are defined in our internal repo. In this PR,

- the required CLI arguments are derived from the CI_MESSAGE
- the test suites are executed in parallel
- the environment is cleaned

[RHCS Job View](https://storage-jenkins-csb-ceph.cloud.paas.psi.redhat.com/view/RHCS%205%20QE/)

### DIFF
--- a/pipeline/5/Jenkinsfile-tier-0.groovy
+++ b/pipeline/5/Jenkinsfile-tier-0.groovy
@@ -1,0 +1,70 @@
+/*
+    Pipeline script for executing Tier 0 test suites for RH Ceph 5.0.
+*/
+// Global variables section
+
+def nodeName = "centos-7"
+def cephVersion = "pacific"
+def sharedLib
+def testStages = ['deploy': {
+                    stage('Deployment suite') {
+                        script {
+                            withEnv([
+                                "sutVMConf=conf/inventory/rhel-8.3-server-x86_64.yaml",
+                                "sutConf=conf/${cephVersion}/cephadm/sanity-cephadm.yaml",
+                                "testSuite=suites/${cephVersion}/cephadm/cephadm_deploy.yaml",
+                                "addnArgs=--post-results --log-level DEBUG"
+                            ]) {
+                                sharedLib.runTestSuite()
+                            }
+                        }
+                    }
+                 }, 'upgrade': {
+
+                    stage('Upgrade suite') {
+                        println "Work in progress"
+                    }
+
+                 }]
+
+// Pipeline script entry point
+
+node(nodeName) {
+
+    timeout(unit: "MINUTES", time: 30) {
+        stage('Install prereq') {
+            checkout([
+                $class: 'GitSCM',
+                branches: [[name: '*/master']],
+                doGenerateSubmoduleConfigurations: false,
+                extensions: [[
+                    $class: 'SubmoduleOption',
+                    disableSubmodules: false,
+                    parentCredentials: false,
+                    recursiveSubmodules: true,
+                    reference: '',
+                    trackingSubmodules: false
+                ]],
+                submoduleCfg: [],
+                userRemoteConfigs: [[url: 'https://github.com/red-hat-storage/cephci.git']]
+            ])
+            script {
+                sharedLib = load("${env.WORKSPACE}/pipeline/vars/common.groovy")
+                sharedLib.prepareNode()
+            }
+        }
+    }
+
+    timeout(unit: "MINUTES", time: 60) {
+        parallel testStages
+    }
+
+    stage('Publish Results') {
+        script {
+            sharedLib.sendEMail("Tier-0")
+            sharedLib.postLatestCompose()
+        }
+    }
+
+}
+
--- a/pipeline/vars/common.groovy
+++ b/pipeline/vars/common.groovy
@@ -1,0 +1,132 @@
+#!/usr/bin/env groovy
+/*
+    Common groovy methods that can be reused by the pipeline jobs.
+*/
+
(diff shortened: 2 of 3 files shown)


pipeline/5/Jenkinsfile-tier-0.groovy
```diff
@@ -0,0 +1,68 @@
...
+                        println "Work in progress"
+                    }
+
+                 }]
+
+// Pipeline script entry point
+
+node(nodeName) {
+
+    stage('Checkout') {
```
sunilkumarn417, line 34: Can we combine checkout and Prepare node as a single pre-requisite task.

pipeline/vars/common.groovy
```diff
@@ -0,0 +1,127 @@
...
+            - sutVMConf
+            - sutConf
+            - testSuite
+            - addnArgs
+    */
+    println "Begin Test Suite execution"
+
+    // generate random instance name
+    def sec = sh(script: "echo \$(date +%s)", returnStdout: true).trim()
+    instanceName = "psi-${sec}"
```
sunilkumarn417, line 60: can we attach tier level string `tier0` into name string instead of `psi`

pipeline/vars/common.groovy
```diff
@@ -0,0 +1,127 @@
...
+    cmd += " --instances-name ${instanceName}"
+
+    // Append test suite specific
+    cmd += " --inventory ${env.sutVMConf}"
+    cmd += " --global-conf ${env.sutConf}"
+    cmd += " --suite ${env.testSuite}"
+
+    // If CI_MESSAGE is populated, the preference is use the information.
+    def ciMessage = "${params.CI_MESSAGE}" ?: ""
+    if (ciMessage?.trim()) {
```
sunilkumarn417, line 76: what is the probability of not getting CI message details, can we raise build error in case not getting CI_Message.

pipeline/vars/common.groovy
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    // Check for additional arguments
+    def addnArgFlag = "${env.addnArgs}" ?: ""
+    if (addnArgFlag?.trim()) {
+        cmd += " ${env.addnArgs}"
+    }
+
+    //command execution
+    rc = sh(script: "PYTHONUNBUFFERED=1 ${cmd}", returnStatus: true)
```
sunilkumarn417, line 112: Do we need to check exit status? Since on test failure jenkins stage will be failed and exited.

pipeline/vars/common.groovy
```diff
@@ -0,0 +1,127 @@
...
+    if (rc != 0) {
+        println "Execution encountered a failure"
+    }
+
+    cleanup_cmd = "PYTHONUNBUFFERED=1 ${env.WORKSPACE}/.venv/bin/python"
```
sunilkumarn417, line 117: IMO, do we need try-catch-finally clause for tear down tasks.

pipeline/5/Jenkinsfile-tier-0.groovy
psathyan: It's possible... I understand joining them would combine the setup process.

pipeline/vars/common.groovy
rakeshgm: this function looks like a common function, which can be used even for other tiers. so I think string `tier0` might not be correct here.

pipeline/vars/common.groovy
psathyan: The method being common and to support re-usability... it's possible that the `ciMessage` is not being set.

pipeline/vars/common.groovy
```diff
@@ -0,0 +1,127 @@
...
+        wget http://magna002.ceph.redhat.com/cephci-jenkins/.osp-cred-ci-2.yaml -O ${HOME}/osp-cred-ci-2.yaml
+        wget http://magna002.ceph.redhat.com/cephci-jenkins/.cephci.yaml -O ${HOME}/.cephci.yaml
+
+        # Install cephci pre-requisistes
+        rm -rf .venv
+        python3 -m venv .venv
+        source .venv/bin/activate
+        python -m pip install --upgrade pip
+        python -m pip install -r requirements.txt
+        deactivate
```
rakeshgm, line 42: is it possible to put this as `bootstrap.sh` and call the file here instead of writing shell script code here

psathyan: As the code was specific to `centos-7` template, I didn't want to create a separate file. However, I could keep it out in a separate file but locally to the `vars` folder.

pipeline/vars/common.groovy
sunilkumarn417: makes sense. However in this case I think we may need to add docker image run-time options to go with right ceph-compose and ceph images, if CI_Message is not set.

psathyan: Good question, that is why I have inserted `addnArgs` environment variable. :)

pipeline/vars/common.groovy
psathyan: Yes, that is correct. I have made the required changes.

pipeline/vars/common.groovy
psathyan: Thanks, taken care

sunilkumarn417 (approved):
Looks good to me.

rakeshgm (approved):

PR #410, opened 2021-03-11
run.py: CLI argument to skip enabling rpms from subscription

pdhiran:
Adding cli argument to skip the addition of rpms provided by the subscription for interop runs.
When specifying the skip argument, the repos specified via a repo file with --add-repo option will be added.

### DIFF
--- a/run.py
+++ b/run.py
@@ -71,6 +71,7 @@ A simple test suite wrapper that executes tests based on yaml test configuration
         [--custom-config-file <file>]
         [--xunit-results]
         [--enable-eus]
+        [--skip-enabling-rhel-rpms]
   run.py --cleanup=name [--osp-cred <file>]
         [--log-level <LEVEL>]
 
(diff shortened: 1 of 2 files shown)


run.py
```diff
@@ -71,6 +71,7 @@
         [--custom-config-file <file>]
         [--xunit-results]
         [--enable-eus]
+        [--skip-subscription-repos]
```
rakeshgm, line 74: `--skip-subscription-repos` actually sounds incorrect. we are not skipping subscription here, we just not enabling rhel rpms. can you use a different verbiage here like `skip-enabling-rhel-rpms` or something similar.

pdhiran: Agree. Made the suggested changes.

pdhiran:
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1615466009169/
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1615466009169/result.html

tests/misc_env/install_prereq.py
```diff
@@ -80,10 +88,13 @@ def install_prereq(
...
-            if enable_eus:
-                enable_rhel_eus_rpms(ceph, distro_ver)
+            if not skip_enabling_rhel_rpms:
+                if enable_eus:
+                    enable_rhel_eus_rpms(ceph, distro_ver)
+                else:
+                    enable_rhel_rpms(ceph, distro_ver)
             else:
-                enable_rhel_rpms(ceph, distro_ver)
+                log.info("Skipped enabling the RPM's provided by the Subscription")
```
rakeshgm, line 97: this is a wrong message. subscription provides lot of rpms are not disabling all of them, we are skipping only rhel-rpms that's all, can you change the message here.

pdhiran: Changed the logging with appropriate message

rakeshgm (approved):

PR #412, opened 2021-03-11
cephadm: Added ceph health verification

sunilkumarn417:

- Included ceph health verification for Tier0 test-suite.

**Results:**
http://magna002.ceph.redhat.com/ceph-qe-logs/Sunil/cephci-run-1615471449607/result.html

### DIFF
--- a/ceph/ceph.py
+++ b/ceph/ceph.py
@@ -607,9 +607,22 @@ class Ceph(object):
                 return metadata
         return None
 
-    def osd_check(self, client):
+    def osd_check(self, client, rhbuild=None):
+        """
+        Check OSD status
+        Args:
+            client: client node to get OSD details
+            rhbuild: ceph build version
 
-        out, err = client.exec_command(cmd="sudo ceph -s -f json")
+        Returns:
+            0 - Successful
+            1 - failure
+        """
+        cmd = "ceph -s -f json"
+        if rhbuild and rhbuild.startswith("5"):
+            cmd = f"cephadm shell -- {cmd}"
+
+        out, err = client.exec_command(cmd=cmd, sudo=True)
         out_json = out.read().decode()
         ceph_status_json = json.loads(out_json)
 
@@ -647,7 +660,7 @@ class Ceph(object):
         Returns:
            int: return 0 when ceph is in healthy state, else 1
         """
-
+        pacific = True if (rhbuild and rhbuild.startswith("5")) else False
         if not client:
             client = (
                 self.get_ceph_object("client")
@@ -662,7 +675,11 @@ class Ceph(object):
         valid_states = ["active+clean"]
 
         while datetime.datetime.now() - starttime <= timeout:
-            out, err = client.exec_command(cmd="sudo ceph -s")
+            cmd = "ceph -s"
+            if pacific:
+                cmd = f"cephadm shell -- {cmd}"
+
+            out, err = client.exec_command(cmd=cmd, sudo=True)
             lines = out.read().decode()
 
             if not any(state in lines for state in pending_states):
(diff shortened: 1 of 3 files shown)


ceph/ceph.py
```diff
@@ -607,9 +607,12 @@ def get_osd_metadata(self, osd_id, client=None):
                 return metadata
         return None
 
-    def osd_check(self, client):
+    def osd_check(self, rhbuild, client):
+        cmd = "ceph -s -f json"
```
psathyan, line 611: IMO, we shouldnt attempt to change the method signature. That i.e., we could do `def osd_check(self, client, rhbuild=None):`

ceph/ceph.py
```diff
@@ -607,9 +607,12 @@ def get_osd_metadata(self, osd_id, client=None):
...
+        if rhbuild.startswith("5"):
```
psathyan, line 612: Continuing from comment 1, we could do `if rhbuild and rhbuild.startswith("5"):`

ceph/ceph.py
```diff
@@ -607,9 +607,12 @@ def get_osd_metadata(self, osd_id, client=None):
                 return metadata
         return None
 
-    def osd_check(self, client):
+    def osd_check(self, rhbuild, client):
```
psathyan, line 610: Kindly add docstrings... I am aware that it wasn't available earlier but it would be helpful if we could add it.

ceph/ceph.py
```diff
@@ -674,18 +681,19 @@ def check_health(self, rhbuild, client=None, timeout=300):
             logger.error("Valid States are not found in the health check")
             return 1
 
-        self.osd_check(client)
+        self.osd_check(rhbuild, client)
```
psathyan, line 684: Continuing from comment 1, we could do `self.osd_check(client, rhbuild)`

tests/ceph_installer/test_cephadm.py
```diff
@@ -104,7 +104,7 @@ def run(ceph_cluster: Ceph, **kwargs) -> int:
     try:
         config = kwargs["config"]
         cephadm = CephAdmin(cluster=ceph_cluster, **config)
-        steps = config.get("steps")
+        steps = config.get("steps", [])
```
psathyan, line 107: Are there any scenarios where `steps` wouldn't be used?

If there is none, can we make it mandatory i.e. `steps = config[steps]`

tests/ceph_installer/test_cephadm.py
```diff
@@ -117,7 +117,12 @@ def run(ceph_cluster: Ceph, **kwargs) -> int:
             func = fetch_method(obj, cfg["command"])
             func(cfg)
 
-        return 0
+        if config.get("verify_health"):
+            cephadm.cluster.check_health(
```
psathyan, line 121: Nice... good idea on re-usability.

ceph/ceph.py
sunilkumarn417: I did noticed osd_check was used only in check_health, but however this makes sense.

ceph/ceph.py
sunilkumarn417: sure

tests/ceph_installer/test_cephadm.py
sunilkumarn417: Currently from this PR, `verify_health` is one which doesn't need `steps`.

ceph/ceph.py
sunilkumarn417: Added changes.

ceph/ceph.py
sunilkumarn417: done

ceph/ceph.py
sunilkumarn417: done.

psathyan (approved):

PR #432, opened 2021-03-23
Avoid failure rhs-ceph-repo if repo doesn't end with a /

VasishtaShastry:
With this fix, run no longer fails even if user doesn't end "--rhs-ceph-repo"'s value with a '/'
Prior -
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1616388243467/
base url was incorrectly formed with below string "RHCEPH-4.2-RHEL-8-20210316.ci.0compose" 

Now -
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1616431770187

# Description

# Checklist:

### DIFF
--- a/ceph/ceph.py
+++ b/ceph/ceph.py
@@ -882,7 +882,8 @@ class Ceph(object):
         """
         repo_file = ""
         for repo in repos:
-            repo_to_use = base_url + "compose/" + repo + "/x86_64/os/"
+            base_url = base_url.rstrip("/")
+            repo_to_use = f"{base_url}/compose/{repo}/x86_64/os"
             r = requests.get(repo_to_use, timeout=10)
             logger.info("Checking %s", repo_to_use)
             if r.status_code == 200:


VasishtaShastry:
@manasagowri @rakeshgm

ceph/ceph.py
```diff
@@ -881,7 +882,7 @@ def generate_repository_file(base_url, repos):
         """
         repo_file = ""
         for repo in repos:
-            repo_to_use = base_url + "compose/" + repo + "/x86_64/os/"
+            repo_to_use = os.path.join(base_url, "compose", repo, "x86_64/os/")
```
psathyan, line 885: It is kind of misleading though it achieves the requirement, IMO
```
repo = repo.rstrip('/')
repo_to_use = f"{base_url}compose/{repo}/x86_64/os"
```

rakeshgm: os.path.join resolves `/` or `//`

VasishtaShastry: @psathyan is it okay if we avoid repo = repo.rstrip('/') and allow it to add two '/' as it wouldn't make any issues ?

ktdreyer: I think `os.path` is a good solution. Usually for URLs I use `posixpath.join()` though, because that will use forward slashes everywhere. Not like we're going to run on MS Windows though :)

ktdreyer: (also, reading this code, we should probably rename `repo_to_use` to `baseurl` in a separate PR)

psathyan: > @psathyan is it okay if we avoid repo = repo.rstrip('/') and allow it to add two '/' as it wouldn't make any issues ?

@VasishtaShastry let's keep it clean. We should strip or add the `/`... 

The double slash is a problem unless the gateway is capable of redirecting the request.

psathyan:
@VasishtaShastry would you be able to implement the changes and merge/close the PR? Let me know if you need any help

VasishtaShastry:
@psathyan 
> @VasishtaShastry would you be able to implement the changes and merge/close the PR? Let me know if you need any help

Yes, I can do it, Will modify codes soon
Sure, Thank you

sunilkumarn417 (approved):
LGTM

PR #439, opened 2021-03-29
RGW: Tier 0 test suite for RHCS 5.0

psathyan:
# Description

Adding the identified tier 0 test suite for RGW which would be executed against every new build of RHCS 5.0. In this PR, the following changes are made to support the addition

- Ceph Cluster object's attribute `rhcs_version` is initialized before the test suite execution
- RGW tier 0 test suite for RHCS 5.0
- Modifications to `sanity_rgw.py` test module to support test execution in the new environment 

## Logs
**RHCS 5** http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/tier0/5/
**RHCS 4** http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/tier0/4/

### DIFF
--- a/tests/rgw/sanity_rgw.py
+++ b/tests/rgw/sanity_rgw.py
@@ -1,6 +1,6 @@
 import logging
 
-# from ceph.ceph import Ceph
+from utility.utils import setup_cluster_access
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


tests/rgw/sanity_rgw.py
```diff
@@ -24,40 +24,56 @@ def run(ceph_cluster, **kw):
...
-        rgw_node.exec_command(cmd="sudo yum install ceph-radosgw -y")
-    rgw_node.exec_command(
-        cmd="sudo pip3 install -r {test_folder}/ceph-qe-scripts/rgw/requirements.txt".format(
-            test_folder=test_folder
+    test_folder_path = f"~/{test_folder}"
+    git_url = "https://github.com/red-hat-storage/ceph-qe-scripts.git"
+    git_clone = f"git clone {git_url} -b master"
+
+    # Clone the repository once for the entire test suite
+    pip_cmd = "venv/bin/pip"
```
rakeshgm, line 43: `venv` here is the virtual env name. and if we are separating out the cmd here I think it would be best to define another variable for virtualenv and use the same where ever possible. i,.e while creating venv like in line  51

tests/rgw/sanity_rgw.py
```diff
@@ -24,40 +24,56 @@ def run(ceph_cluster, **kw):
...
+    python_cmd = "venv/bin/python"
```
rakeshgm, line 44: we are need to call python3 explicitly because `python` in rhel 7.x runs python2 and from rhel 8.x we need to call python3 explicitly. and the same goes for using `pip` too. we need to use `pip3`. There are alternatives we can set using `alternatives --set python /usr/bin/python3` but its not recommended on rhel. 

check here [python in RHEL-8](https://developers.redhat.com/blog/2018/11/14/python-in-rhel-8/)

rakeshgm: since you are creating venv using python3, we can ignore this comment.

rakeshgm:
looks good.

rakeshgm (approved):

PR #450, opened 2021-04-06
utils.py: fix get_node_by_id() docstring

ktdreyer:
Update the docstring to describe this function's behavior.

This fixes a bad copy-and-paste from `get_nodes_by_id()`'s docstring.

### DIFF
--- a/ceph/utils.py
+++ b/ceph/utils.py
@@ -735,12 +735,13 @@ def get_nodes_by_id(cluster, node_names):
 
 def get_node_by_id(cluster, node_name):
     """
-    Get node object(s) by Node Id
-    Returns all nodes object list if node_names is empty,
-    else node list which matched node ID(eg., 'node1')
+    Search cluster for the first node object that matches node_name.
+
+    If this cluster has no nodes with this shortname, return None.
+
     Args:
         cluster: ceph object
-        node_name: node name list (eg., 'node1')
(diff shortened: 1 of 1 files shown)


ceph/utils.py
```diff
@@ -735,12 +735,13 @@ def get_nodes_by_id(cluster, node_names):
 
 def get_node_by_id(cluster, node_name):
     """
-    Get node object(s) by Node Id
-    Returns all nodes object list if node_names is empty,
-    else node list which matched node ID(eg., 'node1')
+    Search cluster for the first node object that matches node_name.
```
ktdreyer, line 738: Note, my description of "the first node object that matches node_name" is ambiguous here. The code actually checks `if node_name in node.shortname`, and if those are both strings, it's doing a sub-string match, not an equality check. Is that intentional?

psathyan (approved):

ceph/utils.py
sunilkumarn417: Thanks @ktdreyer for finding this, I will raise a separate PR for this.

sunilkumarn417 (approved):

ceph/utils.py
ktdreyer: Thanks. I added unit tests for this method in https://github.com/red-hat-storage/cephci/pull/453 . Please delete the `@pytest.mark.xfail` marker when you fix this.

PR #453, opened 2021-04-07
add unit tests for ceph.utils.get_node_by_id()

ktdreyer:
We intend to fix `test_substring_does_not_match()` eventually.

### DIFF
--- a/unittests/ceph/test_utils.py
+++ b/unittests/ceph/test_utils.py
@@ -1,0 +1,44 @@
+import pytest
+
+from ceph.utils import get_node_by_id
+
+
+class FakeCephNode(object):
+    # Fake ceph.ceph.CephNode object
+    def __init__(self, shortname):
+        self.shortname = shortname
+
+
+class FakeCeph(object):
+    # Fake ceph.ceph.Ceph object
+    def __init__(self):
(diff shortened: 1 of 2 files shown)


unittests/ceph/test_utils.py
```diff
@@ -0,0 +1,44 @@
...
+        assert get_node_by_id(cluster, "osd1") == osd1
+        assert get_node_by_id(cluster, "osd2") == osd2
+
+    def test_no_match(self, cluster):
+        osd1 = FakeCephNode("osd1")
+        osd2 = FakeCephNode("osd2")
+        cluster.nodes = [osd1, osd2]
+        assert get_node_by_id(cluster, "bogushost") is None
+
+    @pytest.mark.xfail
```
psathyan, line 39: @ktdreyer as #455 is merged, would you be removing the fail test mark?

psathyan:
@ktdreyer it is a good PR which I would like to merge. As the test case was marked expected to fail, I was hoping that we can remove the mark as the fix for the failure has been merged.

ktdreyer:
Whoops, I think I deleted this branch prematurely. I agree, I think we should merge it eventually.

PR #455, opened 2021-04-08
Fix get_node_by_id to get exact node.

sunilkumarn417:

Since getting exact node by using node sub-string is quite tricky, so used instance naming convention which was used at the VM instances creation.

```
(Pdb) for i in get_nodes_by_id(kw['ceph_cluster'], []): i.shortname
'ceph-sunil1adm-1617805761413-node3-mon-osd-node-exporter-crash'
'ceph-sunil1adm-1617805761413-node4-client'
'ceph-sunil1adm-1617805761413-node2-osd-mon-mgr-mds-node-exporte'
'ceph-sunil1adm-1617805761413-node1-installer-mon-mgr-osd-node-e'

(Pdb) for i in get_nodes_by_id(kw['ceph_cluster'], ["node11"]): i.shortname

(Pdb) for i in get_nodes_by_id(kw['ceph_cluster'], ["node1", "node11", "node3"]): i.shortname
'ceph-sunil1adm-1617805761413-node1-installer-mon-mgr-osd-node-e'
'ceph-sunil1adm-1617805761413-node3-mon-osd-node-exporter-crash'

(Pdb) get_node_by_id(kw['ceph_cluster'], "node1").shortname
'ceph-sunil1adm-1617805761413-node1-installer-mon-mgr-osd-node-e'

(Pdb) get_node_by_id(kw['ceph_cluster'], "node11").shortname
*** AttributeError: 'NoneType' object has no attribute 'shortname'
```

### DIFF
--- a/ceph/utils.py
+++ b/ceph/utils.py
@@ -705,7 +705,24 @@ def get_disk_info(node):
     return disks
 
 
-def get_nodes_by_id(cluster, node_names):
+def get_node_by_id(cluster, node_name):
+    """
+    Search cluster for the first node object that matches node_name.
+
+    If this cluster has no nodes with this shortname, return None.
+
+    Args:
+        cluster: ceph object
+        node_name: node shortname, eg., 'node1'
+    Returns:
+        node
+    """
+    for node in cluster.get_nodes():
+        if node_name in node.shortname:
+            return node
+
+
+def get_nodes_by_ids(cluster, node_names):
     """
     Get node object(s) by Node Id
 
@@ -731,20 +748,3 @@ def get_nodes_by_id(cluster, node_names):
             for name in nodes
             if name in node.shortname
         ]
-
-
-def get_node_by_id(cluster, node_name):
-    """
-    Search cluster for the first node object that matches node_name.
-
-    If this cluster has no nodes with this shortname, return None.
-
-    Args:
-        cluster: ceph object
-        node_name: node shortname, eg., 'node1'
-    Returns:
-        node
-    """
-    for node in cluster.get_nodes():
-        if node_name in node.shortname:
-            return node
--- a/ceph/utils.py
+++ b/ceph/utils.py
@@ -707,15 +707,32 @@ def get_disk_info(node):
 
 def get_node_by_id(cluster, node_name):
     """
-    Search cluster for the first node object that matches node_name.
+    Fetch node using provided node substring
 
-    If this cluster has no nodes with this shortname, return None.
+    As per the naming convention used at VM creation, where each node.shortname
+    is framed with hyphen("-") separated string as below, please refer
+    ceph.utils.create_ceph_nodes definition
+        "ceph-<name>-node1-<roles>"
+
+        name: RHOS-D username or provided --instances-name
+        roles: roles attached to node in inventory file
+
+    In this method we use hyphen("-") appended to node_name string to try fetch exact node.
+    for example,
+        "node1" ----> "node1-"
+
+    Note: But however if node_name doesn't follow naming convention as mentioned in
+    inventory, the first searched node will be returned.
+    for example,
+        "node" ----> it might return any node which matched first, like node11.
+
+    return None, If this cluster has no nodes with this substring.
 
     Args:
         cluster: ceph object
-        node_name: node shortname, eg., 'node1'
+        node_name: node1        # try to use node<Id>
     Returns:
-        node
+        node instance (CephVMNode)
     """
     # Append "-" as per naming convention used at instance creation
     if not node_name.endswith("-"):
@@ -728,19 +745,18 @@ def get_node_by_id(cluster, node_name):
 
 def get_nodes_by_ids(cluster, node_names):
     """
-    Get node object(s) by Node Id
+    Fetch nodes using provided substring of nodes
 
-    Returns all nodes object list if node_names is empty,
-    else node list which matched node ID(eg., 'node1')
+    Returns node list which matched node ID(eg., 'node1')
+    else all nodes object list if node_names is empty.
 
     Args:
         cluster: ceph object
         node_names: node name list (eg., ['node1'])
 
     Returns:
-        node_list: list nodes
+        node_list: list of nodes
     """
-    # Empty list wll pick all cluster nodes
     if not node_names:
         return cluster.get_nodes()
 
--- a/ceph/utils.py
+++ b/ceph/utils.py
@@ -717,6 +717,10 @@ def get_node_by_id(cluster, node_name):
     Returns:
         node
     """
+    # Append "-" as per naming convention used at instance creation
+    if not node_name.endswith("-"):
+        node_name += "-"
+
     for node in cluster.get_nodes():
         if node_name in node.shortname:
             return node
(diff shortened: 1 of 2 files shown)


ktdreyer:
Thanks for adding more docstrings!

In `get_node_by_id()`'s docstring, what does `<name>` refer to? Is it a `username`, `--instance-name`, or what name are we referring to there?

```
"ceph-<name>-node1-<roles>"
```

------
Let's take out these lines in `get_node_by_id()`:
```python
if not isinstance(node_name, str):
    return None
```

In Python we usually rely on Duck Typing, so as long as we can append that `-` character to the `node_name` value, and the other comparison code works on this value, it should not matter if the value is a `str`. Otherwise it's just more codepaths to document, test, and maintain.

------

I'm seeing a similar typing issue with `node_names` in `get_nodes_by_id()`.

```python
node_names = node_names if isinstance(node_names, list) else [node_names]
```

Reading the code in `ceph/ceph_admin/apply.py` that calls `get_nodes_by_id(self.cluster, nodes)`, it appears `nodes` is always a list there. The docstring for `ApplyMixin.apply()` says that `nodes` is `a list of strings` and gives a YAML example of a list of strings.

Let's remove this string-to-list conversion in `get_nodes_by_id()` and always assume the caller wants us to iterate over `node_names`. This makes this method a lot easier to reason about and write unit tests.

------

In `get_nodes_by_id()`, it's better to write the "Empty list will pick all cluster nodes" comment in the docstring than inline in the code, because that behavior is surprising to me as a noob to this codebase. I'm reading that this method "Fetch nodes using provided substring of node name" so I would have expected an empty list input to return an empty list output.

------

I'm having a hard time understanding all the changes in this PR, particularly without unit tests (see #453). **You don't have to do this on this PR**, but for what it's worth, in the future it would be a good idea to break this PR up into three commits:

* One commit should move the `get_node_by_id()` definition above the `get_nodes_by_id()` definition. No functional change, just reorder the method definitions and keep everything else the same.
* One commit should make the miscellaneous docstring tweaks for `get_node_by_id()` without changing behavior. (For example, you've changed `Returns: node` changes to `Returns: node object` and it's hard for me to mentally distinguish all the other changes.)
* The final commit should make all the functional changes.

ceph/utils.py
```diff
@@ -705,46 +705,66 @@ def get_disk_info(node):
...
+    # Empty list wll pick all cluster nodes
+    if not node_names:
+        return cluster.get_nodes()
+    else:
+        nodes_ = []
+        for name in node_names:
+            node_ = get_node_by_id(cluster, name)
+            if node_:
+                nodes_.append(node_)
+        return nodes_
```
psathyan, line 770: We could avoid the `else` statement as there is no further checks/code statements

```
if node_names is None:
  return cluster.get_nodes()

nodes = list()
for name in node_names:
  node = get_node_by_id(cluster, name)
  if node is not None: nodes.append(node)

return nodes
```

ceph/utils.py
```diff
@@ -705,46 +705,66 @@ def get_disk_info(node):
...
+    # Appending "-" as per naming convention used at instance creation
+    node_name += "-"
 
+    for node in cluster.get_nodes():
+        if node_name in node.shortname:
+            return node
 
-def get_node_by_id(cluster, node_name):
+
+def get_nodes_by_id(cluster, node_names):
```
psathyan, line 745: IMO, this method name could be `get_nodes_by_ids`. As the purpose is to return a list of nodes matching the list of given ids. 

If the intent is that then we can avoid the check of `node_names` being an empty list

ceph/utils.py
```diff
@@ -705,46 +705,66 @@ def get_disk_info(node):
...
-    Get node object(s) by Node Id
+    Fetch node using provided substring of node name
 
-    Returns all nodes object list if node_names is empty,
-    else node list which matched node ID(eg., 'node1')
+    As per the naming convention used at VM creation, where each node.shortname
+    is framed with hyphen("-") separated string as below,
+        "ceph-<name>-node1-<roles>"
+
+    In this method we use hyphen("-") appended to node_name string to fetch exact node.
```
psathyan, line 716: I believe it is sufficient to prefix `-`... but do you think we should also suffix the value?

sunilkumarn417:
> Thanks for adding more docstrings!
> 
> In `get_node_by_id()`'s docstring, what does `<name>` refer to? Is it a `username`, `--instance-name`, or what name are we referring to there?
> 
> ```
> "ceph-<name>-node1-<roles>"
> ```
> 
> Let's take out these lines in `get_node_by_id()`:
> 
> ```python
> if not isinstance(node_name, str):
>     return None
> ```
> 
> In Python we usually rely on Duck Typing, so as long as we can append that `-` character to the `node_name` value, and the other comparison code works on this value, it should not matter if the value is a `str`. Otherwise it's just more codepaths to document, test, and maintain.
> 
> I'm seeing a similar typing issue with `node_names` in `get_nodes_by_id()`.
> 
> ```python
> node_names = node_names if isinstance(node_names, list) else [node_names]
> ```
> 
> Reading the code in `ceph/ceph_admin/apply.py` that calls `get_nodes_by_id(self.cluster, nodes)`, it appears `nodes` is always a list there. The docstring for `ApplyMixin.apply()` says that `nodes` is `a list of strings` and gives a YAML example of a list of strings.
> 
> Let's remove this string-to-list conversion in `get_nodes_by_id()` and always assume the caller wants us to iterate over `node_names`. This makes this method a lot easier to reason about and write unit tests.
> 
> In `get_nodes_by_id()`, it's better to write the "Empty list will pick all cluster nodes" comment in the docstring than inline in the code, because that behavior is surprising to me as a noob to this codebase. I'm reading that this method "Fetch nodes using provided substring of node name" so I would have expected an empty list input to return an empty list output.
> 
> I'm having a hard time understanding all the changes in this PR, particularly without unit tests (see #453). **You don't have to do this on this PR**, but for what it's worth, in the future it would be a good idea to break this PR up into three commits:
> 
> * One commit should move the `get_node_by_id()` definition above the `get_nodes_by_id()` definition. No functional change, just reorder the method definitions and keep everything else the same.
> * One commit should make the miscellaneous docstring tweaks for `get_node_by_id()` without changing behavior. (For example, you've changed `Returns: node` changes to `Returns: node object` and it's hard for me to mentally distinguish all the other changes.)
> * The final commit should make all the functional changes.

Hi @ktdreyer ,

Thanks for all pointers and suggestions.
I have updated the commits, which address all the review comments.

ceph/utils.py
sunilkumarn417: Done.

ceph/utils.py
sunilkumarn417: Done. updated.

ceph/utils.py
sunilkumarn417: IMO, for now prefix would be fine. 
As we discussed we will try to eliminate the problem by a node attribute in future.

psathyan (approved):

ceph/utils.py
ktdreyer: Thanks for clarifying that this will be a node attribute in the future. As I was reading this code, I was thinking we could really benefit from storing this as an attribute on the node.

ktdreyer (approved):
Looks good. Sure, I will update https://github.com/red-hat-storage/cephci/pull/453 on top of this.

PR #472, opened 2021-04-26
Fix 468: ERROR - 'str' object has no attribute 'read'

### LINKED ISSUE #468: Ceph health check fails with error - ERROR - 'str' object has no attribute 'read'
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1619152243856/check-ceph-health_1.log

While running upgrade_containerized on a rhel-8 cluster for 4.2z1 RC, the ceph health check failed.

2021-04-23 01:47:52,747 - exec - ERROR - 'str' object has no attribute 'read'

psathyan:
# Description

When `long_running` argument is enabled then the return type is string which has caused the breaking mentioned in #468 

This PR modifies the default behavior to check the type before decoding.

Closes #468 

### Logs:
4: http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/i468/4/ 
5: http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/i468/5/

### DIFF
--- a/tests/misc_env/exec.py
+++ b/tests/misc_env/exec.py
@@ -4,6 +4,8 @@ import logging
 import re
 from typing import Any, Tuple
 
+from paramiko.channel import ChannelFile
+
 from ceph.utils import get_node_by_id
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


tests/misc_env/exec.py
```diff
@@ -27,7 +29,10 @@ def exec_command(node, **kwargs: Any) -> Tuple:
         CommandFailed   when there is an execution failure
     """
     out, err = node.exec_command(**kwargs)
-    return out.read().decode(), err.read().decode()
+
+    out = out.read().decode() if isinstance(out, ChannelFile) else out
```
sunilkumarn417, line 33: Can we avoid checking `ChannelFile`, instead going with `str` would be good.

sunilkumarn417: this way, we could avoid import statement also.

psathyan: In exec_command, we are doing `out.read()` which could return a `str` or `bytes` but mostly `str` in our case. As the need to perform `read` is for a base file type object, I went with this implementation... in future, if we change the return value, I didn't want to break it...

sunilkumarn417 (approved):

PR #483, opened 2021-04-28
Defaulting package installation via yum instead of DNF

### LINKED ISSUE #477: Bail out dnf commands on rhel version
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1619538918459/rgw_sanity_tests_0.log

`dnf` commands should get bail out based on rhel versions.
https://github.com/red-hat-storage/cephci/blob/master/tests/rgw/sanity_rgw.py#L86

psathyan:
# Description

`dnf` utility is not available in OS version older than RHEL 8 hence switching to using `yum` across all the supported OS versions. This works for the short term.

Closes #477


sunilkumarn417 (approved):
LGTM

PR #487, opened 2021-04-29
RHCEPHQE-36: Task 1 publish the test run information to report portal

psathyan:
# Description

This PR is a completer for the user story RHCEPHQE-36 wherein the results of the run would need to be captured in the Report Portal. Here we add the CLI option `--report-portal` to be passed to the executor.

In addition to this change, we are moving to use grafana's beta image.


sunilkumarn417 (approved):
Looks god to me.

Do we need `--grafana-image` in all suites to avoid failure in any condition.

psathyan:
> Do we need `--grafana-image` in all suites to avoid failure in any condition.

Yes. I thought we decided to use this option only for deployment to confirm that the build is good once they have switched to a non-existent build.

Also, some of the stages use `skip-monitoring-stack`

PR #492, opened 2021-04-30
Moving custom image of tier 0 cephadm to beta

psathyan:
# Description

With the beta images available, moving the custom image option for deploy in tier 0 to use it. Along with this change, the wait time between the stages have been reduced as cephci internally throttles the VM.


sunilkumarn417 (approved):
LGTM

PR #506, opened 2021-05-11
fixes fs to bs script issue

Ameenasuhani:
Description:
Filestore to bluestore migration fails  when osd or other daemons are evaluated before the installer with error local variable 'installer_node' referenced before assignment

Failure Logs:
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1620725514928/filestore_to_bluestore_migration_0.log
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1620650713914/filestore_to_bluestore_migration_0.log

### DIFF
--- a/ceph/utils.py
+++ b/ceph/utils.py
@@ -70,7 +70,7 @@ def create_ceph_nodes(
 
         with parallel() as p:
             for node in range(1, 100):
-                sleep(5 * node)
+                sleep(10 * node)
                 node = "node" + str(node)
                 if not ceph_cluster.get(node):
                     break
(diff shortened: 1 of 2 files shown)


tests/ceph_ansible/filestore_to_bluestore.py
```diff
@@ -16,6 +16,7 @@ def run(**kw):
     for cnode in ceph_nodes:
         if cnode.role == "installer":
             installer_node = cnode
```
psathyan, line 18: We could do this way... 
```
installer_node = kwargs["ceph_cluster"].get_nodes(role="installer")[0]

for cnode in ceph_nodes:
  ...
```

sunilkumarn417: @Ameenasuhani , I do suggest same fix as @psathyan pointed out. Since there is possibility of `installer_node` not being set and its good to fail there itself.

PR #508, opened 2021-05-12
Apply spec: Added support for NFS and MDS service

sunilkumarn417:
- NFS and MDS service deployment using spec file.

**Test results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1620808639565

### DIFF
--- a/ceph/ceph_admin/helper.py
+++ b/ceph/ceph_admin/helper.py
@@ -86,6 +86,18 @@ class GenerateServiceSpec:
     def get_labels(node):
         return node.role.role_list
 
+    def get_hostnames(self, node_names):
+        """
+        Return list of hostnames
+        Args:
+            node_names: node names
+
+        Returns:
+            list of hostanmes
+        """
+        nodes = get_nodes_by_ids(self.cluster, node_names)
+        return [node.shortname for node in nodes]
+
     def _get_template(self, service_type):
         """
         Return Jinja template based on the service_type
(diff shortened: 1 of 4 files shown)


ceph/ceph_admin/helper.py
```diff
@@ -200,6 +200,73 @@ def generate_osd_spec(self, spec):
...
+        Returns:
+            service_spec
+        """
+        template = self._get_template("mds")
+        node_names = spec["placement"].pop("nodes", None)
+        if node_names:
+            spec["placement"]["hosts"] = []
+            nodes = get_nodes_by_ids(self.cluster, node_names)
+            for node in nodes:
+                spec["placement"]["hosts"].append(node.shortname)
```
psathyan, line 231: Would it make sense to have a separate method to return the host list which then can we set to `spec["placement"]["hosts"]`? 

I noticed that we are using these lines across multiple methods.

psathyan (commented):
Looks good, a minor improvement.

sunilkumarn417:
**latest results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1621224191694/result.html

ceph/ceph_admin/helper.py
sunilkumarn417: Done.

psathyan (approved):

rakeshgm:
I believe this should be nfs ganesha or nfs ganesha rgw and not just nfs we will also be having nfs for Cephfs too. @sunilkumarn417 can you refactor the verbiage

sunilkumarn417:
> I believe this should be nfs ganesha or nfs ganesha rgw and not just nfs we will also be having nfs for Cephfs too. @sunilkumarn417 can you refactor the verbiage

Since the way NFS is deployed is common, did not wanted to name it with protocol. However the test case is defined with protocol.

PR #528, opened 2021-05-26
RHCEPHQE-63: Adding support for Tier 1 RBD test suite

psathyan:

# Description

This PR adds the tier 1 test suite for RBD that executes the list of test cases identified.


suites/pacific/rbd/tier_1_rbd.yaml
```diff
@@ -0,0 +1,168 @@
+# Tier1: Extended RBD acceptance testing
+#
+# This test suite runs addition test scripts to evaluate the existing functionality of
+# Ceph RBD component.
+#
+# The following testing is carried out
+#   - verification of NBD unmap/map
```
rakeshgm, line 7: not to nitpick, but there is a type here. i think it should be `RBD` instead of `NBD`  :)

rakeshgm: My bad, @psathyan clarified. it's `NBD - network block device. `

sunilkumarn417 (approved):

PR #586, opened 2021-06-25
Execute radosbench sequence during cephadm upgrade test

sunilkumarn417:

- Added radosbench test sequence during cephadm upgrade.
- Include wait function to check completion of all rados bench executors.

**Results**:
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1625646410909 (Negative testing - Upgrade is failed)
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1625645635443/Upgrade_ceph_0.log (Upgrade Successful)

### DIFF
--- a/tests/cephadm/test_cephadm_upgrade.py
+++ b/tests/cephadm/test_cephadm_upgrade.py
@@ -5,8 +5,13 @@ Test module that verifies the Upgrade of Ceph Storage via the cephadm CLI.
 import logging
 
 from ceph.ceph_admin.orch import Orch
+from ceph.rados.rados_bench import RadosBench
 
-LOG = logging.getLogger()
+LOG = logging.getLogger(__name__)
+
+
+class UpgradeFailure(Exception):
+    pass
 
 
 def run(ceph_cluster, **kwargs) -> int:
@@ -35,7 +40,15 @@ def run(ceph_cluster, **kwargs) -> int:
     LOG.info("Upgrade Ceph cluster...")
     config = kwargs["config"]
     orch = Orch(cluster=ceph_cluster, **config)
+
+    client = ceph_cluster.get_nodes(role="client")[0]
+    clients = ceph_cluster.get_nodes(role="client")
+    executor = RadosBench(mon_node=client, clients=clients)
+
     try:
+        # Initiate thread pool to run rados bench
+        executor.run(config=config["benchmark"])
+
         # Set repo to newer RPMs
         orch.set_tool_repo()
 
@@ -53,14 +66,15 @@ def run(ceph_cluster, **kwargs) -> int:
         orch.monitor_upgrade_status()
 
         if config.get("verify_cluster_health"):
-            orch.cluster.check_health(
+            if orch.cluster.check_health(
                 rhbuild=config.get("rhbuild"), client=orch.installer
-            )
-
+            ):
+                raise UpgradeFailure("Cluster is in HEALTH_ERR state")
     except BaseException as be:  # noqa
         LOG.error(be, exc_info=True)
         return 1
     finally:
+        executor.teardown()
         # Get cluster state
         orch.get_cluster_state(
             [
--- a/ceph/rados/rados_bench.py
+++ b/ceph/rados/rados_bench.py
@@ -7,7 +7,7 @@ the low level native object storage API provided by Ceph.
 
 """
 import logging
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import ALL_COMPLETED, ThreadPoolExecutor, wait
 from time import time
 
 from ceph.ceph_admin.common import config_dict_to_string
@@ -125,8 +125,8 @@ class RadosBench:
         ex., rados bench 10 write -p test_bench  --no-cleanup --run-name test2
 
         Args:
-            client: client node (CephVMNode else client will be picked from self.clients)
-            pool_name: osd pool name (pool name else pool will be picked from self.pools)
+            client: client node
+            pool_name: osd pool name
             config: write ops command arguments
 
         config:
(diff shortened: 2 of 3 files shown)


tests/cephadm/test_cephadm_upgrade.py
```diff
@@ -3,10 +3,88 @@
...
+from concurrent.futures import ThreadPoolExecutor
+from time import time
 
 from ceph.ceph_admin.orch import Orch
 
-LOG = logging.getLogger()
+LOG = logging.getLogger(__name__)
+
+
+class RadosBench:
```
rakeshgm, line 14: i think this can lie some where outside this file. probably in utils or create a new rados.py file

tests/cephadm/test_cephadm_upgrade.py
```diff
@@ -3,10 +3,88 @@
...
+        self.pool = pool or self.create_pool()
+
+    @staticmethod
+    def create_run_name():
+        """
+        Return run name
+        """
+        return f"test-run-{int(time())}"
+
+    def create_pool(self):
```
rakeshgm, line 29: we can use PG values as parameters

tests/cephadm/test_cephadm_upgrade.py
```diff
@@ -3,10 +3,88 @@
...
+        """
+        Create pool and enable application
+        """
+        name = f"pool-{int(time())}"
+        self.client.exec_command(cmd=f"ceph osd pool create {name} 128 128", sudo=True)
```
rakeshgm, line 34: we can use PG values and pool names as well as parameters, also with a default value of course.

tests/cephadm/test_cephadm_upgrade.py
```diff
@@ -35,7 +113,27 @@ def run(ceph_cluster, **kwargs) -> int:
...
+
+        """
+        rados = RadosBench(ceph_cluster.get_nodes(role="client")[0], pool)
+        while not upgrade_completed:
+            try:
+                rados.seq(seconds)
+            finally:
+                rados.cleanup()
+        rados.delete_pool()
+        return "Done....."
```
rakeshgm, line 132: nice template though, we can add this too outside this file, probably make a separate package

tests/cephadm/test_cephadm_upgrade.py
psathyan: It is an individual entity hence it could be utils or lib.

psathyan: Could you add documentation about the object/class?

tests/cephadm/test_cephadm_upgrade.py
psathyan: In 5.x, we don't need to execute this with root level privileges... may we could add an attribute holding the RHCS version.

tests/cephadm/test_cephadm_upgrade.py
```diff
@@ -3,10 +3,88 @@
...
+        """
+        Delete pool
+        """
+        self.client.exec_command(
+            cmd="ceph config set global mon_allow_pool_delete true", sudo=True
+        )
+        self.client.exec_command(
+            cmd=f"ceph osd pool rm {self.pool} {self.pool} --yes-i-really-really-mean-it",
+            sudo=True,
+        )
```
psathyan, line 50: IMO,
- we could use a variable list holding the list of commands to be executed
- we could check if root level privileges is required.
- we could check if the configuration `mon_allow_pool_delete` is configured or not.
- we should fail if set to fails... so as to support negative testing.

tests/cephadm/test_cephadm_upgrade.py
```diff
@@ -3,10 +3,88 @@
...
+LOG = logging.getLogger(__name__)
+
+
+class RadosBench:
+    def __init__(self, client, pool=""):
+        """
+        Initialize RadosBench
+        """
+        self.client = client
+        self.pool = pool or self.create_pool()
```
psathyan, line 20: This looks more like an utility... hence, we could attempt not to initialize with client of pool... The attribute of this could be 
- start_time
- end_time
- interval
- max_time

psathyan (commented):
A good starting point

tests/cephadm/test_cephadm_upgrade.py
sunilkumarn417: > i think this can lie some where outside this file. probably in utils or create a new rados.py file

yeah, this part is in the plan now, I was just trying to utilize Radosbench for now. But definitely need to be part of libraries.
Need a plan for generic approach to handle this kind libraries.

-  Thank you.

tests/cephadm/test_cephadm_upgrade.py
sunilkumarn417: > we can use PG values and pool names as well as parameters, also with a default value of course.

yes, agree. will take care when this is added into library.

tests/cephadm/test_cephadm_upgrade.py
sunilkumarn417: > In 5.x, we don't need to execute this with root level privileges... may we could add an attribute holding the RHCS version.

I Agree.

sunilkumarn417: > IMO,
> 
> * we could use a variable list holding the list of commands to be executed

Agreed.

> * we could check if root level privileges is required.

well its depends upon the client nodes used and RHCS version.

> * we could check if the configuration `mon_allow_pool_delete` is configured or not.

I think this should be handled in some library associated with pool.

> * we should fail if set to fails... so as to support negative testing.

yes, we could check this. But need to define these libraries.

tests/cephadm/test_cephadm_upgrade.py
sunilkumarn417: we can discuss more about this.

sunilkumarn417:
**Latest test results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1624955020485/
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1624959864003/Upgrade_ceph_0.log (with ceph health check verification)

sunilkumarn417:
**Latest test results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1625550829888/Upgrade_ceph_0.log (upgrade PASSED)
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1625550665453/Upgrade_ceph_0.log (upgrade FAILED)

sunilkumarn417:
**Latest test results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1625645635443/Upgrade_ceph_0.log

psathyan (approved):
Looks like a good starting point...

PR #599, opened 2021-07-01
Adding Restart service operation in cephadm

preethisn:
# Description - Restarting a Ceph service using ceph orch command

### DIFF
--- a/ceph/ceph_admin/service_ops.py
+++ b/ceph/ceph_admin/service_ops.py
@@ -1,0 +1,140 @@
+"""Module that interfaces with ceph orch  CLI."""
+from typing import Dict, Optional
+
+from .common import config_dict_to_string
+from .typing_ import OrchProtocol
+
+
+def service_op(cls, op, config):
+    """
+       Execute the command ceph orch <start|stop|restart|reconfigure|redeploy> <service>.
+
+      Args:
+          cls: orchestration class
+          config: command and service are passed from the test case.
+          op: operation parameters ex: restart|start|stop|reconfigure|redeploy
+
+      Example:
+          Testing ceph orch restart mon
+
+    config:
+      command: restart
+      service: mon
+
+      Returns:
+          output, error   returned by the command.
+    """
+
+    base_cmd = ["ceph", "orch"]
+
+    if config.get("base_cmd_args"):
+        base_cmd_args_str = config_dict_to_string(config.get("base_cmd_args"))
+        base_cmd.append(base_cmd_args_str)
+
+    base_cmd.append(op)
+    base_cmd.append(config["service"])
+
+    return cls.shell(args=base_cmd)
+
+
+class ServiceOpsMixin:
+    """CLI for service operations using orch commands."""
+
+    def restart(self: OrchProtocol, config: Optional[Dict] = None):
+        """
+           Execute the command ceph orch restart  <args>.
+
+          Args:
+              config: command and service are passed from the test case.
+
+          Example:
+              Testing ceph orch restart mon
+
+        config:
+          command: restart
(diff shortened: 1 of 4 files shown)


suites/pacific/cephadm/tier1_ssl_dashboard_port.yaml
```diff
@@ -183,3 +183,13 @@ tests:
...
+  - test:
+      name: Restart Service
+      desc: Apply Ceph orch restart service for all the daemons
+      module: test_restart.py
+      polarion-id:
+      config:
+        command: restart
+        service: mon
+      destroy-cluster: false
+      abort-on-fail: true
```
psathyan, line 195: It's a good practice to end the file with a newline character.

tests/cephadm/test_restart.py
```diff
@@ -0,0 +1,37 @@
...
+      Returns:
+          output, error   returned by the command.
+    """
+    log.info(run.__doc__)
+    config = kw["config"]
+    cephadm = CephAdmin(cluster=ceph_cluster, **config)
+    arg = config["command"]
+    ser = config["service"]
+    cmd = f"ceph orch {arg} {ser}"
+    out, err = cephadm.shell([cmd])
```
psathyan, line 32: This is a good start however it moves away from the current implementation. The current behavior has a 1 to 1 mapping with the CLI command. Also, this would result in multiple test_<start|stop|reconfig> modules... higher maintenance cost.

IMHO, we should move the behavior as a mixin that could be inherited by `orch` module. This would allow services to customize further if required when a service operation is being performed.

Performing such operations would hamper re-usability when another module would like to perform restart/stop/reapply. Moving to the lib could avoid this issue.

psathyan:
Please run `isort .` from the cloned repo root direct. The PR contains code having unsorted import modules

psathyan:
@preethisn please let me know if you any assistance on this PR

preethisn:
logs are copied - http://magna002.ceph.redhat.com/ceph-qe-logs/Preethi/cephci-run-1625646303841/

ceph/ceph_admin/service_ops.py
```diff
@@ -0,0 +1,154 @@
+"""Module that interfaces with ceph orch  CLI."""
+from typing import Dict, Optional
+
+from .common import config_dict_to_string
+from .typing_ import OrchProtocol
+
+
+class ServiceOpsMixin:
+    """CLI for service operations using orch commands."""
+
```
psathyan, line 10: All the methods in this object have similar code with exception to the op i.e. `start | restart | reconfigure | redeploy`... would it be possible to create an internal method and pass the op as a parameter to it. Call this internal method varying the parameter. For example

```
def _service_op(cls, op, config):
  base_cmd = ["ceph", "orch"]
  if config.get("base_cmd_args"):
    ...
  base_cmd.append(op)
  base_cmd.append(config['service'])
  return cls.shell(args=base_cmd)

class ServiceOpMixin:

  def restart(...):
    return _service_op(self, 'restart', config)
```
This gives us the flexibility of overloading individual operations while having re-usability in place.

preethisn: Updated the changes. Logs attached - http://magna002.ceph.redhat.com/ceph-qe-logs/Preethi/cephci-run-1625744259988/

preethisn: Updated the changes. Logs attached - http://magna002.ceph.redhat.com/ceph-qe-logs/Preethi/cephci-run-1625744259988/

psathyan (approved):

sunilkumarn417 (approved):

PR #623, opened 2021-07-09
Modify testrbd.py to support diffrent branch checkout

ckulal:
Modify testrbd.py to clone based on branch given

logs:
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1626080525853/
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1626083263551/

# Description

# Checklist:

### DIFF
--- a/tests/rbd/test_rbd.py
+++ b/tests/rbd/test_rbd.py
@@ -29,12 +29,6 @@ def one_time_setup(node, rhbuild, branch: str) -> None:
         branch: The branch that needs to be cloned
         rhbuild: specification of rhbuild. ex: 4.3-rhel-7
     """
-    try:
-        node.exec_command(cmd="ls ceph")
-
-        return
-    except BaseException:  # noqa
-        pass
 
     os_ver = rhbuild.split("-")[-1]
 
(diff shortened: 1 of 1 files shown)


tests/rbd/test_rbd.py
```diff
@@ -31,7 +31,13 @@ def one_time_setup(node, rhbuild, branch: str) -> None:
     """
     try:
         node.exec_command(cmd="ls ceph")
+        out, err = node.exec_command(cmd="cd ceph; git branch")
```
psathyan, line 34: Instead of `git branch`, could you use `git rev-parse --abbrev-ref HEAD`? It would return only the branch name

rakeshgm:
The commit looks good, its nice to check on which branch we are on before doing a checkout of the specific branch. But i think we can definitely look for improvement here. 
If we clone ceph repo, the size is around 1 GB and cephci takes a long time to clone. so to reduce the size, time and instead of pulling all the branches, we can just do a clone of that branch and just take the recent commits from that branch using dept =1 

example:

```
git clone --branch <branch_name> --single-branch --depth 1 <repo>
```
you can even checkout only a specific folder too and not download all the src files which are not needed.

@ckulal can you take a look at it.

tests/rbd/test_rbd.py
```diff
@@ -31,8 +31,7 @@ def one_time_setup(node, rhbuild, branch: str) -> None:
     """
     try:
         node.exec_command(cmd="ls ceph")
-
-        return
+        node.exec_command(cmd="rm -rf ceph")
     except BaseException:  # noqa
         pass
```
psathyan, line 37: The try except block is not required as the plan is to remove the folder if found. Hence, we could avoid the block and have `node.exec_command(cmd="rm -rf ceph")`

psathyan (commented):
This look good... a minor improvement.. Another option is to add it to the list of `commands`

psathyan (approved):
LGTM

PR #628, opened 2021-07-11
Fix syntax error in post message

psathyan:

# Description
Global variables is not working hence moving to scoped variable. We definitely need to fix the `defaultFileDir` variable definition as maintenance effort is more when using the current implementation.

__Logs__
https://ceph-downstream-jenkins-csb-storage.apps.ocp4.prod.psi.redhat.com/view/RHCS%205%20QE/job/rhceph-5-tier-0/226/

### DIFF
--- a/pipeline/vars/common.groovy
+++ b/pipeline/vars/common.groovy
@@ -5,9 +5,6 @@
 
 import groovy.json.JsonSlurper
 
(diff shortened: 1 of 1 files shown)


AmarnatReddy (approved):
Looks Good

PR #634, opened 2021-07-12
install_prereq: document subscription-manager register options

ktdreyer:
Add a comment describing the `--serverurl` and `--baseurl` options for `subscription-manager register`.


rakeshgm:
@ktdreyer `--baseurl` and `--serverurl` are optional if we are planning to use cdn right?

ktdreyer:
Thanks @rakeshgm . You're right, both are optional. I updated the text to indicate that both are optional.

rakeshgm (approved):

PR #646, opened 2021-07-16
Remove Node with status error and expired subscription

anrao19:
Signed-off-by: Anuchaithra Rao <anrao@redhat.com>

# Description

# Checklist:

### DIFF
--- a/cleanup_env.py
+++ b/cleanup_env.py
@@ -1,0 +1,260 @@
+"""
+    Utility to cleanup instances in RHOS-D environment that have crossed the maximum
+    allowed duration and instances that are in error status. There are multiple
+    projects/tenants under CephQE purview and each of them have a different configuration.
+
+    For example, ceph-jenkins has the least allowable time as the intent is to enable
+    the pipeline is executed under a stable environment. The durations for each project
+    are
+        ceph-jenkins        3 days
+        ceph-ci             2 weeks
+        ceph-core           2 weeks
+"""
+import smtplib
+import sys
+from datetime import datetime, timezone
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+from typing import Dict
+
+import yaml
+from docopt import docopt
+from libcloud.common.types import LibcloudError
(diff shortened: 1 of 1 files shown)


psathyan:
Can we hold on to this for some time? I think there is a way to retrieve the User email address through libcloud.

cleanup_env.py
```diff
@@ -0,0 +1,228 @@
+""" Utility to cleanup instance/node along with its volumes"""
```
sunilkumarn417, line 1: hope this `--cleanup` option helpful?  https://github.com/red-hat-storage/cephci/blob/master/ceph/utils.py#L190 

I'm not sure it works with error/failed state. But we could make use of this code instead of creating new module.

anrao19: Sure sunil

cleanup_env.py
```diff
@@ -0,0 +1,264 @@
...
+            ] = "Notification: Your subscriptions are marked/removed in RHOS-D"
+            msg["From"] = sender
+            msg["To"] = recipient
+            html = """\
+                    <html><head><style>table, th, td {border: 1px solid black; }</style>
+                        </head><body>
+                            <p>Hi,<br><br>
+                            Below cloud subscriptions are expired/expiring.<br>
+                            The durations for subscriptions expire in project are for ceph-jenkins its 3 days,
+                            for ceph-ci and ceph-core its 2 weeks i.e, 14days.</p>"""
```
rakeshgm, line 173: try to use a template if you can.

rakeshgm:
some improvements I would like to suggest as a whole of coding construct. 
1. Avoid using `global` declarations.
2. We really don't need to mention the function parameter type in the definition and return types, we could add all those information in the doc string.  -- debatable though :)

cleanup_env.py
anrao19: Will take this as an enhancement in next Iteration.

psathyan (approved):
Looks good to me.

psathyan (approved):

PR #649, opened 2021-07-17
Fix 4.x Tier-0 suite failures on RHEL-7

psathyan:

# Description

In the case of RHEL-7, we are seeing inconsistent behaviors during the execution of the Ansible playbook. This is not seen on other platforms. However, if the ansible play is executed without cephci... it succeeds. There are instants where the execution is successful with cephci.

Hence in this PR, we are choosing to ignore the decode exception being thrown when `exec_command` is executed with `long_running` option enabled.

__Logs__
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-DFI0FN/ceph_ansible_0.log

### DIFF
--- a/ceph/ceph.py
+++ b/ceph/ceph.py
@@ -1345,6 +1345,7 @@ class CephNode(object):
             timeout = kw["timeout"]
         else:
             timeout = 120
+
         logger.info("Running command %s on %s", kw["cmd"], self.ip_address)
         stdin = None
         stdout = None
@@ -1352,6 +1353,7 @@ class CephNode(object):
         if self.run_once:
             self.ssh_transport().set_keepalive(15)
             self.rssh_transport().set_keepalive(15)
+
         if kw.get("long_running"):
             logger.info("long running command --")
             channel = ssh().get_transport().open_session()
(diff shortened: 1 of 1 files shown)


ceph/ceph.py
```diff
@@ -1364,12 +1364,18 @@ def exec_command(self, **kw):
                 rl, wl, xl = select([channel], [], [channel], 4200)
                 if len(rl) > 0:
                     data = channel.recv(1024)
-                    read += data.decode()
-                    logger.info(data.decode())
+
+                    # Consistently we are are hitting decode errors when executing the
+                    # plays on RHEL-7. This issue doesn't seem to surface for other
+                    # supported platforms. Moving to ignore as the value is in bytes.
+                    read += data.decode(errors="ignore")
```
sunilkumarn417, line 1371: Maybe we can reduce the use of decode.
```
data = channel.recv(1024)
data = data.decode(errors="ignore")
log.info(f"{data}")
read += data
```

sunilkumarn417 (commented):
Minor comments.

ceph/ceph.py
psathyan: done.

sunilkumarn417 (approved):

PR #657, opened 2021-07-22
Adding log file verification for cephadm

manasagowri:

# Description

- Added a method in test_bootstrap.py to verify the existence of cephadm.log and ceph-volume.log after bootstrap on all the nodes in the cluster.
- Added a config option **verify_log_files** to enable verification of logs after applying different roles like osd, rgw, etc by enabling logging.
- Added a method in test_cephadm.py to enable global logging using **ceph config set global log_to_file true** and verify that the logs have been generated for all the existing roles in every node.

# Checklist:

### DIFF
--- a/ceph/ceph_admin/helper.py
+++ b/ceph/ceph_admin/helper.py
@@ -1,12 +1,15 @@
 """
 Contains helper functions that can used across the module.
 """
+import json
 import logging
+import os
 import tempfile
 from os.path import dirname
 
 from jinja2 import Template
 
+from ceph.ceph import CommandFailed
 from ceph.utils import get_node_by_id, get_nodes_by_ids
 from utility.utils import generate_self_signed_certificate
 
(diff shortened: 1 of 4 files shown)


tests/ceph_installer/test_cephadm.py
```diff
@@ -49,6 +57,75 @@
 )
 
 
+def get_host_osd_map(cls):
```
sunilkumarn417, line 60: Can we move this to helper script file?

tests/ceph_installer/test_cephadm.py
```diff
@@ -49,6 +57,75 @@
...
+    out, err = cls.shell(args=["ceph", "osd", "tree"])
```
sunilkumarn417, line 61: I think there is a provision to get JSON output of `osd tree` using `-f json`, which will be easy to get data.

tests/ceph_installer/test_cephadm.py
sunilkumarn417: please add doc string

tests/ceph_installer/test_cephadm.py
```diff
@@ -49,6 +57,75 @@
...
+        osd_dict[osd[0]] = list()
+        for i in osd[1:]:
+            temp = i.split()
+            for x in temp:
+                if "osd" in x:
+                    osd_dict[osd[0]].append(x)
+    return osd_dict
+
+
+def validate_log_file_after_enable(cls):
```
sunilkumarn417, line 75: can we move this to helper module?

tests/ceph_installer/test_cephadm.py
```diff
@@ -49,6 +57,75 @@
...
+    """
+    Method to verify generation of log files in default log directory when logging not enabled
+    Args:
+        cls: cephadm instance object
+
+    Returns:
+        string
```
sunilkumarn417, line 82: Returns: Boolean

tests/ceph_installer/test_cephadm.py
```diff
@@ -49,6 +57,75 @@
...
+    """
+    out, err = cls.shell(
+        args=["ceph", "config", "set", "global", "log_to_file", "true"]
+    )
+    out, err = cls.shell(args=["ceph", "fsid"])
+    fsid = out.strip()
+    log_file_path = os.path.join("/var/log/ceph", fsid)
+    roles_to_validate = ["mon", "mgr", "osd"]
```
sunilkumarn417, line 90: IMO, Most daemons produces log files, So its good to go with current daemons and its respective log file.
Hope daemon name and log file name is almost same.

tests/ceph_installer/test_cephadm.py
```diff
@@ -49,6 +57,75 @@
...
+    osd_dict = get_host_osd_map(cls)
+
+    for node in cls.cluster.get_nodes():
+        try:
+            if "client" in node.role and len(node.role) == 1:
+                continue
+            out, _ = node.exec_command(
+                cmd=f"ls -l {log_file_path} | egrep -v '^d'", sudo=True
```
sunilkumarn417, line 98: I think Instead of getting list of files, its good to construct log file path and verify it.

tests/ceph_installer/test_cephadm.py
```diff
@@ -49,6 +57,75 @@
...
+            )
+            files = out.read().decode().split("\n")
+            files = list(filter(None, files))
+            LOG.info("Files in log folder : %s" % ", ".join(files))
+            roles = []
+            [
+                roles.append(role)
+                for role in node.role
```
sunilkumarn417, line 106: IMO, This might fail if roles are not deployed before calling log file verification.

tests/ceph_installer/test_cephadm.py
```diff
@@ -49,6 +57,75 @@
...
+                if role not in roles and role in roles_to_validate
+            ]
+            for role in roles:
+                if role == "osd":
```
sunilkumarn417, line 110: Maybe we could avoid if we go with daemon names and construct the log file path.

tests/ceph_installer/test_cephadm.py
manasagowri: Done

tests/ceph_installer/test_cephadm.py
manasagowri: Thanks for this suggestion. I have used it.

tests/ceph_installer/test_cephadm.py
manasagowri: Done

tests/ceph_installer/test_cephadm.py
manasagowri: Done.

tests/ceph_installer/test_cephadm.py
manasagowri: when we enable logging using the command 
ceph config set global log_to_file true
as we discussed for this task, only MON, MGR, OSD, RGW and MDS logs are getting enabled. I will add validation for these roles.

tests/ceph_installer/test_cephadm.py
manasagowri: It is difficult to generate log file names for VMs, since for MGR, MDS and RGW nodes, a random string is added to the log file name as shown below.
**[root@ceph-sshtest-1626245934930-node1-installer-mon-mgr-osd-node-exp f64f341c-655d-11eb-8778-fa163e914bcc]# ls |grep mg
ceph-mgr.ceph-sshtest-1626245934930-node1-installer-mon-mgr-osd-node-exp.jzijtp.log**

tests/ceph_installer/test_cephadm.py
manasagowri: Added a check to verify if node.role is present.

tests/ceph_installer/test_cephadm.py
manasagowri: Like mentioned in the above comment, it is not possible to construct the complete log file path in case of VMs.

tests/ceph_installer/test_cephadm.py
manasagowri: Thanks for the suggestion Sunil. Added verification by constructing log file path.

tests/ceph_installer/test_cephadm.py
manasagowri: Made the necessary changes.

sunilkumarn417:
@manasagowri can you please include test results.

ceph/ceph_admin/helper.py
```diff
@@ -378,3 +386,122 @@ def get_cluster_state(cls, commands=[]):
...
+                continue
+            daemons = daemon_valid[node.hostname]
+            for daemon in daemons:
+                file = os.path.join(log_file_path, f"ceph-{daemon}.log")
+                if "rgw" in daemon:
+                    file = os.path.join(log_file_path, f"ceph-client.{daemon}.log")
+                LOG.info(f"Verifying existence of log file {file} in host {node.hostname}")
+                fileExists = file_or_path_exists(node, file)
+                if not fileExists:
+                    raise BootStrapValidationFailure(
```
sunilkumarn417, line 498: Raise and return are in one block, return wont get executed here.

ceph/ceph_admin/helper.py
```diff
@@ -12,6 +16,10 @@
 LOG = logging.getLogger(__name__)
 
 
+class BootStrapValidationFailure(Exception):
```
sunilkumarn417, line 19: can we define or use this exception in test_bootstrap.py?

tests/cephadm/test_bootstrap.py
```diff
@@ -451,6 +436,50 @@ def validate_ssh_private_key(cls, ssh_private_key):
...
+                if fileExists:
+                    break
+            if not fileExists:
+                raise BootStrapValidationFailure(
+                    f"ceph-volume.log is not present in the node {node.ip_address}"
+                )
+            log.info(f"Log verification on node {node.ip_address} successful")
+        except CommandFailed as err:
+            log.error("Error: %s" % err)
+            return False
```
sunilkumarn417, line 479: return wont have any effect, Since we are not raising exception in `def run()`.

sunilkumarn417: please raise exception here only.

ceph/ceph_admin/helper.py
manasagowri: Done. Removed raise.

sunilkumarn417 (approved):
Thanks @manasagowri for adding log verification part in the 5x tests.

LGTM.

PR #691, opened 2021-08-03
rgw: enable ssl to t2 suite

rakeshgm:
enable ssl setup on t2 suite

[logs - 5x](http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-Z4JV0F)

### DIFF
--- a/suites/pacific/rgw/tier_2_rgw_regression.yaml
+++ b/suites/pacific/rgw/tier_2_rgw_regression.yaml
@@ -85,13 +85,17 @@ tests:
               args:
                 all-available-devices: true
           - config:
-              command: apply
-              service: rgw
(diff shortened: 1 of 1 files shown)


psathyan:
@rakeshgm This would break S3-Tests. IMO, let's pull out the S3Tests and have it in a separate suite. We still continue to have a timed run that includes the s3test as a separate stage.

rakeshgm:
@psathyan i agree on s3tests, I would not like to create another suite just for s3tests. but nonetheless, some of the tests here failed when ssl is enabled. http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-A9P6GF/

need to check before closing the PR.

PR #725, opened 2021-08-18
Added fix for others VM getting deleted if mentioned instance name is available as part of existing instances

vivekanandan-k-rh:
Added fix for others VM getting deleted if mentioned instance name is available as part of existing instances

Example:

If there is a set of VMs available in the name of s3cmd and if instance name is given as cmd, all VMs with the name s3cmd also getting destroyed.

### DIFF
--- a/run.py
+++ b/run.py
@@ -164,7 +164,8 @@ def create_nodes(
         )
 
     log.info("Destroying existing osp instances..")
-    cleanup_ceph_nodes(osp_cred, instances_name)
+    instance_name = f"-{instances_name}-"
+    cleanup_ceph_nodes(osp_cred, instance_name)
     ceph_cluster_dict = {}
 
     log.info("Creating osp instances")


run.py
```diff
@@ -164,7 +164,10 @@ def create_nodes(
         )
 
     log.info("Destroying existing osp instances..")
-    cleanup_ceph_nodes(osp_cred, instances_name)
+    for cluster in conf.get("globals"):
+        cluster_name = cluster.get("ceph-cluster").get("name", "ceph")
+        instance_name = f"{cluster_name}-{instances_name}-"
+        cleanup_ceph_nodes(osp_cred, instance_name)
```
psathyan, line 170: It is a good idea to be specific when we are attempting to redeploy the environment. 

However, this becomes a costly operation as we would be making 2x calls when we could have done with 2. Instead, we could opt to use `instance_name = f"-{instances_name}-"`

run.py
```diff
@@ -312,7 +315,8 @@ def run(args):
             osp_cred = yaml.safe_load(osp_cred_stream)
 
     if cleanup_name is not None:
-        cleanup_ceph_nodes(osp_cred, cleanup_name)
+        instance_name = f"-{cleanup_name}-"
```
psathyan, line 318: IMO, if the user is providing this value with cleanup, we must allow support for partial matches as this is done with the user knowledge. I would like to not reduce this scope/behavior

psathyan (commented):
I don't think this addresses the linked issue wherein the user would like to have a confirmation before any instances are removed when they are deploying/testing a cluster.

vivekanandan-k-rh:
> I don't think this addresses the linked issue wherein the user would like to have a confirmation before any instances are removed when they are deploying/testing a cluster.

@psathyan , I have modified like instead of asking for the confirmation, It will delete the VMs which is having only that instances name instead of other VMs having that text.

run.py
vivekanandan-k-rh: Done. Moved back to existing behavior

run.py
vivekanandan-k-rh: Done. Modified

AmarnatReddy:
Hi @vivekanandan-k-rh,
Can we have an Option something like **--dry-run** which will show all the Vms which will get deleted as part of the run.
Later if the User is ok with the deletion list. he can run it without **--dry-run**.
Just a suggestion from my side.

vivekanandan-k-rh:
Closing this PR as different is planned for this issue (using --dry-run as suggested by Amarnath). Will raise different PR for the same

PR #776, opened 2021-09-08
Build information updates for corner cases

psathyan:

# Description

In this PR, we avoid clearing out the available section and update only the required fields know. For example,

Case 1
The build version is received is higher than the current one. At present, we clear the `latest` section and add only details of the latest platform RPM. This causes a problem when there is no RHEL-7 build or another platform build.

Case 2
There is no repository information but only compose builds.

Also, prevents the compose from sending to `promote.complete`.

__Logs__
https://ceph-downstream-jenkins-csb-storage.apps.ocp4.prod.psi.redhat.com/view/RHCS%20QE/job/rhceph-unsigned-compose-listener/

Builds: 40 to 45


tintumathew10 (approved):
Looks good to me

PR #782, opened 2021-09-09
[Bz 1997080]: fixes docker-to-podman issue in tier1

Ameenasuhani:
Signed-off-by: ameenasuhani <ameenasuhani@gmail.com>

# Description
The add-mon.yml playbook only targets the mons and osds nodes and doesn't included the dashboard bits.
So adding a new monitor via this playbook won't deploy node-exporter on that node.
You can achieve the same by re-running the main playbook with --limit mons option (like you did for osds/rgws).

Bz: https://bugzilla.redhat.com/show_bug.cgi?id=1997080
Doc bz:https://bugzilla.redhat.com/show_bug.cgi?id=2002293

Logs: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-UVK46F/config_roll_over_mon_daemon_0.log

### DIFF
--- a/tests/ceph_ansible/test_ansible_roll_over.py
+++ b/tests/ceph_ansible/test_ansible_roll_over.py
@@ -55,14 +55,10 @@ def run(ceph_cluster, **kw):
             matched_short_names = list(filter(matcher.match, short_name_list))
             if len(matched_short_names) > 1:
                 raise RuntimeError(
-                    "Multiple nodes are matching node-name {node_name}: \n{matched_short_names}".format(
-                        node_name=node_name, matched_short_names=matched_short_names
-                    )
+                    f"Multiple nodes are matching node-name {node_name}: \n{matched_short_names}"
                 )
             if len(matched_short_names) == 0:
-                raise RuntimeError(
-                    "No match for {node_name}".format(node_name=node_name)
-                )
+                raise RuntimeError(f"No match for {node_name}")
             for ceph_node in ceph_cluster:
                 if ceph_node.shortname == matched_short_names[0]:
                     matched_ceph_node = ceph_node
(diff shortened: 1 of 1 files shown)


tests/ceph_ansible/test_ansible_roll_over.py
```diff
@@ -145,6 +145,13 @@ def run(ceph_cluster, **kw):
     )
     out, rc = ceph_installer.exec_command(cmd=cmd, long_running=True)
 
+    if "4.2" in build:
```
sunilkumarn417, line 148: can we have this statement,
```
if build == "4.2":
```

tests/ceph_ansible/test_ansible_roll_over.py
```diff
@@ -145,6 +145,13 @@ def run(ceph_cluster, **kw):
...
+        if daemon == "mon":
+            cmd = "cd {} ; ANSIBLE_STDOUT_CALLBACK=debug;ansible-playbook -vvvv -i hosts {} --limit mons".format(
```
sunilkumarn417, line 150: can we use f-strings,
```
cmd = f"cd {ansible_dir} ; ANSIBLE_STDOUT_CALLBACK=debug;ansible-playbook -vvvv -i hosts {yaml} --limit mons" ```

sunilkumarn417 (commented):
Hi @Ameenasuhani ,

Changes looks good. but few minor changes would be required.

can we have complete test suite run which also executed docker-to-podman playbook   successfully.

-Thanks

tests/ceph_ansible/test_ansible_roll_over.py
Ameenasuhani: build will have 4.2-rhel-7 or 4.2-rhel-8 so this condition will fail

Ameenasuhani:
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-34S52P/

tests/ceph_ansible/test_ansible_roll_over.py
psathyan: We could have `if build.startswith("4.2")` not that the current implementation wouldn't work. I prefer not to change it to `build == '4.2'` now.

psathyan (approved):
Looks good to me

sunilkumarn417 (approved):
lgtm

tests/ceph_ansible/test_ansible_roll_over.py
sunilkumarn417: > build will have 4.2-rhel-7 or 4.2-rhel-8 so this condition will fail

Sorry, Just had pipeline v2.0 in mind. 
Make sense to hold back to current logic.

PR #787, opened 2021-09-13
Add support to upload binaries to IBM COS

psathyan:

# Description

As part of supporting execution of CephQE tests in environment hosted in IBM-C, this PR adds support to copy the development packages to COS.

__Logs__
http://magna002.ceph.redhat.com/cephci-jenkins/ibm_cos/

Post fixing review comments :a: http://magna002.ceph.redhat.com/cephci-jenkins/ibm_cos/logs_after_review_1.log


AmarnatReddy (dismissed):
Looks fine for me

storage/ibm_cos.py
```diff
@@ -0,0 +1,211 @@
...
+class DownloadException(Exception):
+    pass
+
+
+class CloudObjectStorage:
+    """Interface to IBM's Cloud Object Storage."""
+
+    def __init__(self):
+        """Initialize the instance using global configuration."""
+        self._conf = get_cephci_config()["cos"]
```
ktdreyer, line 52: Let's document this new `cos` setting in the sample configuration file, so it's clear how users should use this.

pipeline/scripts/ci/upload_compose.py
```diff
@@ -0,0 +1,151 @@
...
+enabled = 0
+gpgcheck = 0
+
+{% endfor %}
+"""
+
+usage = """Upload Compose.
+
+This script pulls the development RPMs available in the given base_url and uploads an
+archive of it using the provided object_name to the mentioned bucket.
```
ktdreyer, line 41: When I first read this, I thought:

1. "Why not simply use `aws cp --recursive`, or `s3cmd put --recursive`, or some other simple recursive copy tool?"

2. "Why are we using tarballs instead of uploading the compose contents directly as-is?"

I think there are answers to these questions. Let's document them in the code here.

pipeline/scripts/ci/upload_compose.py
```diff
@@ -0,0 +1,151 @@
...
+
+In the context of CephCI,
+    bucket_name     is the name of the RHCS build
+    object_name     is the ceph version
+    base_url        the base url containing the RPMs to be uploaded
+
+  Example:
+    python upload_compose.py rhceph-4.2-rhel-7 ceph-14.6.2-113 http://download.rh.com/x
```
ktdreyer, line 49: Is `base_url` supposed to be the base of a compose, or a MON/OSD/Tools subdirectory? Let's clarify this in the documented example here. Maybe just use a real URL here.

pipeline/scripts/ci/upload_compose.py
```diff
@@ -0,0 +1,151 @@
...
+        subprocess.run(
+            [
+                "reposync",
+                "-l",
+                "-d",
+                "-m",
+                "--newest-only",
+                "--download-metadata",
+                "--repoid",
+                f"{build}-{repo}",
```
ktdreyer, line 105: You can define repos dynamically on the CLI here without having to write them into `/etc/` beforehand. The CLI flag is `--repofrompath`.

Here's an example that dynamically defines a repo named "foobar", and then syncs all the content from that:

```
reposync \
  -l \
  -m \
  --newest-only \
  --download-metadata \
  --repoid foobar \
  --repofrompath=foobar,https://download.example.com/ceph/x86_64/ \
  --download-path /tmp/brew
```

I think that will allow you to remove the `create_repo()` method

pipeline/scripts/ci/upload_compose.py
```diff
@@ -0,0 +1,151 @@
...
+    temp_dir = tempfile.mkdtemp()
+    _file = "".join(random.choices(string.ascii_lowercase + string.digits, k=5))
+    temp_file = join("/tmp", f"{_file}.tar.gz")
+
+    for repo in ["OSD", "MON", "Tools"]:
+        subprocess.run(
+            [
+                "reposync",
+                "-l",
+                "-d",
```
ktdreyer, line 100: `-d`, `-m` and `--download_path` are not valid options for reposync on my computer - need to fix this here.

storage/ibm_cos.py
```diff
@@ -0,0 +1,211 @@
...
+        self._api_key = self._conf["api-key"]
+        self._resource_id = self._conf["resource-id"]
+
+        self.endpoint = self._conf["endpoint"]
+        self.location_constraint = dict(
+            {
+                "LocationConstraint": self._conf.get(
+                    "location-constraint", "au-syd-standard"
```
ktdreyer, line 60: It would be great to document why we're choosing these magic strings here

storage/ibm_cos.py
```diff
@@ -0,0 +1,211 @@
...
+            "s3",
+            ibm_api_key_id=self._api_key,
+            ibm_service_instance_id=self._resource_id,
+            config=Config(signature_version="oauth"),
+            endpoint_url=self.endpoint,
+        )
+
+    def create_bucket(self, name: str) -> None:
+        """
+        Create a bucket with provide name.
```
ktdreyer, line 83:
```suggestion
        Create a bucket with provided name.
```

storage/ibm_cos.py
```diff
@@ -0,0 +1,211 @@
...
+        except BaseException as be:  # no-qa
+            logging.debug(be)
+
+        raise ApplyBucketLifeCycleException(
+            f"Failed to apply default lifecycle policy on bucket - {name}."
+        )
+
+    def bucket_exists(self, name: str) -> bool:
+        """
+        Returns the existences of the provided bucket.
```
ktdreyer, line 145:
```suggestion
        Check whether a bucket exists or not
```

pipeline/scripts/ci/upload_compose.py
```diff
@@ -0,0 +1,151 @@
...
+    Uploads the given file to the provided bucket with the mentioned name.
+
+    Args:
+        file_name: str  Complete path to the file that needs to be uploaded
+        bucket:str      The name of container to which file has to be uploaded
+        item: str       The name to be used for the uploaded object
+    """
+    cos = CloudObjectStorage()
+
+    if not cos.bucket_exists(bucket):
```
ktdreyer, line 129: This type of "check before creating" is the Look Before You Leap pattern. I think it would be simpler and less racy to try create the bucket every time, and then just catch the exception if it fails in a way you don't expect.

https://stackoverflow.com/questions/26871884/how-can-i-easily-determine-if-a-boto-3-s3-bucket-resource-exists indicates that there may be a way to do this with boto that simply gives you the bucket, without raising an exception?

pipeline/scripts/ci/upload_compose.py
psathyan: The options are different in RHEL-8 and centos-7. This could be the reason why one would need be able to see them. I have added a note in the module docstring. It is one of the reason why I have introduced `check=True`.

pipeline/scripts/ci/upload_compose.py
psathyan: `--repofrompath` is unavailable in `centos-7` hence we do need the `create_repo()` method. During our earlier testing, we couldn't use `rhel-8` as we had issues in sending emails from `cephci`.

We have disabled sending emails from `cephci`... let me add a user story/task to have us migrate the Jenkins node to `rhel-8`

psathyan: Once we have moved to `rhel-8`, I would be able to address/use `repofrompath`. I does simplify the workflow. Thanks for the feedback.

AmarnatReddy (approved):

pipeline/scripts/ci/upload_compose.py
ktdreyer: Oh, I see now. Sounds like there are large differences from RHEL 8 onwards.

AmarnatReddy (dismissed):

AmarnatReddy (approved):

PR #847, opened 2021-10-05
Implementation of IBM softlayer modules

ckulal:

# Description

# Checklist:


compute/ibm_vpc.py
```diff
@@ -0,0 +1,451 @@
...
+
+class NodeError(Exception):
+    pass
+
+
+class NodeDeleteFailure(Exception):
+    pass
+
+
+class CephVMNodeV2:
```
psathyan, line 69: Would it make sense not use `CephVMNodeV2`?

compute/ibm_vpc.py
```diff
@@ -0,0 +1,451 @@
...
+        ]
+        with parallel() as p:
+            for instance in del_instance_list:
+                LOG.info(
+                    "Destroying node {node_name} with {timeout} timeout".format(
+                        node_name=instance["name"], timeout=timeout
+                    )
+                )
+                p.spawn(self.service.delete_instance, instance["id"])
+                self.wait_until_node_del(instance["id"])
```
psathyan, line 199: IMO, this method definition would be in`utils.utils` . Secondly, we should wait for the node to be delete either inside `delete_instance` or outside of the `for loop`... when doing outside of the `for loop`, it would need to check for all the nodes in the list

compute/ibm_vpc.py
```diff
@@ -0,0 +1,451 @@
...
+
+
+class NodeDeleteFailure(Exception):
+    pass
+
+
+class CephVMNodeV2:
+    """Represent the VMNode required for cephci."""
+
+    def __init__(self, accessKey: str, **kwargs) -> None:
```
psathyan, line 72: `**kwargs` is not used... do we need it to be an argument?

compute/ibm_vpc.py
```diff
@@ -0,0 +1,451 @@
...
+
+def get_ibm_service(accessKey: str):
+    """
+    Get ibm service
+    Args:
+        accessKey    The access key(API key) of the user.
+    """
+    authenticator = IAMAuthenticator(accessKey)
+    service = VpcV1(authenticator=authenticator)
+    service.set_service_url("https://au-syd.iaas.cloud.ibm.com/v1")
```
psathyan, line 28: I believe it would be good if we got this endpoint from the global configuration file.

compute/ibm_vpc.py
```diff
@@ -0,0 +1,451 @@
...
+            json_obj            json object
+        """
+        resource_url = json_obj["first"]["href"]
+        resource_list_name = re.search(r"v1/(.*?)\?", resource_url).group(1)
+        resource_id = [
+            i["id"]
+            for i in json_obj[resource_list_name]
+            if i["name"] == name_of_resource
+        ][0]
+        return resource_id
```
psathyan, line 100: Let's not traverse through the complete list. Instead return as soon as a match is found.

compute/ibm_vpc.py
```diff
@@ -0,0 +1,451 @@
...
+    def get_id_by_name_zones(self, name_of_zone, json_obj):
+        """
+        Gets the Zone ID with Zone name
+        Args:
+            name_of_zone    Name of the zone.
+            json_obj        json object
+        """
+        resource_id = [
+            i["id"] for i in json_obj["dnszones"] if i["name"] == name_of_zone
+        ][0]
```
psathyan, line 111: Please avoid traversing the entire list. Return as soon as we match a resource.

compute/ibm_vpc.py
```diff
@@ -0,0 +1,451 @@
...
+            subnet_identity_model["id"] = subnet_id
+
+            # Construct a dict representation of a NetworkInterfacePrototype model
+            network_interface_prototype_model = {}
+            network_interface_prototype_model["allow_ip_spoofing"] = False
+            network_interface_prototype_model["subnet"] = subnet_identity_model
+
+            # Construct a dict representation of a InstanceProfileIdentityByName model
+            instance_profile_identity_model = {}
+            instance_profile_identity_model["name"] = "bx2-2x8"
```
psathyan, line 273: Can we avoid hard coding the VM size (flavor)?

compute/ibm_vpc.py
```diff
@@ -0,0 +1,451 @@
...
+            i for i in instance_list.get_result()["instances"] if pattern in i["name"]
+        ]
+        with parallel() as p:
+            for instance in del_instance_list:
+                LOG.info(
+                    "Destroying node {node_name} with {timeout} timeout".format(
+                        node_name=instance["name"], timeout=timeout
+                    )
+                )
+                p.spawn(self.service.delete_instance, instance["id"])
```
psathyan, line 198: Do we remove the volumes and DNS entries?

compute/ibm_vpc.py
```diff
@@ -0,0 +1,451 @@
...
+
+    @property
+    def role(self) -> List:
+        """Return the Ceph roles of the instance."""
+        return self._roles
+
+    @role.setter
+    def role(self, roles: list) -> None:
+        """Set the roles for the VM."""
+        from copy import deepcopy
```
psathyan, line 449: Avoid doing an import statement here.

compute/ibm_vpc.py
```diff
@@ -0,0 +1,451 @@
...
+    @property
+    def volumes(self) -> List:
+        """Return the list of storage volumes attached to the node."""
+        if self.node is None:
+            return []
+        # Removing boot volume from the list
+        volume_attachments = []
+        for i in self.node["volume_attachments"]:
+            if i["volume"]["name"].startswith(self.node["name"]):
+                volume_attachments.append(i)
```
psathyan, line 420: Is there a better way to identify if the volume is boot or not?

compute/ibm_vpc.py
ckulal: Done

psathyan (commented):
Looks good to me... let's wait till we have the DNS code in.

psathyan (approved):
Lets get it merged.

PR #851, opened 2021-10-06
[tier1][cephfs]directory pinning with active-active MDS

anrao19:
Signed-off-by: Anuchaithra Rao <anrao@redhat.com>

# Description

# Checklist:


hkadam134 (dismissed):

tests/cephfs/dir_pinning.py
```diff
@@ -0,0 +1,369 @@
...
+        config = kw.get("config")
+        fs_util = FsUtils(ceph_cluster)
+        fs_util_v1 = FsUtilsV1(ceph_cluster)
+        clients = ceph_cluster.get_ceph_objects("client")
+        build = config.get("build", config.get("rhbuild"))
+        client_info, rc = fs_util.get_clients(build)
+        if rc == 0:
+            log.info("Got client info")
+        else:
+            raise CommandFailed("fetching client info failed")
```
psathyan, line 27: We could do this
```
if rc != 0:
  raise CommandFailed("...")
log.info("Retrieved client information successfully")
```

tests/cephfs/dir_pinning.py
```diff
@@ -0,0 +1,369 @@
+import logging
```
psathyan, line 1: It is a good practice to provide information about the test module. What are the pre-requisites, arguments, workflow, etc...

tests/cephfs/dir_pinning.py
```diff
@@ -0,0 +1,369 @@
...
+                "readwrite",
+            )
+            p.spawn(fs_util.read_write_IO, client3, client_info["mounting_dir"])
+            for op in p:
+                return_counts, rc = op
+
+        result = fs_util.rc_verify("", return_counts)
+
+        client1[0].exec_command(
+            cmd="sudo mkdir %s%s" % (client_info["mounting_dir"], "testdir")
```
psathyan, line 99: use `format-string`

psathyan (commented):
It is a good step.

General comments
 - docstrings
 - use `format-strings`
 - there are instances where an `else` statement can be avoided.

Note: For now, we could use `p.spawn` however there are too many of them. I am not sure how that is going to impact. However, in the long run we would definitely need to have a way to execute tests in parallel. Hence, not provided any comments or thoughts on it.

psathyan (approved):

PR #860, opened 2021-10-12
OKR 1994827: skip ansible task with skip tags "wait for all OSDs to be up" option

sunilkumarn417:
https://bugzilla.redhat.com/show_bug.cgi?id=1994827

- Added provision to include multiple ansible CLI args.
- test suite to run rolling update with `--skip-tags wait_all_osds_up`.

Test results:
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-213G1S


Ameenasuhani:
LGTM

Do we have any plan to cover the other two scenarios of bringing down the osds which was covered during testing of this bz?

Also, can you add polarion id to the test case and mark it automated

sunilkumarn417:
> LGTM
> 
> Do we have any plan to cover the other two scenarios of bringing down the osds which was covered during testing of this bz?
> 
> Also, can you add polarion id to the test case and mark it automated

Currently, the positive test case scenario automated and Don't think other scenarios will be much useful to test `skip-tags` option.
And Polarion ID is already updated in upgrade test case and marked as automated. 

-Thanks.

Ameenasuhani (approved):

PR #912, opened 2021-11-12
Rados tier-2 tests: EC Pool recovery with k shards

pdhiran:

# Description

# Checklist:

### DIFF
--- a/suites/pacific/rados/tier_2_rados.yaml
+++ b/suites/pacific/rados/tier_2_rados.yaml
@@ -196,6 +196,25 @@ tests:
       desc: EC pool with Overwrites & create RBD pool
 
   - test:
+      name: EC Pool Recovery Improvement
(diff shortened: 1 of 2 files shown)


pdhiran:
http://magna002.ceph.redhat.com/ceph-qe-logs/pawan/cephci-run-VEAS35/

psathyan (approved):
Looks good to me.

There are 15 test cases... would it make sense to split the workflow into smaller bits so that we can have a quicker feedback cycle.

pdhiran:
> Looks good to me.
> 
> There are 15 test cases... would it make sense to split the workflow into smaller bits so that we can have a quicker feedback cycle.

Hi Satya, Yes, We will split the tests into smaller sets. That makes sense.
We will follow the multi-suite approach to this and submit the new suites with next PR.

PR #921, opened 2021-11-16
 fragments/rgw: added rgw fragments

udaysk23:
Added RGW fragments and created T0 directory in suite/nautilus/rgw/t0/


udaysk23:
Will add TC logs shortly

fragments/rgw/test_versioning_objects_enable.yaml
```diff
@@ -0,0 +1,10 @@
+tests:
```
psathyan, line 1: We could further structure the folder to `rgw/versioning/test_objects_enable.yaml`

psathyan:
Sorry, the guidelines are not available yet. Hence, we can discuss on how we want to do this

- All fragments must live in fragments folder at root level
- Group fragments into containers/folders/dirs based on purpose, feature, etc
- Descriptive fragment name
- At all costs, avoid duplicates
- Ensure each fragment has a Polarion ID

rakeshgm:
> Sorry, the guidelines are not available yet. Hence, we can discuss on how we want to do this
> 
> * All fragments must live in fragments folder at root level
> * Group fragments into containers/folders/dirs based on purpose, feature, etc
> * Descriptive fragment name
> * At all costs, avoid duplicates
> * Ensure each fragment has a Polarion ID

I think we should rephrase the meaning of the last point. so there may be a fragment which is not a test case, for example, install pre-req and cluster-health check frags. these don't need polarion id. I think we should make polarion id's as mandatory if the fragment is a test. 

I agree with all the other points

udaysk23:
Due to commit issue, Closing this PR

PR #928, opened 2021-11-17
Scheduled scrub automation

SrinivasaBharath:
# Description

<details>

- [x] Create a test case in Polarion reviewed and approved.
- [x] Create a design/automation approach doc. Optional for tests with similar tests already automated.
- [x] Implement the test script and perform test runs
- [x] Update Polarin Test with Automation script details and update automation fields
</details>

### DIFF
--- a/ceph/rados/rados_scrub.py
+++ b/ceph/rados/rados_scrub.py
@@ -1,0 +1,119 @@
+"""
+   This module contains the methods required for scrubbing.
+
+   1.To set the parameters for scrubbing initially required the
+     cluster time and day details.get_cluster_date method provides
+     the details
+
+   2.set_osd_configuration method  used to set the configuration
+     parameters on the cluster.
+
+   3.get_osd_configuration  method is used to get the configured parameters
+     on the cluster.
+
+     NOTE: With set_osd_configuration & get_osd_configuration methods can
+          use to set the get the any OSD configuration parameters.
+
+   4. get_pg_dump  method is used to get the pg dump details from the cluster
+
+   5. verify_scrub  method used for the verification of scheduled scrub
+      happened or not.
+"""
+
+import logging
+from collections import defaultdict
+
+from ceph.rados.core_workflows import RadosOrchestrator
+
+log = logging.getLogger(__name__)
+
+
+class RadosScrubber(RadosOrchestrator):
+    def get_osd_configuration(self, param):
+        """
+        Used to get the osd parameter value
+        Args:
+            params : Parameter to get the value
+
+        Returns : parameter value
+        """
+
+        log.info(f'{"Getting  osd configurations"}')
+        cmd = f"ceph config get osd {param}"
+        out = super().run_ceph_command(cmd=cmd)
+        return out
+
+    def set_osd_configuration(self, param, value):
+        """
+        Used to set the configuration parametrs to the OSD's.
+        Args:
+            params : Parameters to be set for OSD's
+
+        Returns: 0 or 1
+        """
+        cmd = f"ceph config set  osd {param} {value}"
+        out, err = self.node.shell([cmd])
+        actual_value = self.get_osd_configuration(param)
+        if actual_value == value:
+            log.info(f"Configuration parameter {param} is set with {value} on osds")
+            return 0
+        else:
+            return 1
+
+    def get_pg_dump(self, *args):
+
+        """
+        Used to get the pg dump logs
+
+        Args:
+            1. If column value is null then return the pg dump output
+            2. If column is any of the pgid,state,last_scrub_stamp,
+               last_deep_scrub_stamp,last_active ..e.t.c value then
+                returns the details about that column
+        Return:
+              dictionary
+
+        """
+
+        cmd = "ceph pg dump "
+        column_dict = defaultdict(list)
+
+        log.info(f'{"Getting  PG dump of the cluster"}')
+        pgDump = super().run_ceph_command(cmd=cmd)
+        if args:
+            columns = list(args)
+            for detail in pgDump["pg_map"]["pg_stats"]:
+                for column in columns:
+                    column_dict[column].append(detail[column])
+            return column_dict
+        else:
+            return pgDump
+
+    def verify_scrub(self, before_scrub_data, after_scrub_data):
+        """
+        Used to validate the scrubbing done or not
+
+        Args:
+            1.before_scrub_data - It is dictionary which conatin the
+              pgId and last scrub time data before scrubbing.
+            2.after_scrub_data - It is a dictionary that conatins the
+              PDId and lst surb time data after scrubbing.
+        Return: 0 or 1
+        """
+        before_scrub_log = dict(
+            zip(before_scrub_data["pgid"], before_scrub_data["last_scrub_stamp"])
+        )
+        after_scrub_log = dict(
+            zip(after_scrub_data["pgid"], after_scrub_data["last_scrub_stamp"])
+        )
+        number_of_pgs = len(before_scrub_data.keys())
+        count_pg = 0
+        for key in before_scrub_log.keys():
+            if key in after_scrub_log:
+                if before_scrub_log[key] != after_scrub_log[key]:
+                    count_pg = count_pg + 1
+                else:
+                    count_pg = count_pg - 1
+        if count_pg < round(number_of_pgs / 3):
+            return 1
+        return 0
--- a/conf/pacific/rados/tier_3_rados.yaml
+++ b/conf/pacific/rados/tier_3_rados.yaml
@@ -1,0 +1,49 @@
+# System Under Test environment configuration for Rados Tier-3 test suite.
+globals:
+  - ceph-cluster:
+      name: ceph
+
+      node1:
+        disk-size: 20
+        no-of-volumes: 4
+        role:
+          - _admin
+          - installer
+          - mgr
+          - mon
+          - osd
+
+      node2:
+        disk-size: 20
+        no-of-volumes: 4
+        role:
+          - mgr
+          - mon
+          - osd
+
+      node3:
+        disk-size: 20
+        no-of-volumes: 4
+        role:
+          - mon
+          - osd
+
+      node4:
+        disk-size: 20
+        no-of-volumes: 4
+        role:
+          - osd
+
+      node5:
+        disk-size: 20
+        no-of-volumes: 4
+        role:
+          - osd
+
+      node6:
+        role:
+          - rgw
+
+      node7:
+        role:
+          - client
--- a/suites/pacific/rados/tier_3_rados.yaml
+++ b/suites/pacific/rados/tier_3_rados.yaml
@@ -1,0 +1,49 @@
+tests:
+
+  # Cluster deployment stage
+
+  - test:
+      abort-on-fail: true
+      desc: Install software pre-requisites for cluster deployment.
+      module: install_prereq.py
+      name: setup pre-requisites
+
+  - test:
+      abort-on-fail: true
+      config:
+        verify_cluster_health: true
+        steps:
+          - config:
+              command: bootstrap
+              service: cephadm
+              args:
+                mon-ip: node1
+          - config:
+              command: add_hosts
+              service: host
+              args:
+                attach_ip_address: true
+                labels: apply-all-labels
+          - config:
+              command: apply
+              service: osd
+              args:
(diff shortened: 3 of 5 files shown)


SrinivasaBharath:
Test execution logs - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-64KRAZ

conf/pacific/rados/tier_3_rados.yaml
```diff
@@ -0,0 +1,49 @@
+# System Under Test environment configuration for RGW Tier-0 test suite.
```
psathyan, line 1: Could we fix the doc string

suites/pacific/rados/tier_3_rados.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - test:
+      abort-on-fail: true
+      config:
+        verify_cluster_health: true
+        steps:
+          - config:
+              command: bootstrap
+              service: cephadm
+              args:
+                registry-url: registry.redhat.io
```
psathyan, line 20: Let's avoid this key/value... it breaks execution in IBM-C

tests/rados/scheduled_scrub_scenarios.py
```diff
@@ -0,0 +1,82 @@
...
+from ceph.rados.rados_scrub import RadosScrubber
+
+log = logging.getLogger(__name__)
+
+
+def run(ceph_cluster, **kw):
+
+    OSD_SCRUB_MIN_INTERVAL = 1800
+    OSD_SCRUB_MAX_INTERVAL = 3600
+    OSD_DEEP_SCRUB_INTERVAL = 3600
```
psathyan, line 30: As they are inside the method, you can opt to use lower cases... The practice is to use upper cases only when it is global. This would not break anything though.

tests/rados/scheduled_scrub_scenarios.py
```diff
@@ -0,0 +1,82 @@
...
+    # Storing the pg dump log before setting the scrub parameters
+    before_scrub_log = rados_obj.get_pg_dump("pgid", "last_scrub_stamp")
+
+    # Get the cluster time
+    current_time = rados_obj.get_cluster_date()
+
+    # Preparation of cofiguration parameter values from the current
+    # cluster time
+    yy, mm, dd, hh, day = current_time.split(":")
+    date = datetime.datetime(int(yy), int(mm), int(dd), int(hh))
```
psathyan, line 46: I believe `date = datetime.datetime.now()` would help

tests/rados/scheduled_scrub_scenarios.py
```diff
@@ -0,0 +1,82 @@
...
+    rados_obj.set_osd_configuration("osd_scrub_begin_week_day", scrub_begin_weekday)
+    rados_obj.set_osd_configuration("osd_scrub_end_week_day", scrub_end_weekday)
+    rados_obj.set_osd_configuration("osd_scrub_begin_hour", scrub_begin_hour)
+    rados_obj.set_osd_configuration("osd_scrub_end_hour", scrub_end_hour)
+
+    # Scheduled scrub verification
+    endTime = datetime.datetime.now() + datetime.timedelta(minutes=60)
+    while True:
+        if datetime.datetime.now() >= endTime:
+            break
```
psathyan, line 72: you could do 
```
while datetime.datetime.now() > end_time
```
this would avoid the extra `if` statement

tests/rados/scheduled_scrub_scenarios.py
```diff
@@ -0,0 +1,82 @@
...
+        else:
+            after_scrub_log = rados_obj.get_pg_dump("pgid", "last_scrub_stamp")
+            scrub_status = rados_obj.verify_scrub(before_scrub_log, after_scrub_log)
+
+            if scrub_status == 0:
+                return 0
+            log.info(f'{"Scrubbing validation is in progress..."}')
+            time.sleep(240)
```
psathyan, line 80: A sleep of 4 minutes is huge... I understand why we would like to use it. Do we have any other option here?

SrinivasaBharath: The time required to test the scenario is 1 hour. After a couple of runs noticed that once it took minimum 29 minutes.
I don't think a sleep of  4 minutes is huge. Currently, I don't have any other options to reduce the verification time. The same is discussed in the rh-ceph group and came to this conclusion.

tests/rados/scheduled_scrub_scenarios.py
SrinivasaBharath: We required the current time of the container. The "datetime.datetime.now()" provide us the driver machine time. I notice that the timezones are different from the driver machine, installer node to the container. At Line-41, I am getting the container time.
More details at- https://bugzilla.redhat.com/show_bug.cgi?id=2023169

suites/pacific/rados/tier_3_rados.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+  # Cluster deployment stage
+
+  - test:
+      abort-on-fail: true
+      desc: Install software pre-requisites for cluster deployment.
+      module: install_prereq.py
+      name: setup pre-requisites
+
+  - test:
```
pdhiran, line 11: I feel it's better to have all the various components like rgw, fs, rbd deployed on the cluster for executing rados tests.

tests/rados/scheduled_scrub_scenarios.py
```diff
@@ -0,0 +1,82 @@
...
+    log.info(f'{"Scrubbing failed"}')
+    return 1
```
pdhiran, line 82: Can we remove all the config changes made for the purposes of testing and revert configs to original after test completion?

tests/rados/scheduled_scrub_scenarios.py
SrinivasaBharath: As per the discussion I tried with the strptime, In this by default hh:mm:ss are getting added with 0 values. If that is the case again required additional lines in the file.
Code sample:
import os
from datetime import datetime

out=os.popen("date +%Y:%m:%d").read()
out=out.strip()
date_obj=datetime.strptime(out,"%Y:%m:%d")
print("Date object=",date_obj)

Output: Date object= 2021-11-18 00:00:00

tests/rados/scheduled_scrub_scenarios.py
SrinivasaBharath: yes, It is very valid. I will implement that in a separate method.

conf/pacific/rados/tier_3_rados.yaml
SrinivasaBharath: Fixed the issue

suites/pacific/rados/tier_3_rados.yaml
SrinivasaBharath: Fixed the issue.

tests/rados/scheduled_scrub_scenarios.py
SrinivasaBharath: Fixed the issue.

tests/rados/scheduled_scrub_scenarios.py
SrinivasaBharath: Fixed the issue.

SrinivasaBharath:
Please find the cephci log at- http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-R86UTR/

suites/pacific/rados/tier_3_rados.yaml
SrinivasaBharath: It is under discussion that do we need to maintain a global file or a separate file for each tier. According to that, I will implement.

tests/rados/scheduled_scrub_scenarios.py
psathyan: Okay.
During our discussion, I believed that we agreed to modify `get_cluster_date` to retrieve all the values instead of specifics. As below we are using `datetime.timedelta` which does not impact our logic below.

Since, `get_cluster_date` comes from a module, I believe we should retrieve the all the fields and allow calling methods to process/filter their requirement.

The current implementation is okay...

ceph/rados/rados_scrub.py
```diff
@@ -0,0 +1,159 @@
...
+class RadosScrubber:
+    def __init__(self, node: CephAdmin):
+        """
+        initializes the env to run rados commands
+        Args:
+            node: CephAdmin object
+        """
+        self.node = node
+        self.ceph_cluster = node.cluster
+        self.rados_obj = RadosOrchestrator(node=node)
```
psathyan, line 41: IMO, we should inherit `RadosOrchestrator` and the purpose of this class would be scrubbing. This way `RadosScrubber` has a singular purpose and access on all methods of `RadosOrchestrator`.

ceph/rados/rados_scrub.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def run_container_cmd(self, cmd: str):
```
psathyan, line 43: IMO, this method would need to be part of `RadosOrchestrator`

ceph/rados/rados_scrub.py
```diff
@@ -0,0 +1,159 @@
...
+            cmd: Command that needs to be run on container
+
+        Returns : string  value
+        """
+
+        cmd = f"{cmd}"
+        out, err = self.node.shell([cmd])
+        return out
+
+    def get_cluster_date(self):
```
psathyan, line 57: This is a cluster specific operation... meaning that you would need the time being set in the cluster. The cluster deployment can be baremetal or container.

The current implementation would limit you to container specific deployment. 

If that this the intention, we may want to document it.

psathyan (commented):
Few additional thoughts... I understand there are gaps in the high level... 

This would involve rework once high level layer design keep changing. At present, we can get this in if you are ready to rework on these layers again.

ceph/rados/rados_scrub.py
pdhiran: I believe that all/most of the commands executed in 5.x, by default, uses method present with ` CephAdmin `, which by default executes command inside the ` cephadm ` Container itself. 

Even ` RadosOrchestrator ` class makes use of ` CephAdmin ` object for command execution. 

We can make the container calls directly, not sure if another method would be necessary to do the same.

ceph/rados/rados_scrub.py
SrinivasaBharath: Inheriting the RadosScrubber from RadosOrchestrator makes the class tightly coupled. Any changes in the RadosOrchestrator class have an impact on the RadosScrubber which is the child class. That s the region I created a separate class. It will be easy to plug and unplug whenever we required.

tests/rados/scheduled_scrub_scenarios.py
SrinivasaBharath: Added the method

ceph/rados/rados_scrub.py
SrinivasaBharath: Inherited from RadosOrchestrator

SrinivasaBharath:
Log file: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-AUJV8T/

ceph/rados/rados_scrub.py
SrinivasaBharath: Removed the init method

ceph/rados/rados_scrub.py
SrinivasaBharath: deleted the method called shell command

ceph/rados/rados_scrub.py
SrinivasaBharath: Moved the get_cluster_date  method to the core_workflows.py

SrinivasaBharath:
Test execution log- http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-ZY3NZV/

ceph/rados/scheduled_scrub_scenarios.py
```diff
@@ -0,0 +1,116 @@
+"""
```
psathyan, line 1: This is a test module. We already have one in `tests` directory. Do we need this?

SrinivasaBharath: No it is not required. The file had mistakenly been added to the commit. I removed the file

pdhiran (approved):
Looks Good.

PR #931, opened 2021-11-18
Rados tier-2 test suite changes

pdhiran:
1. Removed LRC EC pool test cases, as LRC profile not supported downstream
2. Reduced the number of PG's on each pool.
3. Sometimes, stretch fails to deploy as there might be some stray pools,
 other than default rule. Added steps to find and delete them.

Also fixes the issues with long running tests found in report portal for the run : " suites/pacific/rados/tier_2_rados.yaml (rhel-8.4.0-x86_64-released)#18 "

The failures were as cluster was in bad state( OSD's moved out of default root crush map ), after Stretch mode deployment failed. added flag ` abort-on-fail: true ` after stretch mode tests so that execution ends and further tests won't be attempted.

### DIFF
--- a/suites/pacific/rados/tier_2_rados.yaml
+++ b/suites/pacific/rados/tier_2_rados.yaml
@@ -167,7 +167,7 @@ tests:
         replicated_pool:
           create: true
           pool_name: test_re_pool
-          pg_num: 64
+          pg_num: 16
           size: 2
           disable_pg_autoscale: true
           rados_write_duration: 50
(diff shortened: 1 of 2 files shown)


pdhiran:
http://magna002.ceph.redhat.com/ceph-qe-logs/pawan/cephci-run-VZYKMU/ -> Pass logs for 5.0

Unable to deploy cluster and test with 5.1 builds due to bug : https://bugzilla.redhat.com/show_bug.cgi?id=2024176

tintumathew10 (approved):
LGTM

psathyan (approved):

PR #954, opened 2021-11-28
Refactoring CephBaremetalNode to support IBM-C classic environment

psathyan:

# Description

In this PR, the `CephBaremetalNode` object is refactored to support environments stood up in IBM-C. Also, changing some of the rules in the CLI arguments in case of `--cloud baremetal`.

__Logs__
https://159.23.92.24/job/psathyan_custom/2/console

### DIFF
--- a/run.py
+++ b/run.py
@@ -52,10 +52,11 @@ doc = """
 A simple test suite wrapper that executes tests based on yaml test configuration
 
  Usage:
-  run.py --rhbuild BUILD --inventory FILE
+  run.py --rhbuild BUILD
         (--suite <FILE>)...
         (--global-conf FILE | --cluster-conf FILE)
         [--cloud <openstack> | <ibmc> | <baremetal>]
+        [--inventory FILE]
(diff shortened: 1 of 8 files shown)


AmarnatReddy (approved):
LGTM

run.py
```diff
@@ -346,13 +349,23 @@ def run(args):
...
-    inventory_file = args["--inventory"]
-    osp_cred_file = args["--osp-cred"]
     suite_files = args["--suite"]
+
+    glb_file = args.get("--global-conf")
+    if args.get("--cluster-conf"):
+        glb_file = args["--cluster-conf"]
+
+    # Deciders
+    reuse = args.get("--reuse", None)
```
sunilkumarn417, line 359: by default can be `--store`.

psathyan: It is something we can ponder about. `--store` and `--reuse` are optional hence setting to default `--store` can be change the behavior unless we want to.

sunilkumarn417 (approved):

PR #984, opened 2021-12-10
Adding support for overrides

psathyan:

# Description

Adding support for overrides which enables better suite optimization

__Logs__
http://pastebin.test.redhat.com/1014725
http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/overrides/


init_suite.py
```diff
@@ -61,45 +134,123 @@ def suites(self):
...
+               name: setup pre-requisites
+
+        - Adding new keys
+
+            tests:
+              - test:
+                abort-on-fail: true
+                desc: Install software pre-requisites
+               module: install_prereq.py
+               name: setup pre-requisites
```
rakeshgm, line 181: I think this has to be aligned. follows for other examples too.

rakeshgm:
I was wondering why the `fragments` dir was renamed to `workunits`. Any reason why this was renamed.?  I remember we decided on the name `fragments` was best

psathyan:
> I was wondering why the `fragments` dir was renamed to `workunits`. Any reason why this was renamed.? I remember we decided on the name `fragments` was best

Workunits is an additional folder that contains a collection of fragments. These can be reused as FG's may have a common set of steps to be executed before the actual steps. With this we can avoid copying the soft links in multiple locations.

rakeshgm (approved):

PR #999, opened 2021-12-15
rados/pool_tests: Verifiy Compression on EC pools and ratio set

pdhiran:
1. Verification of effects of compression on pool size
2. Verification of data size on pools with different "compression_required_ratio"

# Description

<details>

- [x] Create a test case in Polarion reviewed and approved.
- [x] Create a design/automation approach doc. Optional for tests with similar tests already automated.
- [x] Implement the test script and perform test runs
- [x] Update Polarin Test with Automation script details and update automation fields
</details>

### DIFF
--- a/ceph/rados/core_workflows.py
+++ b/ceph/rados/core_workflows.py
@@ -943,6 +943,40 @@ class RadosOrchestrator:
         log.info(f"The image details are : {out}")
         return True
 
+    def check_compression_size(self, pool_name: str, **kwargs) -> bool:
+        """
+        Checks the given pool size against "compression_required_ratio" and verifies that data is
+        compressed in accordance to the ratio provided
+        Args:
+            pool_name: Name of the pool
+            **kwargs: additional params needed.
+                Allowed values:
+                    compression_required_ratio: ratio set on the pool for compression
+        Returns: True -> pass, False -> fail
+        """
+        log.info(f"Collecting stats about pool : {pool_name}")
+        pool_stats = self.run_ceph_command(cmd="ceph df detail")["pools"]
+        flag = False
+        for detail in pool_stats:
+            if detail["name"] == pool_name:
+                pool_1_stats = detail["stats"]
+                stored_data = pool_1_stats["stored_data"]
+                ratio_set = kwargs["compression_required_ratio"]
+                if pool_1_stats["data_bytes_used"] >= (stored_data * ratio_set):
+                    log.error(
+                        f"The data stored on pool is not compressed in accordance with the ratio set."
+                        f"Ideal size after compression <= {stored_data * ratio_set} \n"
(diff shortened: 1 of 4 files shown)


pdhiran:
Pass logs : http://magna002.ceph.redhat.com/ceph-qe-logs/pawan/cephci-run-K70RJ6/

ceph/rados/core_workflows.py
```diff
@@ -943,6 +943,34 @@ def verify_ec_overwrites(self, **kwargs) -> bool:
...
+        Args:
+            pool_name: Name of the pool
+            **kwargs: additional params needed.
+                Allowed values:
+                    compression_required_ratio: ratio set on the pool for compression
+        Returns: True -> pass, False -> fail
+        """
+        log.info(f"Collecting stats about pool : {pool_name}")
+        pool_stats = self.run_ceph_command(cmd="ceph df detail")["pools"]
+        pool_1_stats = [detail for detail in pool_stats if detail["name"] == pool_name][
```
psathyan, line 959: IMO, we should break as soon as we get a match. For example
```
for detail in pool_stats:
  if detail["name"] == pool_name:
    pool_1_stats = detail["stats"
    <you could include the below checks>
    break <return True -> if the below checks are included>
else:
  raise ResourceException("no matching detail found.")
```

tests/rados/pool_tests.py
```diff
@@ -134,12 +134,24 @@ def run(ceph_cluster, **kw):
...
         pool_2 = pool_config["pool-2"]
-        if not rados_obj.create_pool(pool_name=pool_1, **pool_config):
-            log.error("could not create pool-1")
-            return 1
-        if not rados_obj.create_pool(pool_name=pool_2, **pool_config):
-            log.error("could not create pool-2")
-            return 1
+
+        if config["Compression_tests"]["pool_type"] == "replicated":
+            if not rados_obj.create_pool(pool_name=pool_1, **pool_config):
```
psathyan, line 139: Would it make sense to have a generic `create_pool` method and have it create `ec-pool` based on the argument provided to it else a regular osd pool.

pdhiran: Hi Satya, the ` create_pool ` method is already generic, and creates the pool, replicated or erasure, based on the args passed. What ` create_erasure_pool ` method does is, it creates a new EC profile for the pool to be created, verifies that the profile is created successfully, and then calls the same ` create_pool `  method, supplying it with the profile created.

psathyan: Thanks for details. It makes sense to have different methods.

ceph/rados/core_workflows.py
pdhiran: made the changes to break off from the loop once the pool is found.

psathyan (approved):

tintumathew10 (approved):

PR #1009, opened 2021-12-17
rbd_utils.py: introducing sleep before pool creation check

pdhiran:

### DIFF
--- a/tests/rbd/rbd_utils.py
+++ b/tests/rbd/rbd_utils.py
@@ -1,3 +1,4 @@
+import datetime
 import json
 import logging
 import random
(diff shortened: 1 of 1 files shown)


tests/rbd/rbd_utils.py
```diff
@@ -104,6 +104,8 @@ def check_pool_exists(self, pool_name: str) -> bool:
         Returns:  True -> pass, False -> fail
 
         """
+        # sleeping for 5 seconds for the status of pool to be updated in ceph df command
+        sleep(5)
```
psathyan, line 108: Can we loop to check if the pool status is updated instead of a blind sleep?

pdhiran: made changes to check status of pool every 2 seconds. Total wait time for pool creation : 200 seconds.

psathyan (approved):

PR #1012, opened 2021-12-17
Throttling tier-x job by using batch processing

psathyan:

# Description

With limited resources and large number of test suites in a stage, we need to limit the number of parallel suites being executed in a pipeline. Allowing other pipelines to have sufficient resource for execution.

__Logs__
https://159.23.92.24/job/tier-x/1/console


sunilkumarn417 (approved):
LGTM.

collate maps is real good.

PR #1025, opened 2021-12-21
Added collocated rgw test cases

sunilkumarn417:
- Added collocated rgw test cases.
- updated code changes in apply module.

**Test Results:**
**5.0** -  http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-TAIN2I
**5.1** - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-YKT0I5

### DIFF
--- a/ceph/ceph_admin/apply.py
+++ b/ceph/ceph_admin/apply.py
@@ -71,35 +71,23 @@ class ApplyMixin:
 
         args = config.get("args")
 
-        node_names = None
         verify_service = False
         placement = args.pop("placement", {})
 
         if placement:
             placement_str = "--placement="
+            sep = placement.get("sep", " ")
+            node_names = list()
             verify_service = True
 
             if "label" in placement:
-                label = placement["label"]
-                node_names = [
-                    node["hostname"] for node in self.get_hosts_by_label(label)
-                ]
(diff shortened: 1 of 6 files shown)


ceph/ceph_admin/apply.py
```diff
@@ -71,63 +71,43 @@ def apply(self: ServiceProtocol, config: Dict) -> None:
...
-                    placement_str += '"*"'
-                    node_names = list()
-                    for node in self.cluster.node_list:
-                        if (
-                            len(node.role.role_list) == 1
-                            and ["client"] == node.role.role_list
-                        ):
-                            continue
-                        node_names.append(node.shortname)
+                    node_names.append("*")
```
psathyan, line 90: Shouldn't we exclude the `client`

sunilkumarn417: Actually its not required, as `"*"` should consider the nodes which is already part of the cluster. 
Earlier we used to find the nodes and replacing the nodes against `"*"` which was used same for validation also. but currently validation is different, so it would not be required.

psathyan: Okay

psathyan (approved):

PR #1028, opened 2021-12-22
Removing resharding TCs as it is not supported

psathyan:

# Description

Removing sharding related test cases from MS


udaysk23 (approved):
lgtm

PR #1087, opened 2022-01-17
[OKR][BZ-2014304]:  Support --limit and --tags option RHCS4

Ameenasuhani:
Signed-off-by: ameenasuhani <ameenasuhani@gmail.com>

Description: Support --limit and --tags option for BZ-2014304
Logs: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-R3ZOF8/

### DIFF
--- a/suites/nautilus/upgrades/tier-2_upgrade_with-tags.yaml
+++ b/suites/nautilus/upgrades/tier-2_upgrade_with-tags.yaml
@@ -1,0 +1,449 @@
+#############################################################################
+# OKR
+# (Note: Automation support is only for RPM Based cluster)
+#
+# Cluster configuration: (conf/nautilus/upgrades/tier-2_upgrade.yaml)
+# use --skip-version-compare args to avoid comparison of ceph versions as
+# partial upgrade results in version mismatch
+# --------------------------------------------------------------------------
+# 5 Nodes:
+#   - 3 MONs, 1 MGRs, 3 OSDs, 1 RGWs
+#
+# Test steps:
+# ---------------------------------------------------------------------
+# - Deploy Nautilus Ceph cluster using CDN RPMS.
+# - Run rolling update with `tags mons,mgrs,osds,rgws,post_upgrade`.
+# - check ceph status, versions, mon and osd dumps for debug info
+#
+#############################################################################
+tests:
+- test:
+    name: install ceph pre-requisites
+    module: install_prereq.py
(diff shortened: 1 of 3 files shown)


tests/ceph_ansible/test_ansible_upgrade.py
```diff
@@ -146,6 +148,15 @@ def run(ceph_cluster, **kw):
     if config.get("ansi_cli_args"):
         cmd += config_dict_to_string(config["ansi_cli_args"])
 
+    if limit_node:
+        short_name_list = [
```
sunilkumarn417, line 152: I would suggest to use below definitions to get node(s). so we could re-use common code.
https://github.com/red-hat-storage/cephci/blob/master/ceph/utils.py#L978

suites/nautilus/upgrades/tier-2_upgrade_with-tags.yaml
```diff
@@ -0,0 +1,597 @@
...
+      sudo: True
+    desc: Check for ceph mon dump debug info
+
+- test:
+    name: check-osd-dump
+    module: exec.py
+    config:
+      cmd: ceph osd dump
+      sudo: True
+    desc: Check for ceph osd dump debug info
```
sunilkumarn417, line 527: please have a look if we could club these commands and have it in one test case.

tests/ceph_ansible/test_ansible_upgrade.py
sunilkumarn417: these methods would be helpful even if we have multiple node names.

sunilkumarn417 (commented):
Hi @Ameenasuhani ,

Thanks for working on this improvements.

IMO, this would fail at version comparison if we use containerized deployment.
please check if skip-version-check would skip this too.

Ameenasuhani:
@sunilkumarn417 as far as I know this is a customer scenario and will not be available to others as discussed earlier,
so are `containerized deployments` into consideration?

sunilkumarn417:
> @sunilkumarn417 as far as I know this is a customer scenario and will not be available to others as discussed earlier, so are `containerized deployments` into consideration?

It would be good to support both.

sunilkumarn417 (commented):
Hi @Ameenasuhani ,

can you please create respective pipeline shell script for this test suite under 4x.

sunilkumarn417 (approved):
LGTM

Issue #1157, opened 2022-02-01
skip_version_compare doesn't work at suite level 


Ameenasuhani:
https://github.com/red-hat-storage/cephci/blob/620e002959c821d0e14bad4c5688ea6d8b62c285/tests/ceph_ansible/test_ansible_upgrade.py#L23 skip_version_compare  doesn't work because at suite level because it is only considered at args level

https://github.com/red-hat-storage/cephci/blob/master/run.py#L472

PR #1197, opened 2022-02-10
Fixes issue #1157 skip_version_compare doesn't work at suite level

Ameenasuhani:
Description

Fixes issue #1157: skip_version_compare doesn't work at suite level

Logs: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-BBTD3B/

### DIFF
--- a/run.py
+++ b/run.py
@@ -787,6 +787,9 @@ def run(args):
             if skip_subscription is True:
                 config["skip_subscription"] = True
 
+            if config.get("skip_version_compare"):
+                skip_version_compare = config.get("skip_version_compare")
+
             if args.get("--add-repo"):
                 repo = args.get("--add-repo")
                 if repo.startswith("http"):
--- a/suites/nautilus/upgrades/tier-2_upgrade_with-tags.yaml
+++ b/suites/nautilus/upgrades/tier-2_upgrade_with-tags.yaml
@@ -20,8 +20,6 @@ tests:
   - test:
       name: install ceph pre-requisites
       module: install_prereq.py
-      config:
-        is_production: true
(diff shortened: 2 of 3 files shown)


run.py
```diff
@@ -787,6 +787,9 @@ def fetch_test_details(var) -> dict:
             if skip_subscription is True:
                 config["skip_subscription"] = True
 
+            if config.get("skip_version_compare"):
```
sunilkumarn417, line 790: can you please re-consider having cli option for version comparison Since, 
- It is already handled and decided at test case level.
```
module: test_ansible_upgrade.py
      config:
        build: '4.x'
        skip_version_compare: true
```
- It's only used in 4x upgrade test module.

suites/nautilus/upgrades/tier-2_upgrade_with-tags.yaml
```diff
@@ -176,47 +192,53 @@ tests:
...
-    desc: Rolling update 4.x with tags mons
-    abort-on-fail: True
+      desc: Rolling update 4.x with tags mons
+      abort-on-fail: True
 
-- test:
-    desc: "generate sosreport"
-    module: sosreport.py
-    name: "generate sosreport"
+  - test:
```
sunilkumarn417, line 201: can you please re-consider about the need of sos report generation after every upgrade test case?

Ameenasuhani: this is a customer scenario where sos report is required at each step to get ceph mon osd dumps etc

sunilkumarn417 (approved):

PR #1209, opened 2022-02-14
Fix download location of Kafka binary

psathyan:

# Description

With the change in infra systems, the download location is corrected in this PR to avoid the below error

```
ERROR: error: sudo: /usr/local/kafka/bin/kafka-console-consumer.sh: command not found
```


mkasturi18 (approved):
looks good

PR #1224, opened 2022-02-21
Avoid failure of run if testname is a number

VasishtaShastry:
This fix avoids failure of run if test name is just a number.
This fix gives free-hand to formulate temporary suite files with test names as per users's convinience

modified:   utility/utils.py

Unhandled traceback without this commit -
  File "run.py", line 1045, in <module>
    rc = run(args)
  File "run.py", line 758, in run
    unique_test_name = create_unique_test_name(tc["name"], test_names)
  File "/home/vasishta/cephci/utility/utils.py", line 567, in create_unique_test_name
    base = "_".join(test_name.split())
AttributeError: 'int' object has no attribute 'split'

Signed-off-by: Vasishta <vashastr@redhat.com>

# Description

<details>

</details>

### DIFF
--- a/utility/utils.py
+++ b/utility/utils.py
@@ -564,7 +564,7 @@ def create_unique_test_name(test_name, name_list):
     Returns:
         unique name for the test case
     """
-    base = "_".join(test_name.split())
+    base = "_".join(str(test_name).split())
     num = 0
     while "{base}_{num}".format(base=base, num=num) in name_list:
         num += 1


psathyan:
@VasishtaShastry thanks for addressing this corner case.

However, do you see a need to use only numbers for our test cases?

VasishtaShastry:
> However, do you see a need to use only numbers for our test cases?
@psathyan 
No, I don't see any need but intention was not to constrain users to remember the rule that test name must contain a character especially when jotting down a temporary suite to test their module otherwise run would fail before testing their module
I'm okay if we want to keep things as it is

psathyan:
Should we check if the naming is as per the prescribed approach? No,  it is guideline that we have in place for readability.

The reason for my question was to check if we are missing any corner cases that would require an exception. Also, when we move to fragments... the directory naming guideline would need to be put in place.

psathyan (approved):

PR #1245, opened 2022-02-26
Adding support for RHEL-9 platform

psathyan:

# Description
The necessary changes to support RHEL-9 platform is carried out in this PR.

- The gathering of ceph-common and ceph-ansible versions is refactored so that we avoid any `NoneType` errors encountered when the packages are not found in the known paths.
- A change is also done as an enabler for `ibmc-baremetal`.
- sos report installation is modified due to warnings messages being displayed.

__cli__
```
/home/psathyan/venv/cephci.venv/bin/python run.py \
  --log-level debug \
  --instances-name prag01 \
  --osp-cred /home/psathyan/.osp-core.yaml \
  --rhbuild 5.1 \
  --platform rhel-9 \
  --build latest \
  --global-conf conf/pacific/rgw/tier-0_rgw.yaml \
  --inventory conf/inventory/rhel-9-latest.yaml \
  --suite suites/pacific/rgw/tier-0_rgw.yaml \
  --reuse rerun/prag01-5ON35E \
  --skip-subscription \
  --add-repo http://download.eng.bos.redhat.com/rhel-9/nightly/RHEL-9/latest-RHEL-9.0/repofile.repo \
  --skip-sos-report
```

__Logs__
http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/rhel-9/
RHEL-7: http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/test_r/

### DIFF
--- a/run.py
+++ b/run.py
@@ -211,9 +211,11 @@ def create_nodes(
             ceph_vmnodes = create_ibmc_ceph_nodes(
                 cluster, inventory, osp_cred, run_id, instances_name
             )
-
-        elif cloud_type == "baremetal":
+        elif "baremetal" in cloud_type:
             ceph_vmnodes = create_baremetal_ceph_nodes(cluster)
+        else:
+            log.error(f"Unknown cloud type: {cloud_type}")
+            raise AssertionError("Unsupported test environment.")
 
         ceph_nodes = []
         root_password = None
@@ -223,7 +225,7 @@ def create_nodes(
 
             if cloud_type == "openstack":
                 private_ip = node.get_private_ip()
-            elif cloud_type == "baremetal":
+            elif "baremetal" in cloud_type:
                 private_key_path = node.private_key if node.private_key else ""
                 private_ip = node.ip_address
                 look_for_key = True if node.private_key else False
@@ -342,6 +344,17 @@ def get_tier_level(files: str) -> str:
     return "Unknown"
 
 
+def get_html_page(url: str) -> str:
+    """Returns the content of the provided link."""
+    resp = requests.get(url, verify=False)
+
+    try:
+        return resp.text
+    except BaseException as be:  # noqa
+        log.debug(f"Failed to retrieve contents at {url}")
+        return "Unknown"
+
+
 def run(args):
 
     import urllib3
@@ -393,7 +406,7 @@ def run(args):
         root.addHandler(handler)
 
         server = f"tcp://{host}:{port}"
-        log._logger.debug(f"Log events are also pushed to {server}")
+        log.debug(f"Log events are also pushed to {server}")
 
     log_dna_config = get_cephci_config().get("log-dna", {})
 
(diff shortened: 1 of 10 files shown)


run.py
```diff
@@ -342,6 +344,17 @@ def get_tier_level(files: str) -> str:
     return "Unknown"
 
 
+def get_html_page(url: str) -> str:
```
sunilkumarn417, line 347: Would suggest to move this under any helper module.

run.py
```diff
@@ -342,6 +344,17 @@ def get_tier_level(files: str) -> str:
...
+    """Returns the content of the provided link."""
+    resp = requests.get(url, verify=False)
+
+    try:
+        return resp.text
+    except BaseException as be:  # noqa
+        log.debug(f"Failed to retrieve contents at {url}")
+        return "Unknown"
```
sunilkumarn417, line 355: Can we set `build type - upstream | rc | releases` rather than `unknown` which will ease debugging about build information?

run.py
```diff
@@ -540,41 +553,31 @@ def run(args):
 
         # get COMPOSE ID and ceph version
         if build not in ["released", "cvp", "upstream", None]:
-            if cloud_type == "openstack" or cloud_type == "baremetal":
-                resp = requests.get(base_url + "/COMPOSE_ID", verify=False)
-                compose_id = resp.text
-            elif cloud_type == "ibmc":
-                compose_id = "UNKNOWN"
+            compose_id = get_html_page(url=f"{base_url}/COMPOSE_ID")
```
sunilkumarn417, line 556: w.r.t above comment, even this can be moved under helper module.

tests/misc_env/sosreport.py
```diff
@@ -10,21 +10,25 @@
 
 def run(**kw):
     """
+    Entry point for execution used by cephci framework.
```
sunilkumarn417, line 13: I would suggest to update doc strings as per google doc style.

run.py
psathyan: This method is not generic and is specific to this module. I would like to keep it here...

run.py
psathyan: In case of `upstream | released`, this method would not be called as we don't check for versions when `build` has the those values.

tests/misc_env/sosreport.py
psathyan: Done.

run.py
sunilkumarn417: > 

Exactly, this method won't be called and ceph versions would be empty without being known as released or upstream.

run.py
sunilkumarn417: ok

run.py
sunilkumarn417: so would it make sense to add else part to set these ceph and ceph-ansible versions.?

psathyan: As discussed, we will pick the changes via a new PR specific to `build-type`

sunilkumarn417 (approved):

PR #1289, opened 2022-03-16
Fix container image while switching from non-containerized to contain…

rahullepakshi:
Fix container image while switching from non-containerized to containerized ceph-daemons
Signed-off-by: rlepaksh <rlepaksh@redhat.com>


rahullepakshi:
logs -  http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-8EBQYW/

Issue #88, opened 2020-04-09
purge cluster fails in most of suites


subhashp7i:
* tests/purge_cluster.py fails with ssh errors

XDsubhash:
need to do rca , Intermittent.

Ameenasuhani:
@XDsubhash RCA.The purge fails as recommendations have changed for 4.x [1].
we need to directly run "ansible-playbook infrastructure-playbooks/purge-cluster.yml" and the copying step needs to be skipped[2].

[1]https://access.redhat.com/documentation/en-us/red_hat_ceph_storage/4/html/installation_guide/purging-storage-clusters-deployed-by-ansible-install
[2]https://github.com/red-hat-storage/cephci/blob/master/tests/purge_cluster.py#L58

XDsubhash:
thanks @Ameenasuhani for pointing out 4.x changes, will include them

I believe this intermittent issue was seen even in 3.x. whenever someone see purge cluster fails they can update logs here

psathyan:
__Update__
Will need to follow with Sunil to gather information on the failure cases. Attempt to close on this issue if we have 4.x priority test suites failing.

psathyan:
With the new approach of test suites, we are no longer using `purge_cluster` in our runs. Closing this issue.

Issue #80, opened 2020-03-24
Fix issues in VM cleanup and floating IPs availability


XDsubhash:
* cleanup script might not be deleting the volumes attached to the instances
* smartly fetching Floating IPs available in a vm network before triggering the job

XDsubhash:
* currently: we can use the openstack cli to check available IPs:
  * openstack ip availability list | grep -e _cci_ -e shared
* each suite trigger might need on average around 15-20 IPs

XDsubhash:
vol cleanup error: https://ceph-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/rhceph-ansible-sanity-4.1-rhel7.8/6/console
https://ceph-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/rhceph-ansible-sanity-4.1-rhel8.1/15/console

XDsubhash:
#93 should fix the cleanup issue

XDsubhash:
#93 increase timeout to 25 seconds after detaching volume (volumes are in available state)  https://ceph-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/rhceph-ansible-sanity-4.1-rhel7.8/37/console

XDsubhash:
if we are still hitting volume cleanup issues, will remove parallel execution.

psathyan:
The refactoring effort carried out as part of #379 takes care of detaching the floating IP address before the VM is removed. Currently, floating IP address are not assigned. Closing this issue.

PR #1315, opened 2022-03-24
Adding logic for optimising pipeline scripts and ability to implement overrides

manasagowri:
# Description

<details>

</details>

### DIFF
--- a/pipeline/scripts/ci/getPipelineStages.py
+++ b/pipeline/scripts/ci/getPipelineStages.py
@@ -1,0 +1,120 @@
+import json
+import logging
+import os
+import random
+import string
+import sys
+
+import yaml
+from docopt import docopt
+
+log = logging.getLogger(__name__)
+doc = """
+This script fetches all the tests to be run for a pipeline based on the RHCS version and overrides
+
+    Usage:
+        getPipelineStages.py --rhcephVersion <VER> --tags <tags>
+                  [--overrides <str>]
+
+        getPipelineStages.py (-h | --help)
+
+    Options:
+        -h --help          Shows the command usage
+        -v --rhcephVersion VER     The rhcephVersion for which test stages need to be fetched
+        -t --tags <str>    tags to be used for filtering test scripts
+        -o --overrides <str>       Overrides to be considered for execution
+"""
+
+
+def generate_random_string(length):
+    """Generate a random alphanumeric string of given length"""
+    return "".join(random.choices(string.ascii_uppercase + string.digits, k=length))
+
+
+def fetch_stages(args):
+    """
+    Fetch test stages to be executed based on the input provided
+    Args:
+        args:
+            Dictionary containing the tier_level, cloud_type and all other necessary override parameters
+            to be used to fetch test stages for execution
+
+            arguments = {
+                "rhcephVersion": <the rhbuild version for which the stages need to be fetched>,
+                "tags": comma separated string containing the tags based on which the stages should be filtered,
+                "overrides": <overrides in json format, all keys accepted by run.py are supported to be overridden
+                            if any key similar to --store which does not have a value it can be passed as "store": ""
+            }
+
+            Example:
+                args = {
+                    "rhcephVersion": 5.1,
+                    "tags": "tier-1,stage-1,ibmc",
+                    "overrides": {"cloud_type": "openstack", "platform": "rhel-8"}
+
+    Returns:
+        Yaml string containing CLI to be executed for each test suite at every stage
+        Examples:
+            "test-cephfs-core-features":
+                "execute_cli": ".venv/bin/python run.py ......",
+                 "cleanup_cli": "...."
+            .....
+    """
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+    metadata_dir = os.path.abspath(f"{current_dir}/../../metadata")
+    metadata_file = f"{metadata_dir}/{args['rhcephVersion']}.yaml"
+    data = yaml.safe_load(open(metadata_file, "r"))
+    tags = args["tags"].split(",")
+    overrides = json.loads(args.get("overrides", {}))
+    cloud_type = "openstack"
+
+    filtered_data = filter(lambda d: all(tag in d["metadata"] for tag in tags), data)
+
+    test_stages = dict()
+    for script in filtered_data:
+        script_name = script.pop("name")
+        del script["metadata"]
+        instances_name = f"ci-{generate_random_string(5)}"
+        cleanup_cli = ".venv/bin/python run.py --osp-cred $HOME/osp-cred-ci-2.yaml"
+        cleanup_cli += f" --cleanup {instances_name}"
+        cleanup_cli += " --log-level DEBUG"
+
+        execute_cli = ".venv/bin/python run.py --osp-cred $HOME/osp-cred-ci-2.yaml"
+        execute_cli += f" --instances-name {instances_name}"
+        execute_cli += " --log-level DEBUG"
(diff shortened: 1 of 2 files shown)


utility/getPipelineStages.py
```diff
@@ -0,0 +1,135 @@
...
+            "stage-1":
+                "test-cephfs-core-features":
+                    "execute_cli": "<workspace>/.venv/bin/python run.py ......",
+                     "cleanup_cli": "...."
+                .....
+            ......
+    """
+    rhceph_version = args.get("rhcephVersion")
+    metadata_file = args.get("metadata_file")
+    if not rhceph_version and not metadata_file:
```
vamahaja, line 76: Is this correct condition? It will log error even in case `metadata_file` is not passed and at line #80, default value is used for same.

utility/getPipelineStages.py
```diff
@@ -0,0 +1,135 @@
...
+        log.error(
+            "Cannot fetch test stages since rhcephVersion and metadata file are not specified"
+        )
+    if not metadata_file:
+        metadata_file = f"pipeline/metadata/{rhceph_version}.yaml"
+    data = load_file(metadata_file)
+    if args.get("tags"):
+        tags = args["tags"].split(",")
+    overrides = json.loads(args["overrides"])
```
vamahaja, line 85: This will raise an exception in case `overrides` not passed

utility/getPipelineStages.py
```diff
@@ -0,0 +1,135 @@
...
+    rhceph_version = args.get("rhcephVersion")
+    metadata_file = args.get("metadata_file")
+    if not rhceph_version and not metadata_file:
+        log.error(
+            "Cannot fetch test stages since rhcephVersion and metadata file are not specified"
+        )
+    if not metadata_file:
+        metadata_file = f"pipeline/metadata/{rhceph_version}.yaml"
+    data = load_file(metadata_file)
+    if args.get("tags"):
```
vamahaja, line 83: Exception will be raised at line #88 in case `tags` is not passed as variable won't get declared. Below code can avoid same -
```
tags = args.get("tags", "").split(",")
```

this will return `''` even in case `tags` not passed, but variable get declared all the time

utility/getPipelineStages.py
```diff
@@ -0,0 +1,135 @@
...
+def fetch_stages(args):
+    """
+    Fetch test stages to be executed based on the input provided
+    Args:
+        args:
+            Dictionary containing the tier_level, cloud_type and all other necessary override parameters
+            to be used to fetch test stages for execution
+
+            Examples:
+                arguments = {
```
vamahaja, line 53: This is not an example, its description. Example would be - 
```
fetch_stages(rhcephVersion="14.2.22-105", tags="ibm,tier0", overrides={}, metadata_file="metadata.yaml")
```

utility/getPipelineStages.py
```diff
@@ -0,0 +1,135 @@
...
+    Options:
+        -h --help          Shows the command usage
+        -v --rhcephVersion VER     The rhcephVersion for which test stages need to be fetched
+        -t --tags <str>    tags to be used for filtering test scripts
+        -o --overrides <str>       Overrides to be considered for execution
+        -m --metadata FILE         Metadata file full path if it is not present in default pipeline/metadata folder
+"""
+
+
+def load_file(file_name):
```
vamahaja, line 31: why separate function for loading file. This can be simply written in more pythonic way -
```
content = yaml.safe_load(open(file_path, "r"))
```

manasagowri: Done.

utility/getPipelineStages.py
manasagowri: Done.

utility/getPipelineStages.py
manasagowri: Done.

utility/getPipelineStages.py
manasagowri: Done.

utility/getPipelineStages.py
manasagowri: Changed the condition.

utility/getPipelineStages.py
```diff
@@ -0,0 +1,128 @@
...
+
+    Returns:
+        Yaml string containing CLI to be executed for each test suite at every stage
+        Examples:
+            "test-cephfs-core-features":
+                "execute_cli": ".venv/bin/python run.py ......",
+                 "cleanup_cli": "...."
+            .....
+    """
+    rhceph_version = args.get("rhcephVersion")
```
vamahaja, line 67: As both the parameters are not mandatory and `metadata_file` is depends on `rhceph_version` parameter, below code could be better -
```
    metadata_file = args.get("metadata_file")
    if not metadata_file and args.get("rhcephVersion"):
    	metadata_file = f"pipeline/metadata/{rhceph_version}.yaml
    else
        rase Exception(
        	"Mandatory parameter 'rhcephVersion' and "
        	"'metadata' file are not specified")
```

vamahaja: Few mistakes in the code -
1. If condition could be - `if not metadata_file and "rhcephVersion" in args.keys()`
2. `rhceph_version` is parameter which is missing and can be fixed as -  `f"pipeline/metadata/{}.yaml".format(args.get("rhcephVersion"))`

manasagowri: But the above condition will throw an exception for the scenario where metadata file is present and rhcephVersion is not present, which is not desired. It should actually just use the given metadata file in that scenario. 

So I guess the below code snippet would do the job.

metadata_file = args.get("metadata_file")
rhceph_version = args.get("rhcephVersion")
if not metadata_file and rhceph_version:
    metadata_file = f"pipeline/metadata/{rhceph_version}.yaml
elif not metadata_file and not rhceph_version:
    raise Exception(
     "Cannot fetch test stages since rhcephVersion and metadata file are not specified")

manasagowri: Also as per the recent discussion I had with @psathyan we are removing the concept of passing metadata file and expecting that the metadata will definitely be in pipeline/metadata/<rhcephVersion>.yaml path.

utility/getPipelineStages.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Returns:
+        Yaml string containing CLI to be executed for each test suite at every stage
+        Examples:
+            "test-cephfs-core-features":
+                "execute_cli": ".venv/bin/python run.py ......",
+                 "cleanup_cli": "...."
+            .....
+    """
+    metadata_file = f"pipeline/metadata/{args['rhcephVersion']}.yaml"
```
psathyan, line 63: This could force us to execute the script from the root directory of the repo. If we could retrieve path of this file and then navigate to the metadata_file would allow us to execute from anywhere. That is

```
current_dir = os.path.dirname(os.path.abspath(__file__))
metadata_dir = os.path.abspath(f"{current_dir}/../pipeline/metadta")
```

psathyan (commented):
Looks good to me...

not so important suggestion.

Would it make sense to have this script under `pipeline/scripts/ci`?

utility/getPipelineStages.py
manasagowri: Done.

manasagowri:
> Looks good to me...
> 
> not so important suggestion.
> 
> Would it make sense to have this script under `pipeline/scripts/ci`?

Sure I have moved the script under pipeline/scripts/ci and made changes in v3.groovy as well accordingly.
Could you please review this again?

psathyan (approved):

PR #1358, opened 2022-04-06
adding test rbd

skompell:
Signed-off-by: Sri Sai Venkata Kasi Subbarayudu Kompella <skompell@skompell.remote.csb>

# Description

migrated suites/pacific/iscsi/tier-1_rbd_test-iscsi  to use fragments


fragments/install/cephadm/add-mds_using_placement.yaml
```diff
@@ -0,0 +1,14 @@
+tests:
```
psathyan, line 1: can we rename this fragment to `add_mds_using_nodes`?

fragments/install/cephadm/configure_client.yaml
```diff
@@ -0,0 +1,15 @@
+tests:
```
psathyan, line 1: We have `test_cephadm_add_client.yaml` fragment, can we override it instead of creating a new one?

fragments/install/cephadm/add-mds_using_placement.yaml
skompell: Yes pragdeesh
I will do that.

fragments/install/cephadm/configure_client.yaml
skompell: Sure pragdeesh...

PR #1373, opened 2022-04-13
RGW s3tests command to be executed with long_running enabled

psathyan:

# Description

In this PR, we move to use `long_running=True` so that we can view the log events immediately. As we are interested only with the exit status of the command... this is an improved addition.

__Logs__
https://159.23.92.24/job/custom_run_amk/54/console


psathyan:
It is not working as expected.

PR #1414, opened 2022-05-02
Add Waiter iterater to encapsulate waiting for timeouts

vamahaja:
Add waiter class to encapsulate steps execution in loop until timeout value elapses. It aims to avoid having to write boilerplate code comparing times.

### DIFF
--- a/ceph/waiter.py
+++ b/ceph/waiter.py
@@ -1,0 +1,32 @@
+"""Helper object to encapsulate waiting for timeouts"""
+
+import time
+
+
+class WaitUntil(object):
+    """A wait-retry loop as iterable.
+
+    This object abstracts away the wait logic allowing functions
+    to write the retry logic in a for-loop.
+    """
+
+    def __init__(self, timeout=60, interval=1):
+        self.timeout = timeout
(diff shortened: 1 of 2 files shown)


ceph/waiter.py
```diff
@@ -0,0 +1,31 @@
+"""Helper object to encapsulate waiting for timeouts"""
+
+import time
+
+
+class Waiter(object):
```
psathyan, line 6: Nice. This is a good.

Could we remain the class i.e. `WaitFor` or `WaitUntil`?

psathyan (commented):
Looks good to me...

I believe the module must be in `utilities` instead of `ceph`

ceph/waiter.py
vamahaja: Changed class name to `WaitUntil`

psathyan (commented):
Looks good to me... kindly provide unit test results or logs

vamahaja:
Test Results - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-I5EDSU/

psathyan (approved):

PR #1424, opened 2022-05-05
[pre-req] Drop iptables entries which affects ceph configuration

sunilkumarn417:

- Drop list of entries from IP tables which affects ceph configuration.
`REJECT     all  --  anywhere             anywhere             reject-with icmp-host-prohibited`

### DIFF
--- a/tests/misc_env/install_prereq.py
+++ b/tests/misc_env/install_prereq.py
@@ -188,6 +188,7 @@ def install_prereq(
         config_ntp(ceph, cloud_type)
 
     registry_login(ceph, distro_ver)
+    update_iptables(ceph)
 
 
 def setup_addition_repo(ceph, repo):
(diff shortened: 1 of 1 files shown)


pdhiran:
Looks Good. But this could be optional as we might want to set the iptables rule. Would it be good to have it in the CLI as a optional arg to be sent for flushing the IPtables?

sunilkumarn417:
> Looks Good. But this could be optional as we might want to set the iptables rule. Would it be good to have it in the CLI as a optional arg to be sent for flushing the IPtables?

To enable execution on baremetal with cluster-configuration file, this would be needed immediately. can we take other things as improvement(s) in next PR? Since we have to identify the fire-wall-d rules to be set for Ceph cluster.

psathyan (approved):

PR #1427, opened 2022-05-06
Fix 'ceph.open_firewall_port' library

vamahaja:
Fix `ceph.open_firewall_port` library. Changes consists of 

- Separate function to check `firewalld` package
- `int|str|list` type support for `open_firewall_port` function
- Update logic to call `open_firewall_port` in library
- Update `test_s3.py` TC to call `ceph.open_firewall_port`
- Remove `utils.open_firewall_port` library

### DIFF
--- a/ceph/ceph.py
+++ b/ceph/ceph.py
@@ -160,22 +160,32 @@ class Ceph(object):
         Open required ports on nodes based on relevant ceph demons types
         """
         for node in self.get_nodes():
+            ports = list()
             if node.role == "mon":
-                node.open_firewall_port(port="6789", protocol="tcp")
+                ports += ["6789"]
+
                 # for upgrades from 2.5 to 3.x, we convert mon to mgr
                 # so lets open ports from 6800 to 6820
-                node.open_firewall_port(port="6800-6820", protocol="tcp")
+                ports += ["6800-6820"]
+
             if node.role == "osd":
-                node.open_firewall_port(port="6800-7300", protocol="tcp")
+                ports += ["6800-7300"]
+
             if node.role == "mgr":
-                node.open_firewall_port(port="6800-6820", protocol="tcp")
+                ports += ["6800-6820"]
+
             if node.role == "mds":
-                node.open_firewall_port(port="6800", protocol="tcp")
+                ports += ["6800"]
+
             if node.role == "iscsi-gw":
-                node.open_firewall_port(port="3260", protocol="tcp")
-                node.open_firewall_port(port="5000-5001", protocol="tcp")
+                ports += ["3260", "5000-5001"]
+
             if node.role == "grafana":
-                node.open_firewall_port(port="6800-6820", protocol="tcp")
+                ports += ["6800-6820"]
+
+            if ports:
+                node.configure_firewall()
+                node.open_firewall_port(port=ports, protocol="tcp")
 
     def setup_ssh_keys(self):
         """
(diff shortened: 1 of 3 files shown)


ceph/ceph.py
```diff
@@ -1615,19 +1622,34 @@ def open_firewall_port(self, port, protocol):
...
+        Opens firewall port on the node
+
+        Args:
+            port(str|int|list): port(s) to be enabled
+            protocol(str): protocol
+        """
+        if self.pkg_type == "rpm":
+            cmd = "firewall-cmd --zone=public "
+
+            if type(port) in (str, int):
```
psathyan, line 1641: IMO, it is a good practice to use `isinstance` instead of `type`. For example

```
if isinstance(port, (str, int)):
```

Here, I agree that it has no impact knowing about the argument.

ceph/ceph.py
```diff
@@ -1598,13 +1611,7 @@ def remove_ceph_object(self, ceph_object):
...
             self.get_allocated_volumes()[0].status = NodeVolume.FREE
 
-    def open_firewall_port(self, port, protocol):
-        """
-        Opens firewall port on the node
-        Args:
-            port(str): port, can be range
-            protocol(str): protcol
-        """
+    def configure_firewall(self):
```
psathyan, line 1614: It is a good practice to have a docString... as we use `readthedocs`... it helps describe the method.

ceph/ceph.py
```diff
@@ -153,22 +153,35 @@ def setup_ceph_firewall(self):
...
+            elif node.role == "iscsi-gw":
+                ports = ["3260", "5000-5001"]
+
+            elif node.role == "grafana":
+                ports = "6800-6820"
+
+            else:
+                continue
+
+            node.open_firewall_port(port=ports, protocol="tcp")
```
psathyan, line 184: I believe this breaks the existing behavior. A node can have multiple roles i.e. Ceph services are co-located. In such a scenario `ports` holds only the last value of the satisfied conditional statement.

psathyan (changes requested):
Some of the changes in this PR breaks existing functionality. 

- Please attach runs of RHCS 4.x and RHCS 5.x

ceph/ceph.py
vamahaja: @psathyan that was my first assumption but if you look at the current scenario, it iterating only once over the node list and exactly matching the `node.role` to word string.
If node has multiple roles, `node.list` should be a list or string with multiple words separated by comma. In this case if condition need to be changed as - 
```
if "mon" in node.role:
```

Please suggest.

ceph/utils.py
```diff
@@ -738,17 +738,6 @@ def search_ethernet_interface(ceph_node, ceph_node_list):
     return ceph_node.search_ethernet_interface(ceph_node_list)
 
 
-def open_firewall_port(ceph_node, port, protocol):
```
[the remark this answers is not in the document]
vamahaja: @psathyan this function was used at only one place. It's wrapper function calling function caling same name function. I am not sure purpose of placing this function here, so I removed it. IMO it doesn't belog here and should not be kept in this `utils` library. Please suggest.

psathyan:
@vamahaja would you be able to attach the logs for this run?

vamahaja:
Below are the results for the tests 

5.x 
-----

- tier-1_upgrade_test-4x-to-5x-rpm.yaml - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-V9QCIN/
- tier-2_dmfg_test-elasticity-after-upgrade-from-4-cdn-to-5-latest.yaml - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-TOYNDS/

4.x
-----

- tier-2_rgw_s3tests.yaml - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-U55HPB/
- tier-1_rgw.yaml - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-PWZS1S/
- tier-0_deploy_rpm_ceph.yaml - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-9H47FL/
- tier-2_deploy_test-ceph-rpms - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-LGW235/

psathyan (approved):

PR #1432, opened 2022-05-08
[OCS] Configure client role on the installer node

psathyan:

# Description

OCS CI requires the client role to be configured on the first node so that they could execute/query the ceph cluster. In this PR, the following changes are being carried out

- Client role is added to `node1`
- `rbd` pool is created by default during the workflow

__Logs__
4.x: http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/odf/4x/
5.x: http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/odf/5x/


tintumathew10 (approved):
Looks good.

PR #1434, opened 2022-05-09
[Stretch] Adding support of network attribute for Ceph datastructure

psathyan:
# Description

This PR adds a new property to `Ceph` object for supporting RHCS deployments that have `cluster` network. The following changes are

```
globals
  -
     ceph_cluster:
       networks:
         public: ["CIDR notation"]
         cluster: ["CIDR notation"]
```

Two new methods have been added to gather the network information
- get_public_networks
- get_cluster_networks

Both the methods return a comma separated `str` consisting of the deployed network CIDR's.

The primary focus of the PR was to enable deploying Ceph in stretched mode. For this reason, the CephVMNode of compute type has a new attribute `location`. The purpose of this attribute is to hold the meta information about the physical location of the system. It could be a data-center or rack or an identifier.

A new method has been added to `Ceph` object to retrieve the list of nodes available in a `location`.

- get_nodes_in_location


HaruChebrolu (approved):
LGTM

PR #1483, opened 2022-05-27
email_gchat_notify

rkhambad:
Signed-off-by: Rajendra Khambadkar <rkhambad@rkhambad.remote.csb>

### DIFF
--- a/pipeline/post-results.groovy
+++ b/pipeline/post-results.groovy
@@ -6,6 +6,10 @@ def nodeName = "centos-7"
 def credsRpProc = [:]
 def sharedLib
 def rpPreprocDir
+def tierLevel
+def stageLevel
+def run_type
+def build_url
 def reportBucket = "qe-ci-reports"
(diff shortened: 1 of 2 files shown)


pipeline/post-results.groovy
```diff
@@ -80,9 +82,25 @@ node(nodeName) {
             }
 
             def testStatus = "ABORTED"
+            /* Publish results through E-mail and Google Chat */
+            def tag = msgMap["pipeline"]["tags"]
```
manasagowri, line 86: Can you add a check whether tags key is present before assigning tierLevel and stageLevel as well before this line?

manasagowri (changes requested):

manasagowri (approved):
Looks good to me

pipeline/post-results.groovy
rkhambad: added the check for tags

psathyan (approved):

PR #1499, opened 2022-05-31
Fix for Sanity and Schedule Stages

rkhambad:
Signed-off-by: Rajendra Khambadkar <rkhambad@rkhambad.remote.csb>


manasagowri (approved):
Looks good to me. @psathyan We have redefined the stages based on the latest execution time for sanity runs. And also added proper staging for periodical suites. Could you please review this?

Issue #1496, opened 2022-05-30
Bug : ceph/utils.py - get_public_network method needs changes to findout public_network with multiple subnets


VasishtaShastry:
**Describe the bug**
get_public_network module retrieves subnet based on one node and if instances gets added with modes in multiple subnet, this makes it impossible to configure cluster using ceph-ansible as public_network config in all.yml needs to be added with comma separated value with all subnets.
 
https://github.com/red-hat-storage/cephci/blob/master/ceph/utils.py#L922-L935

**To Reproduce**
Run needs to have instances from multiple networks.

http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-4S6JPR/ceph_ansible_0.log

**Expected behavior**
get_public_network needs to be extended to gather subnets from all different VMs if subnet is different.

**Screenshots**

**Environment**
A concise description of your environment: RHEL version,
python version, operating system, command used etc.
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-4S6JPR/ceph_ansible_0.log

**Additional context**

sunilkumarn417:
Need changes at,  https://github.com/red-hat-storage/cephci/blob/8be32512153cd9e81c1616e21ae429f3be5569e7/tests/ceph_installer/test_ansible.py#L33

psathyan:
As the fix has been merged, closing this issue as fixed.

PR #1503, opened 2022-05-31
fix-issue-1496: consider all node public subnets

sunilkumarn417:

**Fix-issue-1496:** Bug : ceph/utils.py - get_public_network method needs changes to find out public_network with multiple subnets #1496

**Results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-GYP4JY/ceph_ansible_0.log

### DIFF
--- a/ceph/utils.py
+++ b/ceph/utils.py
@@ -920,20 +920,20 @@ def get_root_permissions(node, path):
     node.obtain_root_permissions(path)
 
 
-def get_public_network(node):
-    """
-    Get the configured public network subnet for nodes in the cluster.
-    This function retrieves the public network subnet from the ceph node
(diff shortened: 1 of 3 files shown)


Ameenasuhani (approved):
LGTM.
Could you please add logs.

VasishtaShastry:
(small doubt)
ceph_nodes contains subnet of all nodes including clients adding which would be redundant right ?
This could be a corner case optimisation thought, please let me know your thoughts

sunilkumarn417:
> (small doubt) ceph_nodes contains subnet of all nodes including clients adding which would be redundant right ? This could be a corner case optimisation thought, please let me know your thoughts

shouldn't be issue now. But we can take this in next PR to skip client-only role node.

psathyan (approved):

PR #1507, opened 2022-06-01
Adding change to handle apply labels when client node is present

manasagowri:

# Description
When client was the only role present in the conf file, the host spec creation was getting abruptly returned with None as the output. Instead I have added a change to continue execution for other nodes if present.

Success Log - https://159.23.92.24/job/rhceph-test-execution-manual/8/console (The RGW IO errors might have been because my branch was a little behind the master while I triggered this job)
Failure Log - without the code update given in this PR - https://159.23.92.24/job/rhceph-test-execution-manual/9/console

<details>

</details>

### DIFF
--- a/ceph/ceph_admin/helper.py
+++ b/ceph/ceph_admin/helper.py
@@ -174,8 +174,8 @@ class GenerateServiceSpec:
         template = self._get_template("host")
         hosts = []
         address = spec.get("address")
-        labels = spec.get("labels")
         for node_name in spec["nodes"]:
+            labels = spec.get("labels")
             host = dict()
             node = get_node_by_id(self.cluster, node_name)
             host["hostname"] = self.get_hostname(node)
@@ -185,7 +185,7 @@ class GenerateServiceSpec:
             if labels:
                 # Skipping client node, if only client label is attached
                 if len(node.role.role_list) == 1 and ["client"] == node.role.role_list:
-                    return
+                    continue
 
                 if isinstance(labels, str) and labels == "apply-all-labels":
                     labels = self.get_labels(node)
--- a/conf/pacific/cephadm/tier-1_5node_cephadm_bootstrap.yaml
+++ b/conf/pacific/cephadm/tier-1_5node_cephadm_bootstrap.yaml
@@ -40,4 +40,7 @@ globals:
         disk-size: 15
       node4:
         role:
+          - rgw
+      node5:
+        role:
           - client
--- a/pipeline/metadata/5.0.yaml
+++ b/pipeline/metadata/5.0.yaml
@@ -68,7 +68,7 @@
     - stage-1
 - name: "test-bootstrap-custom-ssl-dashboard-port-and-apply-spec"
   suite: "suites/pacific/cephadm/tier-1_ssl_dashboard_port.yaml"
-  global-conf: "conf/pacific/cephadm/tier-1_3node_cephadm_bootstrap.yaml"
+  global-conf: "conf/pacific/cephadm/tier-1_5node_cephadm_bootstrap.yaml"
   platform: "rhel-8"
   rhbuild: "5.0"
   inventory:
@@ -83,7 +83,7 @@
     - stage-1
 - name: "test-bootstrap-skip-dashboard-and-custom-ceph-directory-organisation"
   suite: "suites/pacific/cephadm/tier-1_skip_dashboard.yaml"
-  global-conf: "conf/pacific/cephadm/tier-1_3node_cephadm_bootstrap.yaml"
+  global-conf: "conf/pacific/cephadm/tier-1_5node_cephadm_bootstrap.yaml"
   platform: "rhel-8"
   rhbuild: "5.0"
   inventory:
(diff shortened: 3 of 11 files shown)


conf/pacific/cephadm/tier-1_3node_cephadm_bootstrap.yaml
```diff
@@ -41,3 +41,4 @@ globals:
       node4:
         role:
           - client
+          - rgw
```
psathyan, line 44: Not sure if we need to add this role unless dedicated by the test case.

Also, the filename is kind of misleading... it states 3 node environment but actually it has a 4 node test environment.

ceph/ceph_admin/helper.py
```diff
@@ -185,7 +185,7 @@ def generate_host_spec(self, spec):
             if labels:
                 # Skipping client node, if only client label is attached
                 if len(node.role.role_list) == 1 and ["client"] == node.role.role_list:
-                    return
+                    continue
```
psathyan, line 188: Thanks. I think this makes sense.

sunilkumarn417, line 188: would there be template render exception, suppose if we have one spec content for host with only one node with client label?
```
          - config:
              command: apply_spec
              service: orch
              specs:
                - service_type: host
                  address: true
                  labels: apply-all-labels
                  nodes:
                    - node4
```

sunilkumarn417: can we take away client node if we do not want to use it in cluster and continue with test suite scenario.
```
          - config:
              command: apply_spec
              service: orch
              specs:
                - service_type: host
                  address: true
                  labels: apply-all-labels
                  nodes:
                    - node2
                - service_type: host
                  address: true
                  labels: apply-all-labels
                  nodes:
                    - node3
 ```

Comment (sunilkumarn417) [reply ceph/ceph_admin/helper.py:188]: Or you could another node if that make sense.###

File: conf/pacific/cephadm/tier-1_3node_cephadm_bootstrap.yaml (hunk 1 above)
Comment (sunilkumarn417) [inline conf/pacific/cephadm/tier-1_3node_cephadm_bootstrap.yaml:44]: hope there would be changes here, let's have separate client node.

File: ceph/ceph_admin/helper.py (hunk 2 above)
Comment (sunilkumarn417) [reply ceph/ceph_admin/helper.py:188]: and rgw deployment step uses node4 again. I think its better we add one more node to the conf.

Comment (manasagowri) [reply ceph/ceph_admin/helper.py:188]: Yes sunil RGW config was added to node4 so yea I think I will add another node, node5 as a client node. And for client nodes, if the role added in conf file is only client, according to the current code it skips adding that node to the cluster. So with this change I guess we should be good. 

tldr: We will be adding new node for client node5 and this will not be part of the cluster

File: conf/pacific/cephadm/tier-1_3node_cephadm_bootstrap.yaml (hunk 1 above)
Comment (manasagowri) [reply conf/pacific/cephadm/tier-1_3node_cephadm_bootstrap.yaml:43]: Changes done. Could you please review?

File: ceph/ceph_admin/helper.py (hunk 2 above)
Comment (manasagowri) [reply ceph/ceph_admin/helper.py:188]: Changes done. Could you please review?

File: conf/pacific/cephadm/tier-1_3node_cephadm_bootstrap.yaml (hunk 1 above)
Comment (manasagowri) [reply conf/pacific/cephadm/tier-1_3node_cephadm_bootstrap.yaml:43]: Changes done. Could you please review?

File: conf/pacific/cephadm/tier-1_5node_cephadm_bootstrap.yaml [hunk 3]
```diff
@@ -39,5 +39,10 @@ globals:
         no-of-volumes: 6
         disk-size: 15
       node4:
+        role:
+          - rgw
+        no-of-volumes: 6
+        disk-size: 15
```
psathyan, line 45: Any particular reason why `disks` are being added to the VM not having `osd` role

psathyan (commented):
Thanks. 

A minor change with respect to the volumes in the conf file.

ceph/ceph_admin/helper.py
manasagowri: @sunilkumarn417 I tested by adding the spec content like you mentioned and there was no template render exception. 

          - config:
              command: apply_spec
              service: orch
              specs:
                - service_type: host
                  address: true
                  labels: apply-all-labels
                  nodes:
                    - node5

https://159.23.92.24/job/rhceph-test-execution-manual/14/console
https://159.23.92.24/job/rhceph-test-execution-manual/15/console

So we can retain this spec or remove it, it wouldn't make a difference.

conf/pacific/cephadm/tier-1_5node_cephadm_bootstrap.yaml
manasagowri: No particular reason. I was just testing out, I have removed the disks

manasagowri:
@sunilkumarn417 Success Log - https://jenkins.ceph.redhat.com/view/RHCS%20QE/job/rhceph-test-executor/568/console
Can you please review and approve this PR?

psathyan (commented):
LGTM

manasagowri:
@psathyan I have raised a backlog in JIRA for the monitoring service verification issue - https://issues.redhat.com/browse/RHCEPHQE-3787

sunilkumarn417 (approved):
LGTM.

PR #1508, opened 2022-06-01
In upgrade suite abort on fail handled properly 

sunilangadi2:
Signed-off-by: Sunil Angadi <sangadi@redhat.com>

# Description
In this PR if the upgrade fails then it aborts,
Handled **abort on fail** for the whole test not only at individual sub-tests in the parallel run.
initially it was handled in sub tests hence even upgrade fails it goes with other tests eventually those tests doesn't make any sense to run
from the issue: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-BW3XBH/


manasagowri (approved):
LGTM

PR #1602, opened 2022-07-08
fix service validation retry issue

sunilkumarn417:

Issue: http://magna002.ceph.redhat.com/ceph-qe-logs/pawan/cephci-run-EPKXM7/cluster_deployment_0.log

- Fix service validation race-around condition issue.

### DIFF
--- a/ceph/ceph_admin/orch.py
+++ b/ceph/ceph_admin/orch.py
@@ -104,11 +104,17 @@ class Orch(
                 f"{running}/{count} {service_name if service_name else service_type} up... retrying"
             )
 
-            _retries = 3 if (_count != count and running != count) else _retries - 1
-            _count = count
+            if count + running < 1:
+                continue
 
-            if not (count + running) == 0 and count == running and _retries == 0:
-                return True
+            if count == running and _count == count:
+                if _retries < 1:
+                    return True
+                _retries -= 1
+
+            if _count != count:
+                _count = count
+                _retries = 3
 
         # Identify the failure
         out, err = self.ls(check_status_dict)


ceph/ceph_admin/orch.py
```diff
@@ -104,11 +104,17 @@ def check_service_exists(
                 f"{running}/{count} {service_name if service_name else service_type} up... retrying"
             )
 
-            _retries = 3 if (_count != count and running != count) else _retries - 1
-            _count = count
+            if count + running < 1:
+                continue
```
psathyan, line 108: We could leave this as `if count + running == 0: continue`

ceph/ceph_admin/orch.py
```diff
@@ -104,11 +104,17 @@ def check_service_exists(
...
 
-            if not (count + running) == 0 and count == running and _retries == 0:
-                return True
+            if count == running and _count == count:
```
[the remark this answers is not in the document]
psathyan: I believe this is correct... as the possible scenarios are
```
# 1 0/1 <= failure case
# 2  0/1
       1/1 <=success
# 3 0/0
      2/2  <= we still have more 
      6/8
      12/12 <= expected values
```

ceph/ceph_admin/orch.py
```diff
@@ -104,11 +104,17 @@ def check_service_exists(
...
+                if _retries == 0:
```
psathyan, line 111: Can we do `if _retries < 1`?

ceph/ceph_admin/orch.py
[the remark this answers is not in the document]
HaruChebrolu: okay!

ceph/ceph_admin/orch.py
sunilkumarn417: I think Its good to validate positive number.

psathyan: I don't think `count + running` is going to `<= 0`... it is not important... but I thought it would avoid a confusion.

psathyan (approved):

PR #1627, opened 2022-07-14
Added timeout for upstream listener pipeline

vivekanandan-k-rh:
# Description

Added timeout for upstream listener pipeline


psathyan (commented):
IMO, we can have one timeout covering both the stages... what do you think?

vivekanandan-k-rh:
Sounds good as complete execution won't take more than 30 minutes. Let update the code

vivekanandan-k-rh:
@psathyan , Updated the code to include 30 mins of timeout for both Preparing and Staging stages together

psathyan (commented):
looks good to me

sunilkumarn417 (commented):
I think there are chances of file state can be in lock state with this timeout, can we add any `finally` clause.
If it is already covered then it should be good. 

Can we please check this before we merge?

PR #1638, opened 2022-07-18
Introduced node id and location attrs

sunilkumarn417:

Framework and cephadm changes needed to enable MDR stretch mode Ceph cluster using LV paths.

Introduced node id and location attributes in cloud types.
```
        vm.location = params.get("location")
        vm.id = params.get("id")

(Pdb) ceph.vm_node.id
'node1'
(Pdb) ceph.vm_node.location
'DC1'
```


ceph/utils.py
```diff
@@ -187,7 +190,8 @@ def create_ibmc_ceph_nodes(
                 node_dict = ceph_cluster.get(node)
                 node_params = params.copy()
                 node_params["role"] = RolesContainer(node_dict.get("role"))
-
+                node_params["id"] = node_dict.get("id") or node
```
psathyan, line 193: IMO, we could do `node_params["id"] = node_dict.get("id", node)`... not that the above wouldn't work.

psathyan (approved):
Looks good to me

PR #1666, opened 2022-07-26
test_cidr_blocklisting.py: Added suite and conf files for CIDR Blocklisting

pdhiran:

# Description

<details>

</details>

### DIFF
--- a/conf/pacific/rados/6-node-cluster-6-clients.yaml
+++ b/conf/pacific/rados/6-node-cluster-6-clients.yaml
@@ -1,0 +1,94 @@
+# Test Suite to test CIDR blocklisting of clients
+# Deployment for all the ceph daemons , with 3 mon's, 3 mgr's, 15 OSD daemons & 6 Client nodes
+globals:
+  - ceph-cluster:
+      name: ceph
+      node1:
+        networks:
+          - provider_net_cci_12
+        role:
+          - _admin
+          - mon
+          - mgr
+          - installer
+          - node-exporter
+          - alertmanager
+          - grafana
+          - prometheus
+          - crash
+      node2:
+        networks:
+          - provider_net_cci_12
+        role:
+          - mon
+          - mgr
+          - mds
+          - rgw
+          - node-exporter
+          - alertmanager
+          - crash
+      node3:
+        networks:
+          - provider_net_cci_12
+        role:
+          - osd
+          - node-exporter
+          - crash
+        no-of-volumes: 5
+        disk-size: 25
+      node4:
+        networks:
+          - provider_net_cci_12
+        role:
+          - osd
+          - node-exporter
+          - crash
+        no-of-volumes: 5
+        disk-size: 25
+      node5:
+        role:
+          - osd
+          - node-exporter
+          - crash
+        no-of-volumes: 5
+        disk-size: 25
+      node6:
+        networks:
+          - provider_net_cci_12
+        role:
+          - mon
+          - mgr
+          - mds
+          - rgw
+          - node-exporter
+          - crash
+      node7:
+        networks:
+          - provider_net_cci_8
+        role:
+          - client
+      node8:
+        networks:
+          - provider_net_cci_8
+        role:
+          - client
+      node9:
+        networks:
+          - provider_net_cci_9
+        role:
+          - client
+      node10:
+        networks:
+          - provider_net_cci_9
+        role:
+          - client
+      node11:
+        networks:
+          - provider_net_cci_7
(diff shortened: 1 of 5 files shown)


pdhiran:
This is not the complete test case workflow. This is just conf, suite files, and beginning of workflow. Would be completing the workflows in multiple PRs.

pdhiran:
Pass log : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-FSBM3K

conf/pacific/rados/6-node-cluster-6-clients.yaml
```diff
@@ -0,0 +1,94 @@
...
+      node11:
+        networks:
+          - provider_net_cci_7
+        role:
+          - client
+      node12:
+        networks:
+          - provider_net_cci_7
+        role:
+          - client
```
SrinivasaBharath, line 94: As per the test case, the clients should be 9 in the configuration the clients are 6.

conf/pacific/rados/6-node-cluster-6-clients.yaml
```diff
@@ -0,0 +1,94 @@
...
+        networks:
+          - provider_net_cci_9
+        role:
+          - client
+      node11:
+        networks:
+          - provider_net_cci_7
+        role:
+          - client
+      node12:
```
SrinivasaBharath, line 90: There are total 12 nodes exist in the file but the file name is "6-node-cluster-6-clients.yaml"

tests/cephadm/test_client.py
```diff
@@ -43,35 +45,56 @@ def put_file(client, file_name, content, file_mode, sudo=True):
...
-
-    # Hold local copy of the client key-ring in the installer node
-    if config.get("store-keyring"):
-        put_file(cls.installer, client_file, cnt_key, "w")
+    if config.get("nodes"):
+        nodes = config["nodes"]
+        if not isinstance(nodes, list):
+            nodes = [nodes]
+
+        def setup(host):
```
SrinivasaBharath, line 53: The nested method is big. Could you please add a separate method

tests/cephadm/test_client.py
```diff
@@ -43,35 +45,56 @@ def put_file(client, file_name, content, file_mode, sudo=True):
...
+                # Copy the ceph.conf to client
+                put_file(node, "/etc/ceph/ceph.conf", ceph_conf, "w")
+
+            # Copy admin keyring to client node
+            if config.get("copy_admin_keyring"):
+                admin_keyring, _ = cls.shell(
+                    args=["ceph", "auth", "get", "client.admin"]
+                )
+                put_file(
+                    node, "/etc/ceph/ceph.client.admin.keyring", admin_keyring, "w"
```
SrinivasaBharath, line 74: This method copies the keyring file and conf file on all nodes. Is it necessary to copy the files on all nodes?

SrinivasaBharath (changes requested):

conf/pacific/rados/6-node-cluster-6-clients.yaml
pdhiran: Configuring 9 clients would increase the number of nodes. So I have updated  the client count to 6, with which we can sufficiently test the feature. The same count is updated in doc too.

Let me know if you feel we would need 9 clients. Would increase the count.

conf/pacific/rados/6-node-cluster-6-clients.yaml
pdhiran: This would be a special suite, were we have multiple clients, spread across subnets, wich would be used only for CIDR testing.

Naming it 12 node suite would give ppl assumption that we have that many nodes as part of cluster, whereas, here, only 6 are part of cluster, and 6 are clients.

Hence the 6-6 naming convention.  Let me know if you think we can have a better convention here.

tests/cephadm/test_client.py
pdhiran: This would copy the keyrings onto all the client nodes. Yes, it is absolutely necessary for keyrings to be copied/present on all client nodes.

tests/cephadm/test_client.py
pdhiran: Making it a separate method would mean we would have to pass too many params from the  `  put_file() ` method, making it complicated. The contents of that method was put into new nested method as we wanted to invoke parallel execution for client setup. 

Would add a new new method if absolutely necessary.

SrinivasaBharath (approved):
LGTM

PR #1725, opened 2022-08-15
Fix post results failures

psathyan:

# Description

This PR addresses some of the error encountered during post results

- Ensure group write is enabled for the results file
- Enhanced lock file removal even when a write failure is encountered


sunilkumarn417 (approved):
LGTM

PR #1750, opened 2022-08-21
Add support for deploying release version binaries unavailable in CDN

psathyan:

# Description
This PR adds support for deploying a base cluster version that is unavailable in the CDN repository. The reason could be a newer version was released. 

This support addresses the gap wherein a customer skip versions during upgrades.

Scope.
This support is limited to RH environment only.

Example
For 5.x
```
args:
  rhcs-version: 5.1   # RH Ceph Storage version
  release: ga            # Or any of the supported build values - ga, z1, z2, z1-async1, z1-async2
```

For 4.x
```
config:
  use_internal:
    rhcs-version: 4.2
    release: z2
  ansi_config:
    ceph_origin: distro
    ...
```

__Logs__
5: http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/internal/5/
4: http://magna002.ceph.redhat.com/ceph-qe-logs/psathyan/internal/4/

### DIFF
--- a/ceph/ceph_admin/apply.py
+++ b/ceph/ceph_admin/apply.py
@@ -126,6 +126,9 @@ class ApplyMixin:
             return
 
         if not check_service_exists(
-            self.installer, service_name=service_name, service_type=self.SERVICE_NAME
+            self.installer,
+            service_name=service_name,
+            service_type=self.SERVICE_NAME,
+            rhcs_version=self.cluster.rhcs_version,
         ):
             raise OrchApplyServiceFailure(self.SERVICE_NAME)
--- a/ceph/ceph_admin/bootstrap.py
+++ b/ceph/ceph_admin/bootstrap.py
@@ -6,7 +6,7 @@ from typing import Dict
 from ceph.ceph import ResourceNotFoundError
 from ceph.ceph_admin.cephadm_ansible import CephadmAnsible
 from utility.log import Log
-from utility.utils import get_cephci_config
+from utility.utils import fetch_build_artifacts, get_cephci_config
 
 from .common import config_dict_to_string
 from .helper import GenerateServiceSpec, create_ceph_config_file, validate_spec_services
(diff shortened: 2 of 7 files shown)


ceph/ceph_admin/bootstrap.py
```diff
@@ -183,6 +191,25 @@ def bootstrap(self: CephAdmProtocol, config: Dict):
...
         rhbuild = self.config.get("rhbuild")
 
+        # Support installation of the baseline cluster whose version is not available in
+        # CDN. This is primarily used for an upgrade scenario. This support is currently
+        # available only for RH network.
+        _rhcs_version = args.pop("rhcs-version")
+        _rhcs_release = args.pop("release")
+        if _rhcs_release and _rhcs_version:
+            _platform = "-".join(rhbuild.split("-")[1:])
+            _details = fetch_build_artifacts(_rhcs_release, _rhcs_version, _platform)
```
sunilkumarn417, line 201: would it be good, if we can make it as common under a definition in any module, so it could be called in module where build info needed.

ceph/ceph_admin/apply.py
```diff
@@ -126,6 +126,9 @@ def apply(self: ServiceProtocol, config: Dict) -> None:
             return
 
         if not check_service_exists(
-            self.installer, service_name=service_name, service_type=self.SERVICE_NAME
+            self.installer,
```
sunilkumarn417, line 129: I think similar changes needed at the `ceph_admin/helper.py:905`.

ceph/ceph_admin/bootstrap.py
psathyan: Thanks... yes, it would be helpful.

Can we pick this in another PR as the scope of this requirement was for baseline version?

ceph/ceph_admin/apply.py
psathyan: done.

sunilkumarn417 (approved):
LGTM.

PR #1783, opened 2022-09-01
Logrotate functionality

manasagowri:

Success Logs : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-FMJJ42
# Description

<details>

</details>

### DIFF
--- a/ceph/ceph_admin/helper.py
+++ b/ceph/ceph_admin/helper.py
@@ -10,6 +10,7 @@ from os.path import dirname
 from time import sleep
 from typing import Optional
 
+from dateutil import parser
 from jinja2 import Template
 
 from ceph.ceph import CommandFailed
@@ -763,6 +764,37 @@ def file_or_path_exists(node, file_or_path):
     return False
 
 
+def file_or_path_updated(node, file_or_path, interval=120):
+    """
+    Method to check if the given file or directory was updated in the given interval
+    Args:
+        node: node object where file should be exists
+        file_or_path: ceph file or directory path
+        interval: if mentioned checks whether the file existing in the given path
+                              was updated anytime between now and the interval given here.
+
+    Returns:
+        boolean
+    """
+    try:
+        out, _ = node.exec_command(cmd=f"date -u -r {file_or_path}", sudo=True)
+        LOG.info(f"Output : {out}")
+    except CommandFailed as err:
+        LOG.error(f"Error: {err}")
+        return False
+
+    modified_time = parser.parse(out).replace(tzinfo=None)
+    time_now = datetime.utcnow().replace(tzinfo=None)
+    time_diff = (time_now - modified_time).seconds
+    LOG.info(f"Time difference between now and gz file modified date : {time_diff}")
+    if time_diff > interval:
+        LOG.error(
+            f"The gz file was not created within the previous {time_diff} seconds"
+        )
+        return False
+    return True
+
+
 def monitoring_file_existence(node, file_or_path, file_exist=True, timeout=180):
     """Method to monitor a file existence.
 
@@ -827,6 +859,49 @@ def validate_log_file_after_enable(cls):
         except CommandFailed as err:
             LOG.error("Error: %s" % err)
             return False
+
+    return True
+
+
+def validate_log_rotate(cls):
+    """
+    Verify logrotation of cephadm.log file
+    Args:
+        cls: cephadm instance object
+
+    Returns:
+        boolean
+    """
+    # step 1: check for the files in current log directory
+    if not file_or_path_exists(cls.installer, "/var/log/ceph/cephadm.log"):
+        LOG.error("Log rotation failed: Logs not present in /var/log/ceph folder")
+        return False
+
+    # step 2: apply logrotate on cephadm.log
+    out = cls.installer.exec_command(
+        cmd="logrotate --force /etc/logrotate.d/cephadm", sudo=True
+    )
+    # if logrotate doesn't result in logs being zipped return False
+    if not file_or_path_updated(
+        cls.installer, "/var/log/ceph/cephadm.log.1.gz", 120
+    ) and file_or_path_exists(cls.installer, "/var/log/ceph/cephadm.log"):
+        LOG.error("Log rotation failed while enabling rotate")
+        return False
+
+    # step 3: perform any operation that would result in cephadm log be written
+    out, _ = cls.shell(args=["ceph", "orch", "ps", "--format", "json"])
+    daemons = json.loads(out)
+    daemon = [d for d in daemons if "mon" in d["daemon_name"]][0]
+    out, _ = cls.shell(
+        args=["ceph", "orch", "daemon", "redeploy", daemon["daemon_name"]]
+    )
+
+    # step 4: check new cephadm.log file is created
+    if not file_or_path_exists(cls.installer, "/var/log/ceph/cephadm.log"):
+        LOG.error(
+            "Log rotation failed: Logs not present in /var/log/ceph folder after logrotate"
+        )
+        return False
 
     return True
 
--- a/ceph/ceph_admin/restart.py
+++ b/ceph/ceph_admin/restart.py
@@ -28,7 +28,7 @@ class RestartMixin:
         """
         restart_init_time = datetime.datetime.utcnow()
         rc = self.op("restart", config)
-        args = config.get("args")
+        args = config.get("args", {})
         verify = args.pop("verify", True)
         if verify:
             rc = self.check_service_restart(
--- a/suites/pacific/cephadm/tier-1_ssl_dashboard_port.yaml
+++ b/suites/pacific/cephadm/tier-1_ssl_dashboard_port.yaml
@@ -162,6 +162,7 @@ tests:
       config:
         verify_cluster_health: true
         verify_log_files: true
+        verify_log_rotate: true
         steps:
           - config:
               command: shell      # wait for daemons to come up, need to remove this sleep
--- a/tests/ceph_installer/test_cephadm.py
+++ b/tests/ceph_installer/test_cephadm.py
@@ -15,7 +15,11 @@ from ceph.ceph_admin.common import fetch_method
 from ceph.ceph_admin.crash import Crash
 from ceph.ceph_admin.daemon import Daemon
 from ceph.ceph_admin.grafana import Grafana
-from ceph.ceph_admin.helper import get_cluster_state, validate_log_file_after_enable
+from ceph.ceph_admin.helper import (
+    get_cluster_state,
+    validate_log_file_after_enable,
+    validate_log_rotate,
+)
 from ceph.ceph_admin.host import Host
 from ceph.ceph_admin.iscsi import ISCSI
 from ceph.ceph_admin.mds import MDS
@@ -144,6 +148,9 @@ def run(ceph_cluster: Ceph, **kwargs) -> int:
             if not isvalid:
                 LOG.error("Log file validation failure")
                 return 1
+        if config.get("verify_log_rotate") and not validate_log_rotate(cephadm):
+            LOG.error("Log rotate validation failure")
+            return 1
 
     except BaseException as be:  # noqa
         LOG.error(be, exc_info=True)


ceph/ceph_admin/helper.py
```diff
@@ -831,6 +832,57 @@ def validate_log_file_after_enable(cls):
...
+    """
+    Verify logrotation of cephadm.log file
+    Args:
+        cls: cephadm instance object
+
+    Returns:
+        boolean
+    """
+    # step 1: check for the files in current log directory
+    out = cls.installer.exec_command(cmd="ls /var/log/ceph", sudo=True)
```
pranavprakash20, line 845: One suggestion, can we directly check whether "/var/log/ceph/cephadm.log" is present using '-f' or similar rather than parsing the ls output. Please ignore, if I got it wrong

ceph/ceph_admin/helper.py
```diff
@@ -831,6 +832,57 @@ def validate_log_file_after_enable(cls):
...
+    daemon = [d for d in daemons if "node-exporter" in d["daemon_name"]][0]
+    out, _ = cls.shell(
+        args=["ceph", "orch", "daemon", "redeploy", daemon["daemon_name"]]
+    )
+
+    # step 4: check new cephadm.log file is created
+    out = cls.installer.exec_command(cmd="ls /var/log/ceph", sudo=True)
+    # if new logs are not created return False
+    files = re.split("\n| ", out[0])
+    logFileExists = [f for f in files if f == "cephadm.log"]
```
pranavprakash20, line 876: Suggestion : Since this check is been used multiple places, can we make it as a method and optimize the code.

ceph/ceph_admin/helper.py
vamahaja: +1, in this case `exec_command` will raise exception and validation for files is not needed. Also, I would suggest to use `out, _` to collect `exec_command` return values, as it returns tuple of stdout and err values

ceph/ceph_admin/helper.py
```diff
@@ -831,6 +832,57 @@ def validate_log_file_after_enable(cls):
...
+    if not logFileExists:
+        LOG.error("Log rotation failed: Logs not present in /var/log/ceph folder")
+        return False
+
+    # step 2: apply logrotate on cephadm.log
+    out = cls.installer.exec_command(
+        cmd="logrotate --force /etc/logrotate.d/cephadm", sudo=True
+    )
+    # if logrotate doesn't result in logs being zipped return False
+    out = cls.installer.exec_command(cmd="ls /var/log/ceph", sudo=True)
```
vamahaja, line 857: I would suggest to add to check time for file as well. It should be after log rotate command execution.

ceph/ceph_admin/helper.py
```diff
@@ -831,6 +832,57 @@ def validate_log_file_after_enable(cls):
...
+    # step 4: check new cephadm.log file is created
+    out = cls.installer.exec_command(cmd="ls /var/log/ceph", sudo=True)
+    # if new logs are not created return False
+    files = re.split("\n| ", out[0])
+    logFileExists = [f for f in files if f == "cephadm.log"]
+    if not logFileExists:
+        LOG.error(
+            "Log rotation failed: Logs not present in /var/log/ceph folder after logrotate"
+        )
+        return False
```
vamahaja, line 881: In this case, add custom exception and raise same exception, it will be easy for debugging.

ceph/ceph_admin/restart.py
```diff
@@ -28,8 +28,8 @@ def restart(self: Union[OrchProtocol, DaemonProtocol], config: Dict):
         """
         restart_init_time = datetime.datetime.utcnow()
         rc = self.op("restart", config)
-        args = config.get("args")
-        verify = args.pop("verify", True)
+        args = config.get("args", None)
```
vamahaja, line 31: Not required, by default it's `None`

tests/ceph_installer/test_cephadm.py
```diff
@@ -144,6 +148,11 @@ def run(ceph_cluster: Ceph, **kwargs) -> int:
             if not isvalid:
                 LOG.error("Log file validation failure")
                 return 1
+        if config.get("verify_log_rotate"):
+            isverified = validate_log_rotate(cephadm)
+            if not isverified:
+                LOG.error("Log rotate validation failure")
+                return 1
```
vamahaja, line 155: This can be rewritten as -
```
        if config.get("verify_log_rotate") and validate_log_rotate(cephadm):
                LOG.error("Log rotate validation failure")
                return 1
```

vamahaja (changes requested):

ceph/ceph_admin/helper.py
manasagowri: Done.

ceph/ceph_admin/helper.py
manasagowri: Done. I am reusing an already existing method.

ceph/ceph_admin/helper.py
manasagowri: This would not be required, since cephadm.log would be converted into a gz file when we execute logrotate command. And the log file wouldn't exist. After an operation which generates new logs a new cephadm.log file would be created. So I feel its okay to not have a time check on the file.

ceph/ceph_admin/restart.py
manasagowri: Done.

ceph/ceph_admin/helper.py
manasagowri: At the method calling end it is expecting a boolean value true or false for validation of log rotate functionality, so I would prefer logging error to log.error and returning false instead of raising exception. It would be cleaner I feel.

tests/ceph_installer/test_cephadm.py
manasagowri: Done.

ceph/ceph_admin/helper.py
vamahaja: Then gz file might have name with timestamp. Add step to validate gz file.

ceph/ceph_admin/helper.py
vamahaja: You can use below code to raise exception, this will add same error message but as an exception which will be more clear.
```
class LogRotationError(Exception):
    pass
.
.
.
            raise LogRotationError("Log rotation failed: Logs not present in /var/log/ceph folder after logrotate")
```

manasagowri: This is just one of the many validations we are already doing in this file. Now if i have to add exceptions to all of them it would change the scope of this PR. I would prefer taking this up as a different task. Please let me know your thoughts.

ceph/ceph_admin/helper.py
manasagowri: Done.

ceph/ceph_admin/helper.py
pranavprakash20: Since the calling script expects a bool val, I feel the current approach is good. Also in case if a custom exception has to be added, it would be better to have it in the test file (not mandatory, just a suggestion).

LGTM

pranavprakash20 (approved):
LGTM

suites/pacific/cephadm/tier-1_ssl_dashboard_port.yaml
```diff
@@ -62,6 +62,8 @@ tests:
       desc: installation of ceph pre-requisites
       module: install_prereq.py
       abort-on-fail: true
+      config:
+        is_production: true
```
sunilangadi2, line 66: @manasagowri why are we setting this config in this suite?

ceph/ceph_admin/helper.py
```diff
@@ -831,6 +860,49 @@ def validate_log_file_after_enable(cls):
...
+        LOG.error("Log rotation failed: Logs not present in /var/log/ceph folder")
+        return False
+
+    # step 2: apply logrotate on cephadm.log
+    out = cls.installer.exec_command(
+        cmd="logrotate --force /etc/logrotate.d/cephadm", sudo=True
+    )
+    # if logrotate doesn't result in logs being zipped return False
+    if not file_or_path_updated(
+        cls.installer, "/var/log/ceph/cephadm.log.1.gz", 10
```
sunilangadi2, line 883: are you sure the zipped file after log rotate it will be this `cephadm.log.1.gz` name?

manasagowri: Yes sunil, I have tested it and it will be. Everytime the logrotate command renames the old files and creates new file with name cephadm.log.1.gz itself. The old ones will be moved to cephadm.log.2.gz and etc.,

ceph/ceph_admin/helper.py
manasagowri: Thanks Pranav. @vamahaja If you think this is okay. could you please approve the changes?

suites/pacific/cephadm/tier-1_ssl_dashboard_port.yaml
manasagowri: That I had added during testing, since there was issue with stage subscription. Thanks Sunil for pointing it out. I have removed it now.

manasagowri:
> Signed-off-by: mgowri [mgowri@redhat.com](mailto:mgowri@redhat.com)
> 
> Success Logs : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-FMJJ42
> 
> # Description
> Please include Automation development guidelines. Source of Test case - New Feature/Regression Test/Close loop of customer BZs
> 
> click to expand checklist

Latest test logs after all changes - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-D3FJ1R/

ceph/ceph_admin/restart.py
```diff
@@ -29,10 +29,10 @@ def restart(self: Union[OrchProtocol, DaemonProtocol], config: Dict):
...
-            rc = self.check_service_restart(
-                service_name=args.get("service_name"),
-                restart_init_time=restart_init_time,
-            )
+        if args:
+            if args.get("verify", True):
+                rc = self.check_service_restart(
+                    service_name=args.get("service_name"),
+                    restart_init_time=restart_init_time,
+                )
```
vamahaja, line 37: This can be re-written as 
```
	if args and args.get("verify", True):
            rc = self.check_service_restart(
                service_name=args.get("service_name"),
                restart_init_time=restart_init_time,
            )
```

ceph/ceph_admin/helper.py
vamahaja: @pranavprakash20 @manasagowri Yes, script expects boot value but in this case it will simply log error and continue to next test case. 
But when this TC get called in between other TCs and get fails, it will be hard to check for error log. If we add an exception it will be highlighted in the logs and will get stacktrace as well which will be easy for debugging. 
Otherwise it works as expected.

ceph/ceph_admin/helper.py
vamahaja: What if this file is already present? In this case we are looking into incorrect file.

ceph/ceph_admin/helper.py
```diff
@@ -763,6 +764,34 @@ def file_or_path_exists(node, file_or_path):
...
+        LOG.info(f"Output : {out}")
+        modified_time = parser.parse(out).replace(tzinfo=None)
+        time_now = datetime.utcnow().replace(tzinfo=None)
+        time_diff = (time_now - modified_time).seconds
+        LOG.info(f"Time difference between now and gz file modified date : {time_diff}")
+        if time_diff > 120:
+            return False
+        return True
+    except CommandFailed as err:
+        LOG.error(f"Error: {err}")
```
vamahaja, line 790: I see below issues using this try..catch block here -
1. `CommandFailed` exception handled here is specific to only one statement i.e for `node.exec_command`. This makes no effect on adding other statement inside `try..catch` block.
2. In case of command failure we adding simple error log and returning `False` which is checked at calling time and again returning `False` with simple error message. In this case it will be harder to get stacktrace in case of TC failure.
3. [1] using this module will be more reliable than executing command.

[1] https://docs.paramiko.org/en/stable/api/sftp.html#paramiko.sftp_client.SFTPClient.lstat

suites/quincy/cephadm/tier-0.yaml
```diff
@@ -16,6 +16,8 @@ tests:
       desc: Install software pre-requisites for cluster deployment.
       module: install_prereq.py
       name: setup pre-requisites
+      config:
```
sunilangadi2, line 19: @manasagowri i think you forgot to remove that config from here as well, before adding file to git stage just cross check the files to be commited.

ceph/ceph_admin/helper.py
manasagowri: @vamahaja As discussed let us take this up as part of a different PR since the scope for this PR would be changed if we need to change all the validations.

ceph/ceph_admin/helper.py
manasagowri: As discussed the new file will always be cephadm.log.1.gz

ceph/ceph_admin/restart.py
manasagowri: Done.

ceph/ceph_admin/helper.py
manasagowri: I have added only the required statement inside the try catch block. The paramiko implementation, as discussed let us take it up as a different task since it is out of scope for the current PR.

suites/quincy/cephadm/tier-0.yaml
manasagowri: Yes sunil thanks again. I have updated it.

sunilangadi2 (dismissed):
Changes LGTM.

ceph/ceph_admin/helper.py
```diff
@@ -763,6 +764,34 @@ def file_or_path_exists(node, file_or_path):
...
+        return False
+
+    modified_time = parser.parse(out).replace(tzinfo=None)
+    time_now = datetime.utcnow().replace(tzinfo=None)
+    time_diff = (time_now - modified_time).seconds
+    LOG.info(f"Time difference between now and gz file modified date : {time_diff}")
+    if time_diff > interval:
+        return False
```
vamahaja, line 791: Better to add log.error here with expected and actual value.

vamahaja (commented):
LGTM.

ceph/ceph_admin/helper.py
manasagowri: Good catch. thanks Vaibhav. Added the error log.

vamahaja (dismissed):

vamahaja (dismissed):

ceph/ceph_admin/restart.py
```diff
@@ -29,8 +29,7 @@ def restart(self: Union[OrchProtocol, DaemonProtocol], config: Dict):
         restart_init_time = datetime.datetime.utcnow()
         rc = self.op("restart", config)
         args = config.get("args")
```
psathyan, line 31: not important... we could do `args = config.get("args", {})`.

This would avoid `if args and args.get("verify", True)`

`verify` being an internal usage... it was removed from the args dictionary. This allows to process `args` directly 

IMO, we should modify this line and leave the rest.

psathyan (approved):

PR #1835, opened 2022-09-14
Updated mon db trimming to psi only and added slow ops

tintumathew10:
Signed-off-by: tintumathew10 <tmathew@redhat.com>

Updated mon db trimming to psi only and added slow ops

### DIFF
--- a/pipeline/metadata/5.3.yaml
+++ b/pipeline/metadata/5.3.yaml
@@ -402,8 +402,8 @@
     - rados
     - stage-1
 
-- name: "RADOS regression testing for Mon DB trimming"
-  suite: "suites/pacific/rados/tier-3_rados_test-mon-db-trimming.yaml"
+- name: "Testing RADOS regression for slow op requests"
+  suite: "suites/pacific/rados/tier-2_rados_test-slow-op-requests.yaml"
   global-conf: "conf/pacific/rados/11-node-cluster.yaml"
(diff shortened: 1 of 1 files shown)


pipeline/metadata/5.3.yaml
```diff
@@ -402,14 +402,29 @@
...
+    - stage-2
+
 - name: "RADOS regression testing for Mon DB trimming"
   suite: "suites/pacific/rados/tier-3_rados_test-mon-db-trimming.yaml"
   global-conf: "conf/pacific/rados/11-node-cluster.yaml"
   platform: "rhel-8"
   rhbuild: "5.3"
   inventory:
     openstack: "conf/inventory/rhel-8-latest.yaml"
-    ibmc: "conf/inventory/ibm-vpc-rhel-8-latest.yaml"
```
psathyan, line 412: In additional to this removal, you would need to remove the tags `ibmc, openstack` and add the tag `openstack-only`

psathyan (commented):
Do we need to modify 6.0 file also?

pipeline/metadata/5.3.yaml
tintumathew10: done

tintumathew10:
> Do we need to modify 6.0 file also?

yes, I will ceate another PR for the same.

psathyan (approved):

PR #1849, opened 2022-09-21
Adding TC ID in cephci that is already covered in automation.

viduship:
Adding TC ID in cephci that is already covered in the automation.
Signed-off-by: viduship <vimishra@redhat.com>

### DIFF
--- a/suites/pacific/rgw/tier-1_rgw.yaml
+++ b/suites/pacific/rgw/tier-1_rgw.yaml
@@ -100,7 +100,7 @@ tests:
   - test:
       name: Dynamic Resharding tests
       desc: Resharding test - dynamic
-      polarion-id: CEPH-83571740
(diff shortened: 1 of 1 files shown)


anrao19 (approved):

ckulal (approved):

Ameenasuhani (commented):
@viduship rather than adding as comment can we not replace it any one of the suite?

eg:https://github.com/red-hat-storage/cephci/blob/master/suites/pacific/rgw/tier-1_rgw_test_upgrade-5-to-latest.yaml#L104-L112
https://github.com/red-hat-storage/cephci/blob/master/suites/nautilus/rgw/tier-1_rgw_ssl_test-upgrade-4-to-latest.yaml#L94-L102

TejasC88 (approved):

PR #1852, opened 2022-09-22
Updating Polarian ID in cephci suites

viduship:
TC "CEPH-83571740" provides automation coverage to the TCs ceph-11479 and ceph-11477.

Hence, mentioning these 2 TCs (11479 and -11477) in cephci, before updating their automation status in the Polarian.

Signed-off-by: viduship <vimishra@redhat.com>

### DIFF
--- a/suites/pacific/rgw/tier-1_rgw.yaml
+++ b/suites/pacific/rgw/tier-1_rgw.yaml
@@ -100,7 +100,7 @@ tests:
   - test:
       name: Dynamic Resharding tests
       desc: Resharding test - dynamic
-      polarion-id: CEPH-83571740 # also applies to ceph-11479
+      polarion-id: CEPH-83571740 # also applies to ceph-11479, ceph-11477
       module: sanity_rgw.py
       config:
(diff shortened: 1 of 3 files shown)


Ameenasuhani (commented):
Adding polarion id as comment would not help us when tfa is taken into consideration in analysis of failure. It would be great to replace the new polarion id with the old if there are multiple occurrence.

eg: This polarion id is available CEPH-83571740 is available in below suites also you could update anyone polarion id with ceph-11479 and other with ceph-11477
https://github.com/red-hat-storage/cephci/blob/38111307204b052161483649fca74206e9dc33b7/suites/pacific/rgw/tier-1_rgw_test_upgrade-5-to-latest.yaml#L136-L144
https://github.com/red-hat-storage/cephci/blob/38111307204b052161483649fca74206e9dc33b7/suites/nautilus/rgw/tier-1_rgw_ssl_test-upgrade-4-to-latest.yaml#L111-L119

TejasC88 (approved):

PR #1873, opened 2022-09-28
Fix parallel.py

ckulal:

previously we used consider module name for updating test result dictionary in parallel run, 
which was causing an issue when we have same test module name in the test-cases, then it used to replace an result each time and only result for last testcase used to present in result dictionary.

hence changing to test-case name instead of module name since we have unique test-case name

Pass log: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-6RRB17/Parallel_run_0.log

# Description

<details>

</details>

### DIFF
--- a/tests/parallel/test_parallel.py
+++ b/tests/parallel/test_parallel.py
@@ -73,6 +73,7 @@ def execute(test, args, results: dict):
     file_name = test.get("module")
     mod_file_name = os.path.splitext(file_name)[0]
     test_mod = importlib.import_module(mod_file_name)
+    testcase_name = test.get("name", "Test Case Unknown")
 
     rc = test_mod.run(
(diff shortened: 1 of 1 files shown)


udaysk23 (approved):

tests/parallel/test_parallel.py
```diff
@@ -73,6 +73,7 @@ def execute(test, args, results: dict):
     file_name = test.get("module")
     mod_file_name = os.path.splitext(file_name)[0]
     test_mod = importlib.import_module(mod_file_name)
+    testcase_name = test.get("name")
```
psathyan, line 76: Can we make it mandatory or the default is `Test Case Unknown`?

anrao19 (approved):

psathyan (approved):

PR #1893, opened 2022-10-11
Added yamls for bucket granular sync policy

TejasC88:
Signed-off-by: Tejas Chandramouli <tchandra@redhat.com>

Pass log : http://magna002.ceph.redhat.com:/ceph-qe-logs/tejas/cephci-run-GWGDG4/

### DIFF
--- a/pipeline/metadata/5.3.yaml
+++ b/pipeline/metadata/5.3.yaml
@@ -639,6 +639,22 @@
     - rgw
     - stage-2
 
+- name: "Testing extended RGW multisite bucket granular sync policy from primary"
+  suite: "suites/pacific/rgw/tier-1_rgw_ms-sync-policy-from-primary.yaml"
+  global-conf: "conf/pacific/rgw/rgw_multisite.yaml"
+  platform: "rhel-8"
+  rhbuild: "5.3"
+  inventory:
+    openstack: "conf/inventory/rhel-8-latest.yaml"
+    ibmc: "conf/inventory/ibm-vpc-rhel-8-latest.yaml"
+  metadata:
+    - sanity
+    - tier-2
+    - openstack
+    - ibmc
+    - rgw
+    - stage-4
+
 - name: "RGW Regression testing"
   suite: "suites/pacific/rgw/tier-2_rgw_regression.yaml"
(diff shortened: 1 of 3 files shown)


anrao19 (approved):

viduship (approved):

suites/pacific/rgw/tier-1-extn_rgw_multisite-secondary-to-primary.yaml
```diff
@@ -352,8 +352,34 @@ tests:
       polarion-id: CEPH-83575360
       module: sanity_rgw_multisite.py
       clusters:
-        ceph-sec:
+        ceph-pri:
```
ckulal, line 355: Looks good, But IMO, this suite file contains all the test-cases that are executed in secondary site and validated from primary site.

udaysk23 (approved):

suites/pacific/rgw/tier-1-extn_rgw_multisite-secondary-to-primary.yaml
TejasC88: @ckulal that is correct , but does it make sense to create a new tier-1 suite yaml for just 2 test cases?

ckulal: IMO, we can create new suites or else we can change suite name accordingly,
if we make the suite name change please make sure to change the names in respective metadata file also.

pipeline/metadata/5.3.yaml
```diff
@@ -639,6 +639,22 @@
...
+  inventory:
+    openstack: "conf/inventory/rhel-8-latest.yaml"
+    ibmc: "conf/inventory/ibm-vpc-rhel-8-latest.yaml"
+  metadata:
+    - sanity
+    - tier-2
+    - openstack
+    - ibmc
+    - rgw
+    - stage-2
```
anrao19, line 656: Looks like for tag "sanity, tier-2, stage-2" we already have 5 suites, so might need to change the stage.
And also if this suite need to be executed as a part of 6.0 testing then changes required in 6.0 metadata file

anrao19 (approved):

ckulal (approved):

PR #1956, opened 2022-11-16
renamed rados suites to correct tier names

tintumathew10:
Signed-off-by: tintumathew10 <tmathew@redhat.com>

For pis runs the report portal launches will pick tier names from the suite file names. 
Made changes to rados suite file names to correct tier names appearing in report portal.


manasagowri (approved):
LGTM

PR #2038, opened 2022-12-14
Support to execute multiple daemon operations.

sunilkumarn417:

- Support to execute multiple daemon operations.

```
  - test:
      name: OSD deployment on single SSD
      desc: Add OSD services on single SSD.
      module: test_daemon.py
      polarion-id: CEPH-9268
      config:
        steps:
          - config:
              command: add
              service: osd
              pos_args:
                - "cali003.ceph.redhat.com"
                - "/dev/nvme0n1"
          - config:
              command: add
              service: osd
              pos_args:
                - "cali003.ceph.redhat.com"
                - "/dev/sda"
          - config:
              command: add
              service: osd
              pos_args:
                - "cali004.ceph.redhat.com"
                - "/dev/nvme0n1"
```

**Results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-0FJLZB/OSD_deployment_on_single_SSD_0.log

### DIFF
--- a/ceph/ceph_admin/add.py
+++ b/ceph/ceph_admin/add.py
@@ -39,7 +39,7 @@ class AddMixin:
         pos_args = config["pos_args"]
         node = pos_args[0]
         host_id = get_node_by_id(self.cluster, node)
-        host = host_id.shortname
+        host = host_id.hostname
 
         if service == "osd":
             base_cmd.extend([f"{host}:{','.join(pos_args[1:])}"])
--- a/tests/cephadm/test_daemon.py
+++ b/tests/cephadm/test_daemon.py
@@ -26,16 +26,21 @@ def run(ceph_cluster, **kw):
           output, error   returned by the command.
     """
     config = kw.get("config")
+    tests = config.get("steps", [{"config": config}])
 
-    # Manage Ceph using ceph-admin orchestration
-    command = config.pop("command")
-    log.info("Executing service %s service" % command)
-    daemon = Daemon(cluster=ceph_cluster, **config)
+    for test in tests:
+        # Manage Ceph daemon using ceph-admin daemon orchestration commands
+        _test_cfg = test["config"]
+        command = _test_cfg.pop("command")
+        log.info("Executing service %s service" % command)
+        daemon = Daemon(cluster=ceph_cluster, **_test_cfg)
 
-    try:
-        method = fetch_method(daemon, command)
-        method(config)
-    finally:
-        # Get cluster state
-        get_cluster_state(daemon)
+        try:
+            method = fetch_method(daemon, command)
+            method(_test_cfg)
+        except Exception as err:
+            log.error(err)
+        finally:
+            # Get cluster state
+            get_cluster_state(daemon)
     return 0


tests/cephadm/test_daemon.py
```diff
@@ -26,16 +26,22 @@ def run(ceph_cluster, **kw):
...
-    log.info("Executing service %s service" % command)
-    daemon = Daemon(cluster=ceph_cluster, **config)
-
-    try:
-        method = fetch_method(daemon, command)
-        method(config)
-    finally:
-        # Get cluster state
-        get_cluster_state(daemon)
+    tests = config.get("steps") or [{"config": config}]
```
vamahaja, line 29: Can't use `tests = config.get("steps", [{"config": config}])` ?

tests/cephadm/test_daemon.py
```diff
@@ -26,16 +26,22 @@ def run(ceph_cluster, **kw):
...
+
+        try:
+            method = fetch_method(daemon, command)
+            method(_test_cfg)
+        except Exception as err:
+            log.error(err)
+            return 1
+        finally:
+            # Get cluster state
+            get_cluster_state(daemon)
```
vamahaja, line 46: What exactly this function exactly validating ?

tests/cephadm/test_daemon.py
```diff
@@ -26,16 +26,22 @@ def run(ceph_cluster, **kw):
...
+        _test_cfg = test["config"]
+        command = _test_cfg.pop("command")
+        log.info("Executing service %s service" % command)
+        daemon = Daemon(cluster=ceph_cluster, **_test_cfg)
+
+        try:
+            method = fetch_method(daemon, command)
+            method(_test_cfg)
+        except Exception as err:
+            log.error(err)
```
vamahaja, line 42: Why need to catch exception and return `1`, even exception is raised this test case wiil fail

ceph/ceph_admin/add.py
```diff
@@ -39,7 +39,7 @@ def add(self: DaemonProtocol, config: Dict):
         pos_args = config["pos_args"]
         node = pos_args[0]
         host_id = get_node_by_id(self.cluster, node)
-        host = host_id.shortname
+        host = host_id.hostname
```
vamahaja, line 42: Why this sudden change? `shortname` and `hostname` are different. It might affect all tests which are using this module.

vamahaja (changes requested):

tests/cephadm/test_daemon.py
sunilkumarn417: yes, we can change this. since `run.py` handles this.

tests/cephadm/test_daemon.py
sunilkumarn417: It's just cluster state after operation.

ceph/ceph_admin/add.py
sunilkumarn417: switching to hostname was there from long time. this module wasn't updated, since it didn't throw exception for PSI/IBM  instances, where shortname and hostname are same.
https://github.com/red-hat-storage/cephci/commit/098f32cb1b713460269d93183aebf6c70b8ef1c6

tests/cephadm/test_daemon.py
vamahaja: Do you think that would be sufficient ?

tests/cephadm/test_daemon.py
vamahaja: In this case we don't need to try..exception here, but need to raise proper exception.

vamahaja (changes requested):

tests/cephadm/test_daemon.py
sunilkumarn417: I did not change this part of the code now. This PR is to support multi daemon operations for some of the RADOS tests.

vamahaja (approved):

AmarnatReddy (approved):
LGTM

PR #2051, opened 2022-12-19
Update `rp_client` utility to get rp launch details

vamahaja:
# Add support in `utility/rp_client.py` to get report portal launcher details for specific launch ID

Added below new parameters to get details -
1. --launch_id
2. --output

e.g

- Upload result to Report Portal
`$ rp_client.py --config_file <str> --payload_dir <str>`

- Get Report Portal launcher details
`$ rp_client.py --config_file <str> --launch_id <int>`

- Store Report Portal to file path
`$ rp_client.py --config_file <str> --launch_id <int> --output <file-path>`


utility/rp_client.py
```diff
@@ -258,11 +294,29 @@ def match_start_line(line):
...
+    if launch_id:
+        launch_details = get_launch_details(config_file, launch_id)
+        log.info(f"Report portal launch id '{launch_id}' details are -\n{launch_details}")
+
+        file_path = args.get("--output")
+        if file_path:
+            with open(file_path, "w") as outfile:
+                outfile.write(json.dumps(launch_details, indent=4))
+
+        sys.exit(0)
```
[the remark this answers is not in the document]
vamahaja: No, we can't. As this code is inside `__main__` which is not function, it gives syntax error.

utility/rp_client.py
```diff
@@ -61,23 +63,65 @@
...
+        testsuites = list()
+        for ts in results:
+            testsuite = {
+                "id": ts.get("id"),
+                "name": ts.get("name"),
+                "status": ts.get('status'),
+                "total": ts.get('statistics').get('executions').get('total'),
+                "passed": ts.get('statistics').get('executions').get('passed'),
+            }
+            testsuite.update({"failed": int(testsuite.get("total")) - int(testsuite.get("passed"))})
```
manasagowri, line 105: There could be some skipped tests as well right, when we have enabled abort-on-fail, the tests after this test will not get executed at all. Is there any way we could accomodate them?

manasagowri (changes requested):

utility/rp_client.py
vamahaja: @manasagowri if there are no failures or skipped tests then rp response don't include this value. Removed step to calculate failed tcs, as rp response returns the same in case there are any. Using default value `0` incase `failed` and `skipped` values are missing from response.

tintumathew10 (commented):
LGTM

manasagowri (approved):
Good Work @vamahaja :)
LGTM

PR #2066, opened 2022-12-27
[LibFix] Add check for response keys in `rp_client` script

vamahaja:
`utility/rp_client.py` fails with an error `TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'` when report portal response doesn't contain `startTime` or `endTime` keys. Add check for respective keys in `response.content` and perform required compution.

Signed-off-by: Vaibhav Mahajan <vamahaja@redhat.com>


manasagowri (approved):
LGTM

PR #2108, opened 2023-01-19
[Fix] Update CephAdm Ansible test module to support boostraping using any builds

vamahaja:
Fix consists of -
- Refactoring of test case `tests/cephadm/test_cephadm_ansible_wrapper.py`
- Add test suite `test_cephadm_ansible_wrapper.yaml` for quincy
- Update playbook `bootstrap-cluster.yml` to add `image` parameter for downstream ceph image

Signed-off-by: Vaibhav Mahajan <vamahaja@redhat.com>

### DIFF
--- a/playbooks/bootstrap-cluster.yml
+++ b/playbooks/bootstrap-cluster.yml
@@ -1,7 +1,7 @@
 ---
 
 - name: Test 'cephadm_bootstrap' module
-  hosts: admin
+  hosts: installer
   gather_facts: false
   become: true
   any_errors_fatal: true
@@ -9,6 +9,7 @@
   tasks:
     - name: Bootstrap with dashboard set to `true` without credenatials
       cephadm_bootstrap:
+        image: "{{ image | default(False) }}"
         mon_ip: "{{ mon_ip }}"
         dashboard: true
         monitoring: false
--- a/suites/quincy/cephadm/test_cephadm_ansible_wrapper.yaml
+++ b/suites/quincy/cephadm/test_cephadm_ansible_wrapper.yaml
@@ -1,0 +1,34 @@
+#===============================================================================================
+# Tier-level: 1
+# Test-Suite: test_cephadm_ansible_wrapper.yaml
+# Test-Case: Perform cephadm operations using cephadm ansible modules
+#
+# Cluster Configuration:
+#    conf/quincy/cephadm/tier-1_5node_cephadm_bootstrap.yaml
+#
+#    4-Node cluster(RHEL-8.3 and above)
+#    3 MONS, 2 MDS, 1 MGR, 3 OSD and 2 RGW service daemon(s)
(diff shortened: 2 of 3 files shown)


cli/utilities/utils.py
```diff
@@ -199,3 +200,17 @@ def put_cephadm_ansible_playbook(
...
+
+
+def get_builds_by_rhbuild(rhbuild):
+    """
+    Get RHCS and RHEL version from rhbuild
+
+    Args:
+        rhbuild (str): build version
+    """
+    match = re.search(RHBUILD_PATTERN, rhbuild)
```
pranavprakash20, line 212: Vaibhav, to confirm, the RHBUIL_PATTERN expects the data to be in this format "5.3-rhel-8.7" correct? And the re.search in that case return, 
('5.3', 'rhel-8')

But as per our current approach, we are passing rhbuild value as  build value alone (e.g rhbuild 5.3)
So to accommodate this do we have to change the parameter passing as well?

If the idea is to get the rhel platform, can't we use the `--platform` value?

Please correct me if I got it wrong

utility/install_prereq.py
```diff
@@ -1,14 +1,15 @@
...
+from cli.utilities.utils import get_builds_by_rhbuild
 from utility.log import Log
-from utility.utils import fetch_build_artifacts
 
-RHEL8_EPEL_RELESE_RPM = (
-    "https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm"
-)
-RHCEPH5_TOOLS_REPO = "rhceph-5-tools-for-rhel-{}-x86_64-rpms"
-RHEL8_ANSIBLE_REPO = "ansible-2.9-for-rhel-8-x86_64-rpms"
+RHCS5_REPOS = {
```
pranavprakash20, line 5: One suggestion, can't we add these repos in a separate file rather than within the code? Maybe a 'repositories.yaml/ini' file under conf/inventory or any other location and then append to that file for future releases.

cli/utilities/utils.py
vamahaja: @pranavprakash20 yes, you are correct about regex pattern and `rhbuild` parameter accepts 5.3. But this `rhbuild` is used to get build details from receipe file which gives details same as  `RHBUIL_PATTERN` pattern.

utility/install_prereq.py
vamahaja: @pranavprakash20 agree, need to work on it

vamahaja:
Test Run - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-55wcb/

vamahaja:
Test Run - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-rf9aa/

MohitBis (dismissed):
LGTM

pranavprakash20 (dismissed):
LGTM

MohitBis (approved):
LGTM

PR #2116, opened 2023-01-24
Changing the metadata file format to a more readable one

manasagowri:

This PR is created to change the metadata file format to a more readable one and to support addition of metadata (suites to be tested) for different pipelines. The changes done in this PR are as given below

1. Change the metadata file format to the one mentioned in Row number 1 of the table in this document - https://docs.google.com/document/d/1955Llm8zgqSse7i28cqkLAzXmwkR7mkPny3iBKv9yVU/edit 
2. Create a python script to fetch the required data from the metadata file. This python module does the following
- Takes comma separated filters called tags as the input to filter test suites to be executed.
- Takes overrides to be applied to the run cli as a json input
- Takes the RHCS version for which metadata needs to be fetched (Metadata file is named after its RHCS version)
- Fetches all the metadata from the file specified. 
- If there is no pipeline name specified in tags parameter, the method returns empty dictionary as shown below and gives out an error message in logs saying the pipeline is not specified.
        {
            "scripts": {},
            "final_tier": false,
            "final_stage": false,
        }

- If a pipeline is specified, then based on whether any tiers or stages are specified or not, it fetches all the test suites necessary to be executed based on the tags specified and returns them in the same dictionary format above. When a tier and stage value is specified in tags, it also returns information about whether the tier and stage specified are final ones in the pipeline.
3. Modify tier_executor.groovy to adhere to the new pipeline names specified in metadata file (sanity, schedule, sanity_openstack_only, schedule_openstack_only, etc.,)
4. Modify openstack triggers for the pipeline psi-only-executor.groovy (which triggers everytime a new build is available) to only trigger sanity_openstack_only suites. And psi-scheduled-pipeline-executor.groovy (which triggers pipeline once every week) to only trigger schedule_openstack_only suites.
5. Modify pipeline/vars/v3.groovy's fetchstages method, to fetch metadata from the newly created python script for the regular pipelines. Upstream pipeline still uses the old metadata format, so the method to fetch is retained to be the older one for this.
6. Value of one of the parameters in the UMB message sent for unsigned-container-image-listener is changed. This is to ensure we handle triggering of both nightly and rc builds properly at the IBM end with the new metadata format changes. Once this PR is merged I will also be raising a PR to change the image listener in IBM to adapt to this new change. 

Few pipeline executions were done to test out the new changes. 
https://159.23.92.24/job/rhceph-pipeline-manasa/19/ 
https://159.23.92.24/job/rhceph-pipeline-manasa/22/

Some of the executions have been logrotated and removed from jenkins pipeline. But all in all, the code changes are working as expected in our pipeline.

### DIFF
--- a/pipeline/metadata/5.3.yaml
+++ b/pipeline/metadata/5.3.yaml
@@ -7,1283 +7,1028 @@
 # parameter required to create a similar custom file is suite name, suite yaml file, global configuration file,
 # platform, rhbuild, inventory and metadata information like frequency of execution, tier, cloud type, functional group and stage.
 #=====================================================================================
-suites:
-  - name: "Smoke test of RH Ceph build"
-    suite: "suites/pacific/cephadm/tier-0.yaml"
-    global-conf: "conf/pacific/cephadm/tier-0.yaml"
-    platform: "rhel-8"
-    rhbuild: "5.3"
-    inventory:
-      openstack: "conf/inventory/rhel-8-latest.yaml"
-      ibmc: "conf/inventory/ibm-vpc-rhel-8-latest.yaml"
-    metadata:
-      - sanity
-      - tier-0
-      - cvp
-      - rc
-      - live
-      - openstack
-      - ibmc
-      - dmfg
-      - stage-1
+overrides:
+  rhbuild: "5.3"
+  # log_dir: "" #other override parameters common to cloud can also be specified here.
+
+pipelines:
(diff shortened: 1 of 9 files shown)


anrao19 (dismissed):

tintumathew10 (commented):
LGTM

sunilkumarn417 (approved):

PR #2128, opened 2023-01-30
ODF InterOp support for RHCS 6.0

psathyan:

# Description

Adding support for `overrides` for InterOp testing

__Note__
The override for `6` needs to be removed once it is GA'ed


AmarnatReddy (approved):
LGMT

PR #2252, opened 2023-02-23
[Fix] Update logic to compare RHCS Version

vamahaja:
Fix consists of -
`cluster._Ceph__rhcs_version` comparsion fails when string value with rhel version is passed. Added regex to validate string and get rhcs version

### DIFF
--- a/ceph/ceph_admin/cephadm_ansible.py
+++ b/ceph/ceph_admin/cephadm_ansible.py
@@ -5,6 +5,8 @@ playbooks supported,
 - cephadm-purge-cluster.yaml
 - cephadm-clients.yaml
 """
+import re
+
 from ceph.ceph_admin.common import config_dict_to_string
 from utility.log import Log
 
@@ -39,7 +41,14 @@ class CephadmAnsible:
 
     def install_cephadm_ansible(self):
         """Enable ansible rpm repos and install cephadm-ansible."""
-        if float(self.cluster._Ceph__rhcs_version) >= 6.0:
+        rhcs_version = re.match(r"(\d)\.(.?)(.*)", self.cluster._Ceph__rhcs_version)
+        if not rhcs_version:
+            raise CephadmAnsibleError(
+                f"Invalid string for RHCS version - '{self.cluster.__Ceph_rhcs_version}'"
+            )
+
+        rhcs_version = rhcs_version.groups(0)[0]
+        if float(rhcs_version) >= 6.0:
             self.admin.exec_command(
                 cmd="yum install ansible-core -y --nogpgcheck",
                 sudo=True,


ceph/ceph_admin/cephadm_ansible.py
```diff
@@ -39,7 +41,8 @@ def __init__(self, cluster):
 
     def install_cephadm_ansible(self):
         """Enable ansible rpm repos and install cephadm-ansible."""
-        if float(self.cluster._Ceph__rhcs_version) >= 6.0:
+        rhcs_version = re.findall(r"(\d.\d)(.*)", self.cluster._Ceph__rhcs_version)
+        if rhcs_version and float(rhcs_version) >= 6.0:
```
pranavprakash20, line 45: Hi Vaibhav,

As per the recent failure 

"""
 if float(self.cluster._Ceph__rhcs_version) >= 6.0:
ValueError: could not convert string to float: '5.3-rhel-8'

"""
so the value of rhcs_version  is `5.3-rhel-8` right? the re.findall on this value returns => ` [('5.3', '-rhel-8')]`

When trying the above logic,

"""
ceph__rhcs_version = "5.3-rhel-8"
rhcs_version = re.findall(r"(\d.\d)(.*)", ceph__rhcs_version)
print(rhcs_version)
if rhcs_version and float(rhcs_version) >= 6.0:
    print("RHCEPH 6")"""
    
    
I can see the below error
"""
TypeError: float() argument must be a string or a number, not 'list'
"""

pranavprakash20 (changes requested):

ceph/ceph_admin/cephadm_ansible.py
vamahaja: @pranavprakash20 good catch, missed to call group function

ceph/ceph_admin/cephadm_ansible.py
```diff
@@ -39,7 +41,8 @@ def __init__(self, cluster):
 
     def install_cephadm_ansible(self):
         """Enable ansible rpm repos and install cephadm-ansible."""
-        if float(self.cluster._Ceph__rhcs_version) >= 6.0:
+        rhcs_version = re.match(r"(\d)\.(.?)(.*)", self.cluster._Ceph__rhcs_version)
+        if rhcs_version and float(rhcs_version.groups(0)[0]) >= 6.0:
```
pranavprakash20, line 45: The logic works fine for rhcs_versions like '4.x', '5.3-rhel-8' etc. 

One suggestion here, can we do the grouping along with the re.match itself rather than in the if condition? Because, that will help us reuse the rhcs_version variable again without adding the .groups 

rhcs_version = re.match(r"(\d)\.(.?)(.*)", self.cluster._Ceph__rhcs_version).groups(0)[0]
if rhcs_version and float(rhcs_version) >= 6.0:

vamahaja: Updated code.  Using `.groups` directly with `re.match` may throw error when unexpected string is passed.

pranavprakash20 (approved):
LGTM

PR #2253, opened 2023-02-23
[Lib] Add support to install ceph storage licence

vamahaja:
Add new class `cli.utility.SetupLicence` class to install ceph storage licence


pranavprakash20 (dismissed):
LGTM, but we need to keep a backlog to identify the build type

vamahaja:
Closing this PR, will raise separate PR for library and tests changes.

PR #2254, opened 2023-02-23
[LibFix] Update `Rpm.query` library to return package version

vamahaja:
`Rpm.query` raising exception instead of returning package version. Fix library and update respective usages.

### DIFF
--- a/tests/cephadm/test_cephadm_ansible_wrapper.py
+++ b/tests/cephadm/test_cephadm_ansible_wrapper.py
@@ -2,7 +2,7 @@ import os
 
 from cli.cephadm.ansible import Ansible
 from cli.cephadm.exceptions import ConfigNotFoundError
-from cli.utilities.packages import Rpm, RpmError
+from cli.utilities.packages import Rpm
 from cli.utilities.utils import (
     get_node_by_id,
     get_node_ip,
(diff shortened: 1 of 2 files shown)


tests/cephadm/test_cephadm_ansible_wrapper.py
```diff
@@ -66,9 +66,8 @@ def setup_cluster(ceph_cluster, config):
     cloud_type = config.get("cloud-type")
     build_type = config.get("build_type")
 
-    try:
-        Rpm(installer).query("cephadm-ansible")
-    except RpmError:
+    pkg = "cephadm-ansible"
+    if Rpm(installer).query(pkg) == f"package {pkg} is not installed":
```
pranavprakash20, line 70: One suggestion here, wouldn't it be better for the query to return True/False based on whether the package is installed or not? In that case the `== f"package {pkg} is not installed"` comparison can be done in packages.py/query method itself.

so then, we can rephrase the above statement as

```
if not Rpm(installer).query(pkg):
     <package is not installed and we can proceed with installation>
```

Or if the idea is to get the package version via query command, then we can return None if not installed and package version when the package is installed. In both the approaches, the above statement will work .

A suggestion only, if you feel otherwise please ignore.

pranavprakash20 (changes requested):

tests/cephadm/test_cephadm_ansible_wrapper.py
vamahaja: done

pranavprakash20 (approved):
LGTM

vamahaja:
Test Run - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-y89g5/

We can ignore failed TC's as failures are due to config error

PR #2272, opened 2023-03-01
[Lib] Add support to install ceph storage licence

vamahaja:
Add new class cli.utility.SetupLicence class to install ceph storage licence


pranavprakash20 (approved):
LGTM

PR #2273, opened 2023-03-01
[Fix] Add steps to enable licence to installing cephadm

vamahaja:
Install `ibm-cepg-storage-license, when `--custom-config ibm-build=True` is passed through `run.py`. Updated below modules to install license based on the parameter status -
1. ` ceph.ceph_admin.__init__` - To install ibm storage license 
2. `tests.cephadm.test_cephadm_ansible_wrapper` - To pass appropriate parameter to `utility.install_prereq`
3. `cli.utilities.utils` - To get appropriate `base_url`
4. `utility.install_prereq` - To  install ibm storage license & execute preflight with appropriate parameter

### DIFF
--- a/ceph/ceph_admin/__init__.py
+++ b/ceph/ceph_admin/__init__.py
@@ -9,6 +9,7 @@ scenarios for verifying and validating cephadm.
 """
 from typing import Dict
 
+from utility.install_prereq import SetupLicence
 from utility.log import Log
 
 from .bootstrap import BootstrapMixin
@@ -204,6 +205,8 @@ class CephAdmin(BootstrapMixin, ShellMixin):
             cmd += " --nogpghceck"
 
         for node in self.cluster.get_nodes(ignore="client"):
+            if self.config.get("ibm_build"):
+                SetupLicence.run(node)
             node.exec_command(
                 sudo=True,
                 cmd="yum install cephadm -y --nogpgcheck",
--- a/run.py
+++ b/run.py
@@ -830,6 +830,11 @@ def run(args):
                 docker_tag,
             )
 
+            if custom_config:
+                ibm_build = [c for c in custom_config if "ibm-build=" in c]
+                if ibm_build:
+                    config["ibm_build"] = bool(ibm_build[0].split("=")[1])
+
             config["ceph_docker_registry"] = docker_registry
             report_portal_description += f"docker registry: {docker_registry}"
             config["ceph_docker_image"] = docker_image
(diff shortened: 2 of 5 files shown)


tests/cephadm/test_cephadm_ansible_wrapper.py
```diff
@@ -68,6 +69,7 @@ def setup_cluster(ceph_cluster, config):
 
     if not Rpm(installer).query("cephadm-ansible"):
         SetUpSSHKeys.run(installer, nodes)
+        SetupLicence.run(installer)
```
pranavprakash20, line 72: This will install for all the ceph version right Vaibhav?

vamahaja: @pranavprakash20 yes, it's simple package `ibm-storage-ceph-license` independent of ceph version

pranavprakash20 (approved):
LGTM

pranavprakash20 (approved):
LGTM

run.py
```diff
@@ -77,6 +77,7 @@
         [--docker-image <image>]
         [--docker-tag <tag>]
         [--insecure-registry]
+        [--install-license]
```
sunilkumarn417, line 80: Instead of new arg, can we use `--custom-config ibm-build=True`?. However we have to go through new args for IBM later for complete pipeline.

sunilkumarn417: Let's not add anything in the run.py. so it would be easy to decouple this when we have proper  solution.

vamahaja: @sunilkumarn417 thanks for feedback. Updated code accordingly.

ceph/ceph_admin/__init__.py
```diff
@@ -204,6 +205,8 @@ def install(self, **kwargs: Dict) -> None:
             cmd += " --nogpghceck"
 
         for node in self.cluster.get_nodes(ignore="client"):
+            if self.config.get("ibm_build"):
```
pranavprakash20, line 208: One doubt here Vaibhav, So this if will pass even if we pass `ibm-build=False` also this if statement will work right? In that case, can we remove the =True for the ibm-build tag? Just `ibm-build` is sufficient right? Please ignore if I got it wrong

pranavprakash20 (approved):

vamahaja:
Test Result - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-hqyr9

sunilkumarn417 (approved):

PR #2284, opened 2023-03-02
update ocs global conf.

sunilkumarn417:
New RHCS multi-client code changes will not install cephadm on the client node, it should be done explicitly by configure client test case. 
Global configuration files should not have client role on the installer node.  

In 7_node config had installer node which has client role as well, which is avoided by above code commits.
- Fix issue with ocs interop conf file by moving client role from installer node to new node. 

> ceph.ceph.CommandFailed: cephadm -v shell -- ceph status Error:  bash: cephadm: command not found
>  10.0.208.199

### DIFF
--- a/conf/pacific/integrations/7_node_ceph.yaml
+++ b/conf/pacific/integrations/7_node_ceph.yaml
@@ -10,7 +10,6 @@ globals:
           - installer
           - mgr
           - mon
-          - client
 
       node2:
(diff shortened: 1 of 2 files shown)


AmarnatReddy (approved):
LGTM

AmarnatReddy (commented):
LGTM

PR #2320, opened 2023-03-10
Adding test coverage for datalog and mdlog trim

viduship:
# Description

Automating mdlog and datalog trimming.

This PR would test log trimming (mdlog, datalog, and bilog) happens after the default interval of 20 minutes

The test covered: CEPH-10540, and CEPH-10544.

logs http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-A5MZHX


anrao19:
@viduship, Currently this suite will take amost  "1h 55m 51s" for execution, and this suite is a part of sanity.
After adding TC of trimming execution time increased to "4h 7s".

Since execution time is more than 2 hours.
1. we might need to move this as a part of sheduled run or
2. Split this suite into 2 suite then add both as a part of sanity

anrao19 (approved):

TejasC88 (approved):

PR #2337, opened 2023-03-16
CEPH-83571714: Verify reduction in blocks when data in an object is removed

harshkumarRH:
[CEPH-83571714](https://polarion.engineering.redhat.com/polarion/#/project/CEPH/workitem?id=CEPH-83571714): Tier 2 test case to verify block allocation and reduction when data is written to an object and removed from the object respectively

_ceph-objectstore-utility_ is used to fetch the object dump from the osd node

Test suites modified:
- suites/pacific/rados/tier-2_rados_test-osd-rebalance.yaml

Test case and modules added:
- tests/rados/test_objectstore_block.py
 _systemctl_ in ceph/utils.py
- _get_config_ in tests/rados/monitor_configurations.py
- _configure_osd_flag_ in ceph/rados/utils.py

References:
- https://access.redhat.com/documentation/en-us/red_hat_ceph_storage/5/html/troubleshooting_guide/troubleshooting-ceph-objects#listing-objects_diag
- https://access.redhat.com/documentation/en-us/red_hat_ceph_storage/5/html/administration_guide/osd-bluestore#tuning-ceph-bluestore-using-bluestore-min-alloc-size-parameter_admin
- https://docs.ceph.com/en/quincy/rados/configuration/bluestore-config-ref/#minimum-allocation-size

**Steps:**

1. Configure Ceph cluster, create pool and write data to an object
2. Fetch the acting pg set for the object and corresponding osd nodes
3. Stop any one of the OSDs from the acting pg set using systemctl
4. Capture the object dump using ceph-objectstore-tool
5. Start the OSD which was previously stopped
6. Overwrite the data in the object with a smaller chunk
7. Stop the same OSD and fetch object dump using ceph-objectstore-tool
8. Verify the difference in block stats and decrease in the number of blocks

**Logs:**
RHCS 5.3 - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-6FJFWL/
Test passes 10/10 on RHCS 5.3 but only 5/10 on RHCS 6.0

### DIFF
--- a/ceph/ceph_admin/osd.py
+++ b/ceph/ceph_admin/osd.py
@@ -236,3 +236,22 @@ class OSD(ApplyMixin, Orch):
         osd_id = config["pos_args"][0]
         base_cmd.append(str(osd_id))
         return self.shell(args=base_cmd)
+
+    def flag(self, config: dict):
+        """
+        Execute the command ceph osd set/unset flag.
+        Args:
+            config (Dict): OSD flag configuration parameters
+        Returns:
+          output, error   returned by the command.
+        Example::
+            config:
+                command: set/unset
+                base_cmd_args:
+                    verbose: true
+                flag: value of flag
+        """
+        command = config.get("command")
+        flag = config.get("flag")
+        cmd = f"ceph osd {command} {flag}"
+        return self.shell(args=[cmd])
--- a/ceph/rados/core_workflows.py
+++ b/ceph/rados/core_workflows.py
@@ -925,12 +925,14 @@ class RadosOrchestrator:
         log.info(f"Created the ec profile : {profile_name} and pool : {pool_name}")
         return True
 
-    def change_osd_state(self, action: str, target: int) -> bool:
+    def change_osd_state(self, action: str, target: int, timeout: int = 15) -> bool:
         """
         Changes the state of the OSD daemons wrt the action provided
         Args:
-            action: operation to be performed on the service, i.e start, stop, restart
+            action: operation to be performed on the service, i.e
+            start, stop, restart, disable, enable
             target: ID osd the target OSD
+            timeout: timeout in seconds, (default = 10s)
         Returns: Pass -> True, Fail -> False
         """
         cluster_fsid = self.run_ceph_command(cmd="ceph fsid")["fsid"]
@@ -943,8 +945,34 @@ class RadosOrchestrator:
             f"Performing {action} on osd-{target} on host {host.hostname}. Command {cmd}"
         )
         host.exec_command(sudo=True, cmd=cmd)
-        # Sleeping for 5 seconds for status to be updated
-        time.sleep(5)
+        # verifying the osd state
+        if action in ["start", "stop"]:
+            start_time = datetime.datetime.now()
+            timeout_time = start_time + datetime.timedelta(seconds=timeout)
+
+            while datetime.datetime.now() <= timeout_time:
+                osd_status, status_desc = self.get_daemon_status(
+                    daemon_type="osd", daemon_id=target
+                )
+                log.info(f"osd_status: {osd_status}, status_desc: {status_desc}")
+                if (osd_status == 0 or status_desc == "stopped") and action == "stop":
+                    break
+                elif (
+                    osd_status == 1 or status_desc == "running"
+                ) and action == "start":
+                    break
+                time.sleep(3)
+
+            if action == "stop" and osd_status != 0:
+                log.error(f"Failed to stop the OSD.{target} service on {host.hostname}")
+                return False
+            if action == "start" and osd_status != 1:
+                log.error(
+                    f"Failed to start the OSD.{target} service on {host.hostname}"
+                )
+                return False
+        else:
+            time.sleep(5)
         return True
 
     def fetch_host_node(self, daemon_type: str, daemon_id: str = None):
@@ -1139,3 +1167,37 @@ class RadosOrchestrator:
         pg_query = self.run_ceph_command(cmd=cmd)
         log.debug(f"The status of pg : {pg_id} is {pg_query['state']}")
         return pg_query["state"]
+
+    def get_osd_map(self, pool: str, obj: str, nspace: str = None) -> dict:
+        """
+        Retrieve the osd map for an object in a pool
+        Args:
+            pool: pool name to which the object belongs
+            obj: object name whose osd map is to retrieved
+            nspace (optional): namespace
+        Returns:  dictionary output of ceph osd map command
+        """
+        cmd = f"ceph osd map {pool} {obj}"
+        if nspace:
+            cmd += " " + nspace
+
+        return self.run_ceph_command(cmd=cmd)
+
+    def get_daemon_status(self, daemon_type, daemon_id) -> tuple:
+        """
+        Returns the status of a specific daemon using ceph orch ps utility
+        Usage: orch ps --daemon_type <> --daemon_id <>
+        Args:
+            daemon_type: type of daemon known to orchestrator
+            daemon_id: id of daemon provided in daemon_type
+        Returns: tuple containing status of the daemon (0 or 1) and
+                 status description (running or stopped)
+        """
+
+        cmd_ = (
+            f"ceph orch ps --daemon_type {daemon_type} "
+            f"--daemon_id {daemon_id} --refresh"
+        )
+        orch_ps_out = self.run_ceph_command(cmd=cmd_)[0]
+        log.debug(orch_ps_out)
+        return orch_ps_out["status"], orch_ps_out["status_desc"]
--- a/ceph/rados/utils.py
+++ b/ceph/rados/utils.py
@@ -204,3 +204,25 @@ def add_osd(ceph_cluster, host, device_path, osd_id):
     log.info(f"Executing daemon {config.pop('command')} service")
     daemon = Daemon(cluster=ceph_cluster, **config)
     daemon.add(config)
+
+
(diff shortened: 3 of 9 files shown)


tests/rados/test_objectstore_block.py
```diff
@@ -0,0 +1,302 @@
...
+        "of blocks when data is written to an object and removed from an object"
+    )
+
+    try:
+        # create pool with given config
+        if config["create_pool"]:
+            rados_obj.create_pool(pool_name=pool_name)
+
+        # retrieve value of config parameter bluestore_min_alloc_size_hdd
+        bluestore_min_alloc_size, _ = installer_node.exec_command(
```
pdhiran, line 55: Although not wrong, If commands needs to be run inside cephadm shell & on installer node, ` cephadm.shell() ` method can be used here. The method opens a shell on installer node and executes the command.

tests/rados/test_objectstore_block.py
```diff
@@ -0,0 +1,302 @@
...
+            obj_name=object_name,
+            nobj=(iterations - 1),
+        )
+
+        osd_map = rados_obj.get_osd_map(pool=pool_name, obj=object_name)
+        acting_pg_set = osd_map["acting"]
+        log.info(f"Acting pg set for {object_name} in {pool_name}: {acting_pg_set}")
+        if not acting_pg_set:
+            log.error("Failed to retrieve acting pg set")
+            return 1
```
pdhiran, line 76: Please Raise Exceptions in case of failures instead of return 1.

tests/rados/test_objectstore_block.py
```diff
@@ -0,0 +1,302 @@
...
+
+    try:
+        # create pool with given config
+        if config["create_pool"]:
+            rados_obj.create_pool(pool_name=pool_name)
+
+        # retrieve value of config parameter bluestore_min_alloc_size_hdd
+        bluestore_min_alloc_size, _ = installer_node.exec_command(
+            sudo=True,
+            cmd="cephadm shell -- ceph config get osd bluestore_min_alloc_size_hdd",
```
pdhiran, line 57: Please write this as a method `config_get(**kwargs)`

tests/rados/test_objectstore_block.py
```diff
@@ -0,0 +1,302 @@
...
+        )
+
+        # Delete the created osd pool
+        if config.get("delete_pool"):
+            rados_obj.detete_pool(pool=pool_name)
+
+    return 0
+
+
+def check_osd_status(
```
pdhiran, line 257: We can merge this method with : ` change_osd_state ` . Please handle restart command

harshkumarRH: Done, merged the code with `change_osd_state` in `ceph/rados/core_workflows.py`. Verification check is only being performed for start and stop, for everything else, a blind sleep of 5 secs is being used.

tests/rados/test_objectstore_block.py
harshkumarRH: Replaced all applicable return statements with raise exceptions

tests/rados/test_objectstore_block.py
harshkumarRH: Done, added a new module `get_config(**kwargs)` in `tests/rados/monitor_configurations.py`

tests/rados/test_objectstore_block.py
harshkumarRH: Noted, as new module `get_config(**kwargs)` has been added in `tests/rados/monitor_configurations.py`
Execution with `cephadm.shell()` is no longer necessary

harshkumarRH:
Updated logs after addressing review comments
RHCS 5.3 - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-XET016

ceph/ceph_admin/osd.py
```diff
@@ -267,3 +267,20 @@ def osd_in(self, config: dict):
...
+                base_cmd_args:
+                    verbose: true
+                flag: value of flag
+        Returns:
+          output, error  returned by the command
+        """
+        command = config.get("command")
+        flag = config.get("flag")
+        cmd = f"ceph osd {command} {flag}"
+        return self.shell(args=[cmd])
```
pdhiran, line 286: Suggestion: Instead of user processing the o/p by the calling function, we can perform those checks in this method itself and return true/false based on status! Would avoid code repetition in future.

ceph/rados/utils.py
```diff
@@ -236,3 +236,25 @@ def add_osd(ceph_cluster, host, device_path, osd_id):
...
+        flag - value of the flag
+          example:pause|noup|nodown|noout|noin|nobackfill|
+            norebalance|norecover|noscrub|nodeep-scrub|notieragent
+    Returns: True/False
+    """
+    config = {"command": action, "service": "osd", "flag": flag}
+    log.info(f"Executing OSD {action} {flag}")
+    osd = OSD(cluster=ceph_cluster, **config)
+    out, err = osd.flag(config)
+    if f"{flag} is {action}" in err:
```
pdhiran, line 258: are the returns of this commands directed to err stream?

tests/rados/monitor_configurations.py
```diff
@@ -204,6 +204,23 @@ def set_config(self, **kwargs):
...
+        """
+        Retrieve the config param in monitor config database
+        Usage - config get <section> [<name>]
+        Config parameters
+        Args:
+            **kwargs: Any other param that needs to be set
+                1. section: which section of daemons to target
+                    allowed values: global, mon, mgr, osd, mds, client
+                2. name: name of the config param for the selection
+        Returns: tuple output of ceph config get <> command
```
pdhiran, line 217: please add the example of the return here, so caller functions know what to expect. ( since we are returning a tuple, not just the value requested.

Also, Please handle scenarios were the configs are not yet set. If you try to fetch those, the command would throw exception.

tests/rados/test_objectstore_block.py
```diff
@@ -0,0 +1,223 @@
...
+            ]
+            log.info(new_acting_pg_set)
+            if new_acting_pg_set == acting_pg_set:
+                break
+            elif i == 1:
+                log.error(
+                    f"{acting_osd_id} could not return to PG {new_acting_pg_set} within timeout. "
+                    f"New acting pg set({new_acting_pg_set} and old acting pg set({acting_pg_set}) are not same"
+                )
+                raise Exception("acting pg set has changed")
```
pdhiran, line 151: Please set `nobackfill`, `norebalance`, `norecover`, `noout`, `nodown` . This should prevent the OSDs from being marked out, and the PG sets changing.

tests/rados/test_objectstore_block.py
```diff
@@ -0,0 +1,223 @@
...
+        assert rados_obj.change_osd_state(action="start", target=acting_osd_id)
+        assert rados_obj.change_osd_state(action="enable", target=acting_osd_id)
+
+        for i in range(2):
+            time.sleep(3)
+            new_acting_pg_set = rados_obj.get_osd_map(pool=pool_name, obj=object_name)[
+                "acting"
+            ]
+            log.info(new_acting_pg_set)
+            if new_acting_pg_set == acting_pg_set:
```
pdhiran, line 144: If the below flags work, this check should pass

pdhiran (commented):
Looks good with minor changes required!

harshkumarRH:
Updated logs after addressing review comments
RHCS 5.3 - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-GCSV5W

ceph/rados/core_workflows.py
```diff
@@ -1175,3 +1203,21 @@ def get_osd_df_stats(
             cmd += " " + filter
 
         return self.run_ceph_command(cmd=cmd)
+
+    def get_daemon_status(self, daemon_type, daemon_id) -> tuple:
+        """
+        Returns the status of a specific daemon using ceph orch ps utility
+        Args:
+            daemon_type:
```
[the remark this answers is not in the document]
harshkumarRH: Updated with latest commit

ceph/ceph_admin/osd.py
harshkumarRH: Yes agreed, however this function is being called by `configure_osd_flag` in [ceph/rados/utils.py](https://github.com/red-hat-storage/cephci/pull/2337/files#diff-5232d7b6ad4cdb18e68353cedf6d3198168562d59e87e2e6882b99ffffc7be1b) 
I have followed the workflow which is already in place for OSD related commands.

tests/rados/monitor_configurations.py
harshkumarRH: Example has been added and return has been changed from tuple to string.

Even if the configs are not set, it will not raise an exception, rather return an empty string (verified manually) or the default compiled value.
Ref - https://docs.ceph.com/en/latest/rados/configuration/ceph-conf/#commands

tests/rados/test_objectstore_block.py
harshkumarRH: Setting these flags did not help with RHCS 6.0 execution, and for RHCS 5.3 there flags are not required.

tests/rados/test_objectstore_block.py
harshkumarRH: As discussed over call, the osd no recovery flags are not needed for RHCS 5.3, this approach will be revisited once RHCS 6.0 failure has been investigated

pdhiran (approved):
Looks Good

PR #2343, opened 2023-03-17
test_clay_ecpool.py: Addition of workflows for CLAY profile based EC pools

pdhiran:
Steps Performed:
1. Create EC Pool with CLAY Profile
2. Create replicated metadata pool and use the Clay poll created as data pool for RBD Images
3. Test the pool with various autoscaler values
4. Test the pool with Bulk flags ( PG Splits / Merge )
5. Scrubs and Deep Scrubs on the pool
6. Enable and verify various inline compression algorithms
7. Perform Repair on the PGs on Clay pool
8. Recovery with only K shards of EC Pool


tintumathew10 (approved):
LGTM

PR #2344, opened 2023-03-17
[Lib] Add libraries to execute RESTful API

vamahaja:
Add below libraries to support RESTful API -

- cephapi - `Api` class to execute basic methods with request library.
- cephapi.auth.Auth - `Auth` class to supoort `api/auth` endpoint.
- cephapi.auth.check.Check - `Check` class to support `api/auth/check` endpoind.
- cephapi.auth.check.Logout - `Logout` class to support `api/auth/logout` endpoind.

### DIFF
--- a/api/__init__.py
+++ b/api/__init__.py
@@ -1,0 +1,104 @@
+import json
+
+import requests
+
+from utility.log import Log
+
+LOG = Log(__name__)
+
+# Constant for request URL verifier
+DEFAULT_VERIFY = False
+
+
+# Exception for status code `400`
+class BadRequestError(Exception):
+    pass
+
+
+# Exception for status code `401`
+class UnauthorizedError(Exception):
+    pass
+
+
+# Exception for status code `403`
+class ForbiddenError(Exception):
+    pass
+
+
+# Exception for status code `500`
+class InternalServerError(Exception):
+    pass
+
+
+class Api:
+    """Interface for request API methods"""
+
+    def __init__(self, url, api):
+        # Disable insecure request warning in response
+        requests.packages.urllib3.disable_warnings(
+            requests.packages.urllib3.exceptions.InsecureRequestWarning
+        )
+
+        # Remove '/' from url and append with API
+        self.url = f"{url.strip('/')}/{api}"
+
+    def _response(self, response):
+        """Validate request response
+
+        Args:
+            response (request.response) : Request response
+        """
+        if response.status_code == 400:
+            raise BadRequestError(response.content)
+
+        elif response.status_code == 401:
+            raise UnauthorizedError(response.content)
+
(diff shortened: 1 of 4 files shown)


api/cephapi/auth/logout.py
```diff
@@ -0,0 +1,9 @@
+from api import Api
+
+
+class Logout(Api):
```
pranavprakash20, line 4: Do we really need a class for Logout? This could be a method under auth.py itslef? 
Please correct me if I have got it wrong

api/__init__.py
```diff
@@ -0,0 +1,40 @@
+import requests
+
+DEFAULT_CONTENT_TYPE = "application/json"
+DEFAULT_ACCEPT = "application/vnd.ceph.api.v1.0+json"
```
sunilkumarn417, line 4: Since this is generic api package module, Go as per the user requests and make it independent.

This suggestion is because,Since we may have to work on prometheues, grafana and any other APIs in future.

sunilkumarn417: Get headers from the ceph Restful API module.

api/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+    ):
+        headers = {
+            "Content-Type": content_type,
+            "Accept": accept,
+        }
+
+        return requests.post(url=self.url, verify=verify, headers=headers, data=data)
+
+    def put(self):
+        pass
```
sunilkumarn417, line 34: please define everything

api/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def post(
+        self,
+        data,
+        verify=DEFAULT_VERIFY,
+        content_type=DEFAULT_CONTENT_TYPE,
+        accept=DEFAULT_ACCEPT,
+    ):
+        headers = {
```
sunilkumarn417, line 26: Accept more headers for future use.

api/__init__.py
vamahaja: +1, will have default header instead. which will be overridden while calling method

api/__init__.py
vamahaja: +1. Instead of these values, will accept header

api/__init__.py
vamahaja: Added these values as placeholders for now. Will define them once add methods to automate.

MohitBis (approved):
LGTM

pranavprakash20 (approved):
LGTM

MohitBis (approved):
LGTM

vamahaja:
Testrun - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-5oih5/

PR #2385, opened 2023-03-27
CEPH-9499 - Shrink images on local/primary cluster while mirroring is happening

rkhambad:


suites/quincy/rbd/tier-2_rbd_mirror_regression.yaml
```diff
@@ -296,3 +296,30 @@ tests:
...
+  - test:
+      name: Image shrink in primary cluster
+      module: test_rbd_mirror_shrink_image_primary.py
+      clusters:
+        ceph-rbd1:
+          config:
+            journal:
+              ec_pool_config:
+                size: 2G
+                io_total: 10
```
sunilangadi2, line 309: I think you are not using this `io_total` config option we can remove it?

rkhambad: I am using that for running IOs , I am not going  with default values for io_total.

tests/rbd_mirror/test_rbd_mirror_shrink_image_primary.py
```diff
@@ -0,0 +1,131 @@
...
+        image = config[pool_type]["image"]
+        imagespec = pool + "/" + image
+        resize = "1G"
+        mirror1.resize_image(imagespec=imagespec, size=resize)
+
+        if config[pool_type]["mode"] == "image":
+            mirror1.create_mirror_snapshot(imagespec)
+            mirror2.wait_for_status(imagespec=imagespec, state_pattern="up+replaying")
+        else:
+            mirror2.wait_for_status(imagespec=imagespec, state_pattern="up+replaying")
```
VasishtaShastry, line 42: I think appropriate condition would be to verify that `config[pool_type]["mirrormode"]` is snapshot right ?

I'm not sure why are we waiting for up+replaying here
wouldn't it be the same before shrinking the image ?

tests/rbd_mirror/test_rbd_mirror_shrink_image_primary.py
```diff
@@ -0,0 +1,131 @@
...
+    except Exception as e:
+        log.exception(e)
+        return 1
+
+    finally:
+        mirror1.clean_up(peercluster=mirror2, pools=[pool])
+
+
+def run(**kw):
+    log.info("Starting RBD mirroring test case - CEPH-9499")
```
VasishtaShastry, line 65: Please add any statements after the method doc
https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google

tests/rbd_mirror/test_rbd_mirror_shrink_image_primary.py
```diff
@@ -0,0 +1,131 @@
...
+                mode: image
+                mirrormode: snapshot
+                size: 2G
+                io_total: 10
+    Returns:
+        0 - if test case pass
+        1 - it test case fails
+
+    Test case covered -
+    CEPH-11416 - Pool Mirror - Try marking the secondary on the relationship for deletion, and observe the behaviour
```
VasishtaShastry, line 81: Is this the correct one?

Guess its not based on PR description.
Is this module designed only to handle a particular TC ? If yes, can we make this doc as module doc instead method doc ?

tests/rbd_mirror/test_rbd_mirror_shrink_image_primary.py
```diff
@@ -0,0 +1,131 @@
...
+            if test_image_shrink_primary(
+                mirror_obj.get("rep_rbdmirror"), "rep_pool_config", **kw
+            ):
+                return 1
+
+            log.info("Executing test on ec pool")
+            if test_image_shrink_primary(
+                mirror_obj.get("ec_rbdmirror"), "ec_pool_config", **kw
+            ):
+                return 1
```
VasishtaShastry, line 125: It can be observed that there are similar lines of code
Can we make use of iterator ?

tests/rbd_mirror/test_rbd_mirror_shrink_image_primary.py
```diff
@@ -0,0 +1,131 @@
...
+        out2 = rbd2.image_info(pool, image)
+        if out1["size"] == out2["size"]:
+            log.info(f"New image size {resize} is updated in secondary")
+            return 0
+        log.error(f"New image size {resize} is not updated in secondary")
+        return 1
+
+    except Exception as e:
+        log.exception(e)
+        return 1
```
VasishtaShastry, line 58: I think we could use run method doc if we don't have anything special (explicit) to perform here in exception block
BAsed on no explicit operation in exception block, can we make use of run method's exception block ?

VasishtaShastry (changes requested):

tests/rbd_mirror/test_rbd_mirror_shrink_image_primary.py
rkhambad: Done

tests/rbd_mirror/test_rbd_mirror_shrink_image_primary.py
rahullepakshi: Just to understand, will the state_pattern be "up+replaying" during time it takes to resize?

VasishtaShastry: I think yes, shrinking image wouldn't affect image state right ?

suites/quincy/rbd/tier-2_rbd_mirror_regression.yaml
sunilangadi2: ok got it.

tests/rbd_mirror/test_rbd_mirror_shrink_image_primary.py
rkhambad: > I think appropriate condition would be to verify that `config[pool_type]["mirrormode"]` is snapshot right ?
> 
> I'm not sure why are we waiting for up+replaying here wouldn't it be the same before shrinking the image ?

As per the initial_rbd_config method, we are not setting mirror mode for journal based mirroring, so we can not check as config[pool_type]["mirrormode"]` is snapshot because of the keyerror so went with "mode".

rkhambad: Added the new method for snapshot mirroring wait_for_snapshot_complete(), to check for snapshot completion and then check for image size in secondary cluster.
For journal based mirroring using existing wait_for_replay_complete() method to wait for replay completion and then check for image size in secondary cluster

tests/rbd_mirror/rbd_mirror_utils.py
```diff
@@ -475,6 +475,24 @@ def wait_for_replay_complete(self, imagespec):
...
+        """
+        while True:
+            log.info(f"verifying for snapshot completion for {imagespec}")
+            data = self.mirror_status("image", imagespec, "description").split(", ")[1:]
+            for value in data:
+                out = eval(value)
+                if out["local_snapshot_timestamp"] == out["remote_snapshot_timestamp"]:
+                    log.info("Snapshot is successfully completed")
+                return 0
+            else:
```
manasagowri, line 491: Did you test this change? if and else are at different indentation levels. It should give some error I believe

manasagowri (changes requested):

preethisn (commented):
LGTM

tests/rbd_mirror/rbd_mirror_utils.py
```diff
@@ -475,6 +475,24 @@ def wait_for_replay_complete(self, imagespec):
                 if int(out.split("=")[-1]) == 0:
                     return 0
 
+    def wait_for_snapshot_complete(self, imagespec):
+        """verify for snapshot completion
+        Args:
+            imagespec: Image specification.
+        """
+        while True:
```
manasagowri, line 483: Could you please add an optimistic timeout, maybe some 300 to 600 seconds, post which if snapshot isn't complete, it exits from the method. Else this loop will continue indefinitely right?

manasagowri (changes requested):

tests/rbd_mirror/rbd_mirror_utils.py
rkhambad: Added timeout as 600 sec.

tests/rbd_mirror/test_rbd_mirror_shrink_image_primary.py
manasagowri: LGTM

tests/rbd_mirror/test_rbd_mirror_shrink_image_primary.py
manasagowri: LGTM

tests/rbd_mirror/test_rbd_mirror_shrink_image_primary.py
manasagowri: LGTM

tests/rbd_mirror/rbd_mirror_utils.py
manasagowri: LGTM

tests/rbd_mirror/rbd_mirror_utils.py
manasagowri: LGTM

manasagowri (approved):
Group review done. All review comments have been addressed.

PR #2403, opened 2023-03-29
Rados tier-4 tests: Pool - OSD resiliency tests

pdhiran:
Tests include the entire suite, performing the workflows required for the tests.
1. Reboot of single osd and health check
2. Rolling Reboot OSD hosts
3. Stopping and starting OSD daemons
4. restart all OSD daemons belonging to single pg
5. Removal and addition of OSD daemons
6. Removal and addition of OSD Hosts

Pass log : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-ZEERGO

### DIFF
--- a/ceph/rados/core_workflows.py
+++ b/ceph/rados/core_workflows.py
@@ -1314,3 +1314,73 @@ class RadosOrchestrator:
             pgid_list.append(pg_stats["pgid"])
 
         return pgid_list
+
+    def run_pool_sanity_check(self):
+        """
+        Runs sanity on the pools after triggering scrub and deep-scrub on pools, waiting 600 Secs
+
+        This method is used to assess the health of Pools after any operation, where in a scrub and deep scrub is
+        triggered, and the method scans the cluster for few health warnings, if generated
+
+        Returns: True-> Pass,  false -> Fail
+        """
+        self.run_scrub()
+        self.run_deep_scrub()
+        time.sleep(10)
+
+        end_time = datetime.datetime.now() + datetime.timedelta(seconds=600)
+        flag = False
+        while end_time > datetime.datetime.now():
+            status_report = self.run_ceph_command(cmd="ceph report")
+            ceph_health_status = status_report["health"]
+            health_warns = (
+                "PG_AVAILABILITY",
+                "PG_DEGRADED",
+                "PG_RECOVERY_FULL",
+                "TOO_FEW_OSDS",
+                "PG_BACKFILL_FULL",
+                "PG_DAMAGED",
+                "OSD_SCRUB_ERRORS",
+                "OSD_TOO_MANY_REPAIRS",
+                "CACHE_POOL_NEAR_FULL",
+                "SMALLER_PGP_NUM",
+                "MANY_OBJECTS_PER_PG",
+                "OBJECT_MISPLACED",
+                "OBJECT_UNFOUND",
+                "SLOW_OPS",
+            )
+
+            flag = (
+                False
+                if any(
+                    key in health_warns for key in ceph_health_status["checks"].keys()
+                )
+                else True
+            )
+            if flag:
+                log.info("No warnings on the cluster")
+                break
+
+            log.info(
+                f"Observing a health warning on cluster {ceph_health_status['checks'].keys()}"
+            )
+            time.sleep(10)
+
+        if not flag:
+            log.error(
+                "Health warning generated on cluster and not cleared post waiting of 600 seconds"
+            )
+            return False
+
+        log.info("Completed check on the cluster. Pass!")
+        return True
+
+    def get_osd_hosts(self):
+        """
+        lists the names of the OSD hosts in the cluster
+        Returns: list of osd host names as used in the crush map
+
+        """
+        cmd = "ceph osd tree"
+        osds = self.run_ceph_command(cmd)
+        return [entry["name"] for entry in osds["nodes"] if entry["type"] == "host"]
--- a/conf/pacific/rados/13-node-cluster.yaml
+++ b/conf/pacific/rados/13-node-cluster.yaml
@@ -1,0 +1,75 @@
+# Rados tier-4 test configuration
+# Used for testing pools with OSD addition, deletion , start, stop & restart.
+# Suite also contains spare OSD hosts with label "osd-bak" for OSD deployment as required
+# Contains 12 RHCS Cluster hosts and 1 RHCS Client.
+globals:
+  - ceph-cluster:
+      name: ceph
+      node1:
+        role:
+          - _admin
+          - mon
+          - mgr
+          - installer
+          - alertmanager
+          - grafana
+          - prometheus
+      node2:
+        role:
+          - mon
+          - mgr
+          - mds
+          - nfs
+          - alertmanager
+      node3:
+        role:
+          - osd
+        no-of-volumes: 4
+        disk-size: 25
+      node4:
+        role:
+          - osd
+        no-of-volumes: 4
+        disk-size: 25
+      node5:
+        role:
+          - osd
+        no-of-volumes: 4
+        disk-size: 25
+      node6:
+        role:
+          - mon
+          - mgr
+          - mds
+          - nfs
+      node7:
+        role:
+          - mon
+          - rgw
+      node8:
+        role:
+          - client
+      node9:
+        role:
+          - osd
+        no-of-volumes: 4
+        disk-size: 25
+      node10:
+        role:
+          - osd
+        no-of-volumes: 4
+        disk-size: 25
+      node11:
+        role:
+          - mon
+          - rgw
+      node12:
+        role:
+          - osd-bak
+        no-of-volumes: 4
+        disk-size: 25
+      node13:
+        role:
+          - osd-bak
+        no-of-volumes: 4
+        disk-size: 25
--- a/suites/pacific/rados/tier-4_rados_test-pool-osd-recovery.yaml
+++ b/suites/pacific/rados/tier-4_rados_test-pool-osd-recovery.yaml
@@ -1,0 +1,209 @@
+# Suite contains tier-4 tests to verify and test ceph pools upon OSD changes
+tests:
+  - test:
+      name: setup install pre-requisistes
+      desc: Setup phase to deploy the required pre-requisites for running the tests.
+      module: install_prereq.py
+      abort-on-fail: true
+
+  - test:
+      name: cluster deployment
+      desc: Execute the cluster deployment workflow.
+      module: test_cephadm.py
+      config:
+        verify_cluster_health: true
+        steps:
+          - config:
+              command: bootstrap
+              service: cephadm
+              base_cmd_args:
+                verbose: true
+              args:
+                mon-ip: node1
+
+  - test:
+      name: Add host
+      desc: Add new host node with IP address
+      module: test_host.py
+      config:
+        command: add_hosts
+        service: host
+        args:
+          nodes:
+            - node1
+            - node2
+            - node3
+            - node4
+            - node5
+            - node6
+            - node7
+            - node9
+            - node10
+            - node11
+          attach_ip_address: true
+          labels: apply-all-labels
+      destroy-cluster: false
+      abort-on-fail: true
+
+  - test:
+      name: Mgr and Mon deployment
+      desc: Add Mgr and Mon daemons
+      module: test_cephadm.py
+      polarion-id: CEPH-83573746
+      abort-on-fail: true
+      config:
+        steps:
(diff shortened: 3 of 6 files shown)


ceph/rados/core_workflows.py
```diff
@@ -1264,3 +1264,73 @@ def get_osd_stat(self):
...
         return osd_stats
+
+    def run_pool_sanity_check(self):
+        """
+        Runs sanity on the pools after triggering scrub and deep-scrub on pools, waiting 600 Secs
+
+        This method is used to assess the health of Pools after any operation, where in a scrub and deep scrub is
+        triggered, and the method scans the cluster for few health warnings, if generated
+
+        Returns: True/Pass for Pass/Fail
```
harshkumarRH, line 1275: Is this supposed to be True/False for Pass/Fail?

ceph/rados/core_workflows.py
```diff
@@ -1264,3 +1264,73 @@ def get_osd_stat(self):
...
+        if not flag:
+            log.error(
+                "Health warning generated on cluster and not cleared post waiting of 600 seconds"
+            )
+            return False
+
+        log.info("Completed check on the cluster. Pass!")
+        return True
+
+    def get_osd_hosts(self):
```
harshkumarRH, line 1328: IMHO, later on we should plan to rename this module to `get_osd_hostnames(self)` because just going by the name, one would interpret the osd node objects to be returned instead of hostnames.
Also as we already have this info available with us, we can skip executing a ceph command to reduce execution time, please ignore if irrelevant
```
osd_nodes = ceph_cluster.get_nodes(role="osd")
return [node.hostname for node in osd_nodes]
```

harshkumarRH: On second thought I assume it is being done this way to avoid false failures if a host is actually an OSD host but does not have appropriate labels, so fetching the list from `ceph osd tree` would eliminate the possibility of such a host being missed.

tests/rados/test_pool_osd_recovery.py
```diff
@@ -0,0 +1,348 @@
...
+    )
+
+    log.info("\n\n---- Starting All workflows ----\n\n")
+    # rebooting one OSD and checking cluster health
+    log.info(
+        "---- Starting workflow ----\n---- 1. Reboot of single osd and health check ----"
+    )
+    for pool in pools:
+        pool_id = pool_obj.get_pool_id(pool_name=pool)
+        pgid = f"{pool_id}.0"
```
harshkumarRH, line 73: Is it necessary to fetch pool id, as we can get the pg acting set using pool name directly

tests/rados/test_pool_osd_recovery.py
```diff
@@ -0,0 +1,348 @@
...
+    cephadm.shell([rm_cmd])
+    time.sleep(5)
+
+    # Checking if the host still exists on the cluster
+    ls_cmd = "ceph orch host ls"
+    hosts = rados_obj.run_ceph_command(cmd=ls_cmd)
+    for host in hosts:
+        if host["hostname"] == rm_host.hostname:
+            log.error(f"Host : {rm_host.hostname} still present on the cluster")
+            raise Exception("Host not removed error")
```
harshkumarRH, line 333: Small suggestion, instead of traversing complete list of hosts on the cluster we can check out output of cmd: ` ceph orch host ls --host_pattern rm_host.hostname` to determine whether the host still exists

harshkumarRH:
Looks good to me

ceph/rados/core_workflows.py
```diff
@@ -1264,3 +1264,73 @@ def get_osd_stat(self):
...
+        Runs sanity on the pools after triggering scrub and deep-scrub on pools, waiting 600 Secs
+
+        This method is used to assess the health of Pools after any operation, where in a scrub and deep scrub is
+        triggered, and the method scans the cluster for few health warnings, if generated
+
+        Returns: True/Pass for Pass/Fail
+        """
+        self.run_scrub()
+        self.run_deep_scrub()
+        time.sleep(10)
```
SrinivasaBharath, line 1279: Instead of sleep check that scrub and deep-scrub is actually happening in the cluster.

tests/rados/test_pool_osd_recovery.py
```diff
@@ -0,0 +1,348 @@
...
+    method_should_succeed(rados_obj.run_pool_sanity_check)
+    log.info("----Completed workflows 1. Reboot of single osd and health check----")
+
+    log.info("---- Starting workflow ----\n---- 2. Rolling Reboot OSD hosts ----")
+    for nodes in osd_nodes:
+        log.info(f"Rebooting node : {nodes.hostname}")
+        nodes.exec_command(sudo=True, cmd="reboot", long_running=True)
+
+        # Sleeping for 10 seconds and starting verification.
+        time.sleep(10)
```
SrinivasaBharath, line 98: IMO below method waits for recovery to complete. can we remove time.sleep above?

conf/pacific/rados/13-node-cluster.yaml
```diff
@@ -0,0 +1,73 @@
+# Rados tier-4 test configuration
```
SrinivasaBharath, line 1: Any reason why we are targeting pacific instead of Quincy?

ceph/rados/core_workflows.py
pdhiran: Hello Harsh. Great that you noticed. And there's a pretty good reasoning for the same.

The objects that are fetched via : ceph_cluster.get_nodes(role="osd") , The nodes are fetched from the conf file that we provide, not fetching it from the cluster deployed.

In most cases, I bet this would work as expected, without any issues.

But in our scenario, We are not adding all the nodes in the conf file into the cluster. So I need to fetch the host names from the cluster itself. If I call the get_nodes(), it would return me nodes that are not yet added to cluster also, Making verification pretty difficult.

tests/rados/test_pool_osd_recovery.py
pdhiran: In this case, If the node isn't present, the o/p won't be in json, which will be bit more complicated.

If the above comment is a requirement, I will update the code. Let me know.

tests/rados/test_pool_osd_recovery.py
pdhiran: The above method does not wait for recovery.

ceph/rados/core_workflows.py
pdhiran: Completion of scrub and checking the results is not part of test workflow. I believe that is being covered as part of other tests. 

If the check is a necessity, would incorporate the changes. Let me know.

conf/pacific/rados/13-node-cluster.yaml
pdhiran: no reason. the suite needs to be added to both, and I will add the same in the next PR.

This PR is alrady huge with around ~700 lines of code. Did not want to make it bigger with new suite and conf files for quincy.

Let me know. If necessary to add, will add in the same PR.

ceph/rados/core_workflows.py
pdhiran: Updated the return Values.

tests/rados/test_pool_osd_recovery.py
pdhiran: Good Catch. Since i'm anyway always passing pool_id.0 always, It would return me the same results.

Updated the code to pass pool name directly and get the OSD list.

tests/rados/test_pool_osd_recovery.py
```diff
@@ -0,0 +1,348 @@
...
+            )
+
+    # Checking cluster health after the tests
+    method_should_succeed(rados_obj.run_pool_sanity_check)
+    log.info(
+        "--- Completed workflows 4. restart OSD daemons belonging to single pg ---"
+    )
+
+    log.info("---- Starting workflow ----\n---- 5. Removal and addition of OSD daemons")
+    for pool in pools:
```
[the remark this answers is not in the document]
pdhiran: Updated the code, now it does not warrant a method to fetch the primary OSD.

tests/rados/test_pool_osd_recovery.py
harshkumarRH: The o/p would be an empty list if a host does not exist, however, this is very trivial and works either way, resolving this thread.

tests/rados/test_pool_osd_recovery.py
```diff
@@ -0,0 +1,337 @@
...
+
+    log.info("Completed reboot of all the OSD hosts, Checking cluster health status")
+    # Checking cluster health after the test
+    method_should_succeed(rados_obj.run_pool_sanity_check)
+    log.info(
+        "---- Completed workflows 2. Rolling Reboot OSD hosts and health check ----"
+    )
+
+    log.info(
+        "---- Starting workflow ----\n---- 3. Stopping and starting OSD daemons ----"
```
SrinivasaBharath, line 107: As per the test case mentioned, both scenario performs the same scenarios. Reboot of all the Primary PG OSD followed by the remaining acting set OSD. Step 3 & 4 in the TC CEPH-11032. Please modify the TC if required.

tests/rados/test_pool_osd_recovery.py
```diff
@@ -0,0 +1,337 @@
...
+
+    log.info("Completed start and stop for all targeted OSDs. Checking cluster health")
+    # Checking cluster health after the tests
+    method_should_succeed(rados_obj.run_pool_sanity_check)
+    log.info("--- Completed workflows 3. Stopping and starting OSD daemons ---")
+
+    log.info(
+        "---- Starting workflow ----\n---- 4. restart all OSD daemons belonging to single pg----"
+    )
+    for pool in pools:
```
SrinivasaBharath, line 141: This scenario and  scenario 1. Reboot of single osd and health check are looks similar. Any additional verification are we doing here?

tests/rados/test_pool_osd_recovery.py
pdhiran: Once the nodes are rebooted, I have observed that it takes some time to start balancing, identifying that nodes have gone down.

This 10 seconds helps achieve that. So cannot remove the 10 seconds. Can be reduced, but I have seen that 10 secs should cover all.

tests/rados/test_pool_osd_recovery.py
pdhiran: No additional checks. The above scenario asked for only primary, then later all the PGs. The same has been followed.

tests/rados/test_pool_osd_recovery.py
pdhiran: Reboot is different from start and stop. 

In reboot OSDs are generally not marked out. In stop scenario, we are waiting for the data in the down OSD to get copied into other OSD.

In reboot, data is not usually copied, but the delta of the data written during downtime is copied.

Of course this is not verified explicitly, but rather we wait for PGs to get clean with a OSD down.

Let me know what is the same scenario here, or if something needs modification

suites/pacific/rados/tier-4_rados_test-pool-osd-recovery.yaml
```diff
@@ -0,0 +1,208 @@
...
+      abort-on-fail: false
+
+  - test:
+      name: Failure recovery on Replicated & EC
+      module: test_pool_osd_recovery.py
+      polarion-id: CEPH-11032
+      config:
+        pool_configs:
+          - type: replicated
+            conf: sample-pool-2
```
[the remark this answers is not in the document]
pdhiran: Updated test params!

tests/rados/test_pool_osd_recovery.py
pdhiran: Hey Bharath, these are the scenarios :
1. Reboot of single osd (Primary PG OSD) and health check.
2. Rolling Reboot OSD hosts ( All OSDs belonging to 1 sample PG in the pool ).
3. Stopping and starting OSD daemons (Primary PG OSD) .
4. restart all OSD daemons belonging to single pg ( All OSDs belonging to 1 sample PG in the pool ) -> Difference from scenario 2 is that entire host is rebooted, here in 4, only the OSD service is rebooted.
5. Removal and addition of OSD daemons.
6. Removal and addition of OSD Hosts.

pdhiran:
Pass logs : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-PVRKG5

tests/rados/test_pool_osd_recovery.py
SrinivasaBharath: LGTM

SrinivasaBharath (approved):

PR #2407, opened 2023-03-29
Subscription registration using CDN account.

sunilkumarn417:
Since stage account has registration issues. Pointing subscription manager registration to CDN account.


manasagowri (approved):
Changes LGTM

PR #2443, opened 2023-04-06
[CEPH-83574719] Validate pwl without exclusive-lock

sunilkumarn417:
**CEPH-83574719** - exclusive lock disabled, check cache image status in SSD mode
- Create pool and image.
- Disable `exclusive-lock feature` on the image.
- Configure PWL on cache client.
- Run FIO and validate the cache file not created & PWL won't work there.

**Test results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-WBMHVZ/PWL_cache_creation_with_exclusive_lock_0.log

### DIFF
--- a/suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
+++ b/suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
@@ -287,3 +287,28 @@ tests:
           desc: RBD Persistent Cache cache path validation image level
           module: test_rbd_persistent_write_back_cache.py
           name: PWL cache path validation at image level
+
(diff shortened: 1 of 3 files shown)


sunilangadi2 (approved):
LGTM, Please cover these newly added tests in Pacific as well.

manasagowri (changes requested):
Please add a new test suite for pacific for these tests.
Also add the tests into pipeline metadata so they can be run as part of the CI pipeline.

manasagowri (approved):
Since this is a baremetal test case, will plan to add it to pipeline in a different PR after discussing with Amar.

PR #2457, opened 2023-04-11
[Fix] Allow client node to add additional repo

vamahaja:
In `install_prereq.py`, skips client node to add repo passed through run.py parameter `--add-repo`. This causes issue while adding RHEL external repos.

Remove check for client node to allow additional repo.


AmarnatReddy (approved):
LGTM

PR #2465, opened 2023-04-13
Temporary execution switch onto centos-8 builds.

sunilkumarn417:
Since Ceph builds on centos-9 are not generated, moving back the execution to run on centos-8 builds.
In future we should revert back to latest builds with latest os version.


anrao19 (approved):

tintumathew10 (approved):
LGTM

PR #2499, opened 2023-04-20
Integrate CEPH-83575142 to cephCI

TejasC88:
On a bucket granular sync policy at bucket level , we have added filtering objects through prefix and tag.

Pass log :
http://magna002.ceph.redhat.com/ceph-qe-logs/tejas/cephci-run-ZS9433/

### DIFF
--- a/suites/pacific/rgw/tier-1_rgw_ms-sync-policy-from-primary.yaml
+++ b/suites/pacific/rgw/tier-1_rgw_ms-sync-policy-from-primary.yaml
@@ -285,6 +285,19 @@ tests:
       name: create non-tenanted user
 
   - test:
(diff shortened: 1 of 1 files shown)


viduship (approved):

hmaheswa (approved):
in the logs, objects are not written to io_info file because objects are not uploaded using resource_op in upload_object_with_tagging [here](https://github.com/red-hat-storage/ceph-qe-scripts/blob/master/rgw/v2/tests/s3_swift/reusable.py#L330)
that should be fixed in the next PR

anrao19 (approved):

ckulal (approved):

PR #2518, opened 2023-04-25
Adding clone resize to luks encryption tests

manasagowri:
This PR is to fix the issues faced while testing out the following combinations in luks encryption

1) Parent not encrypted, clone is luks1/luks2 encrypted
2) Parent luks1 encrypted, clone luks2 encrypted

Scenarios and corresponding steps
    1) Parent and clone are luks1 encrypted
        Resize parent at beginning to compensate for header size,
        clone resize not required
    2) Parent luks1/luks2 and clone is not encrypted
        Resize parent at beginning to compensate for header size,
        clone resize not required since it has no header
    3) Parent luks2 and clone luks1
        Resize parent at beginning to compensate for header size,
        resize clone with --allow-shrink since luks2 header is bigger than luks1
    4) Parent not encrypted and clone is luks1/luks2
        Resize not required before clone since parent is not encrypted, resize
        clone with --allow-shrink post cloning to compensate for header
    5) Parent luks1 and clone luks2
        Resize parent before clone, resize both parent and
        clone with --allow-shrink post cloning

    Justification:
    Since LUKS2 header is usually bigger than LUKS1 header, the rbd resize command at
    the beginning temporarily grows the parent image to reserve some extra space in the
    parent snapshot and consequently the cloned image. This is necessary to make all parent
    data accessible in the cloned image. The rbd resize command at the end shrinks the
    parent image back to its original size and does not impact the parent snapshot and
    the cloned image to get rid of the unused reserved space

    The same applies to creating a formatted clone of an unformatted image,
    since an unformatted image does not have a header at all.

Reference: https://access.redhat.com/documentation/en-us/red_hat_ceph_storage/5/html/block_device_guide/image-encryption 

Success logs : 
5.3 - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-m6hfb/
6.1 - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-xmp7r/

### DIFF
--- a/tests/rbd/rbd_utils.py
+++ b/tests/rbd/rbd_utils.py
@@ -473,6 +473,14 @@ class Rbd:
             image_name: image of the pool
             size: new size for the image
             kw: any other extra arguments including but not limited to encryption-passphrase-file
+                To perform resize on encrypted images pass the encryption-passphrase-file info as given below
+                kw = {
+                    "passphrase":[
+                        "encryption-passphrase-file": <path to the file where clone passphrase is
+                                                        stored if image is clone>,
+                        "encryption-passphrase-file": <path to the file where parent passphrase is stored>
+                    ]
+                }
         Returns:
 
         """
@@ -481,6 +489,8 @@ class Rbd:
             for passphrase in kw.get("passphrase"):
                 for key, value in passphrase.items():
                     cmd += f" --{key} {value}"
+            if kw.get("allow_shrink"):
+                cmd += " --allow-shrink"
             return self.exec_cmd(cmd=cmd, all=kw.get("all", False))
 
         if kw.get("flag") == "increase":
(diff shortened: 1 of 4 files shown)


sunilangadi2 (dismissed):
LGTM

tests/rbd/rbd_utils.py
```diff
@@ -481,6 +481,8 @@ def image_resize(self, pool_name, image_name, size, **kw):
             for passphrase in kw.get("passphrase"):
                 for key, value in passphrase.items():
                     cmd += f" --{key} {value}"
+            if kw.get("allow_shrink"):
```
rkhambad, line 484: can you rebase the branch, there are some more changes in same method in master branch , Please check and do changes accordingly

rkhambad (changes requested):

tests/rbd/test_rbd_encryption.py
rkhambad, line 1: [optional] Steps says first flatten the clone and then check for data integrity, but we are checking for data integrity first and then flattening the clone.
It might not affect the functionality wise, just flow is not matching, either of both can be changed.

tests/rbd/rbd_utils.py
manasagowri: Done.

tests/rbd/test_rbd_encryption.py
manasagowri: I think the functionality is not affected, so I will change it in the test steps itself. PR can remain as it is.

rkhambad: I did not observe the necessary changes in the steps.
11. Flatten the clone
12. Verify data integrity between clone image and the original image

rkhambad (dismissed):
Approving with optional changes

tests/rbd/rbd_utils.py
```diff
@@ -481,6 +481,8 @@ def image_resize(self, pool_name, image_name, size, **kw):
             for passphrase in kw.get("passphrase"):
                 for key, value in passphrase.items():
```
VasishtaShastry, line 482: [optional]
Sorry for adding this now

Can we add info on passphrase in method doc ?
Also, if you can give me context on relationship with "encryption-passphrase-file" from method doc, that would be helpful - is it specific to encrypted image ?

tests/rbd/rbd_utils.py
VasishtaShastry, line 484: Sorry if I'm missing something, in line 491 we are adding --allow-shrink
Can't we make use of it (optimise) ?

tests/rbd/rbd_utils.py
```diff
@@ -1158,3 +1160,150 @@ def verify_image_size(rbd, pool, image, size):
...
+        Resize parent before clone, resize both parent and
+        clone with --allow-shrink post cloning
+
+    Justification:
+    Since LUKS2 header is usually bigger than LUKS1 header, the rbd resize command at
+    the beginning temporarily grows the parent image to reserve some extra space in the
+    parent snapshot and consequently the cloned image. This is necessary to make all parent
+    data accessible in the cloned image. The rbd resize command at the end shrinks the
+    parent image back to its original size and does not impact the parent snapshot and
+    the cloned image to get rid of the unused reserved space
```
VasishtaShastry, line 1192: [Optional]

1. Should we be moving some wrappers out of rbd_utils ? (may be something like rbd_encryption_utils) for this method to keep rbd_utils more maintainable ?
2. [minor] Justification part could be code comment or some link large method docs are okay ?

tests/rbd/rbd_utils.py
```diff
@@ -1158,3 +1160,150 @@ def verify_image_size(rbd, pool, image, size):
...
+                "parent_encryption_type": "luks1",
+                "parent_spec": "pool/image",
+                "parent_passphrase": "passphrase.bin",
+                "clone_encryption_type": "luks2",
+                "clone_spec": "pool/clone",
+                "clone_passphrase": "clone_passphrase.bin",
+                "size": <image_size>
+            }
+    """
+    log.info(f"Performing resize operation {resize_sequence}")
```
VasishtaShastry, line 1210: [Optional] [Minor]
Can we provide context on resizing + encryption?
This applies to first line of the method doc also

tests/rbd/test_rbd_encryption.py
manasagowri: Done the changes now.

tests/rbd/rbd_utils.py
manasagowri: encryption-passphrase-file is a parameter passed to rbd resize command when the image in consideration is encrypted.
What more context do you want Vasi? Could you please elaborate?

I have also added some method doc with example to image_resize method. So please let me know if you need any more info.

tests/rbd/rbd_utils.py
manasagowri: So there are few customized command arguments I am expecting when it is an encrypted image, for example "all", and the passphrase in itself. And for encrypted images I would like to pass --allow-shrink only when it is specified by user. 

Because of these reasons I would like to retain the code as is. Also I am returning the value from line 493 itself when image is encrypted so it won't even reach line 498. 

Please let me know if this explanation suffices or if you have any better way of optimisation.

tests/rbd/rbd_utils.py
manasagowri: 1. We could do it, but for now I would like to keep it here, in my other draft PR for tier-3 scenarios I am already working on splitting relevant workflows into different python files, we could probably take this optimization up as part of that.
2. Is there any problem with large method docs? I am not aware, let me know if I need to remove justification I can.

tests/rbd/rbd_utils.py
manasagowri: Please let me know if the current doc and log looks okay?

tests/rbd/rbd_utils.py
VasishtaShastry: Got it
Thank you

VasishtaShastry (approved):

PR #2553, opened 2023-05-10
[Lib] Add utility to cleanup cluster

vamahaja:
Add new utility under `cephci/cleanup.py` which cleanup cluster from `Openstack` with prefix which supports below parameters -
```
    Usage:
        cephci/cleanup.py --log-level <LOG> --cloud-type <CLOUD> --cloud-creds <CRED> --pattern <STR>
        cephci/cleanup.py --help

    Options:
        -h --help                  Help
        -l --log-level <LOG>       Log level for log utility
        -t --cloud-type <CLOUD>    Cloud type [openstack|ibmc|baremetal]
        -c --cloud-creds <CRED>    Credentials for cloud
        -p --pattern <STR>         Resource name pattern
```


cli/utilities/waiter.py
```diff
@@ -0,0 +1,32 @@
...
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        if self._start is None:
+            self._start = time.time()
+        if time.time() - self._start > self.timeout:
+            self.expired = True
+            raise StopIteration()
+        if self._attempt != 0:
```
pranavprakash20, line 29: are we keeping track of the attempts performed? If not, do we need this variable?

cli/exceptions.py
```diff
@@ -0,0 +1,35 @@
+class ConfigError(Exception):
```
pranavprakash20, line 1: Good approach +1. But can we have a general class for all the exceptions? Or the idea is to club only those exceptions which falls under the scope of the module it is placed?

cli/cluster/node.py
```diff
@@ -0,0 +1,85 @@
...
+        elif cloud == "ibmc":
+            pass
+        else:
+            raise ConfigError(f"Unsupported cloud provider '{cloud}'")
+
+    @staticmethod
+    def nodes(pattern, cloud="openstack", **config):
+        """Get nodes with pattern"""
+        _cloud = Node._get_cloud(cloud, **config)
+        return _cloud.get_nodes_by_pattern(pattern)
```
pranavprakash20, line 40: can't we replace this with return self._cloud.get_nodes_by_pattern(pattern) ?

cli/utilities/waiter.py
vamahaja: yes, we can remove this variable, not required here. Will remove it.

vamahaja: @pranavprakash20 `_attempt` variable is used for avoiding sleep at first attempt. If we remove this variable at very first attempt it will wait for given `interval`.

pranavprakash20 (approved):
LGTM

pranavprakash20 (dismissed):
LGTM

pranavprakash20 (approved):
LGTM

PR #2561, opened 2023-05-15
[Lib] Add library to read configs

vamahaja:
Add new format for `cephci.yaml` config which will have below major sections and library to validate and read the configs -

1. Credentials
2. Repositories
3. Packages

### DIFF
--- a/cephci/utils/configs.py
+++ b/cephci/utils/configs.py
@@ -1,0 +1,172 @@
+import yaml
+
+from cli.exceptions import ConfigError
+from utility.log import Log
+
+log = Log(__name__)
+
+DEFAULT_CONFIG_PATH = "~/.cephci.yaml"
+CONFIG = None
+
+
+def get_configs(config=DEFAULT_CONFIG_PATH):
+    """Read configurations from yaml
+
+    Args:
+        config (str): Config file path
+    """
+    global CONFIG
+    if CONFIG:
+        return CONFIG
+
+    log.info(f"Loading config file - {config}")
+    with open(config, "r") as _stream:
+        try:
+            CONFIG = yaml.safe_load(_stream)
+        except yaml.YAMLError:
+            raise ConfigError(f"Invalid configuration file '{config}'")
+
+
+def _get_credentials():
+    """Get credentials from config"""
+    if not CONFIG:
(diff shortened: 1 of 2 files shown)


cephci/utils/configs.py
```diff
@@ -0,0 +1,99 @@
...
+    if not CONFIG:
+        raise ConfigError("Configuration is not passed")
+
+    try:
+        return CONFIG["credentials"]
+    except KeyError:
+        raise ConfigError("Credentials configurations are missing from config")
+
+
+def get_subscription_credentials(server):
```
pranavprakash20, line 31: one suggestion, the server variable here will have the value `cdn`, `stage` etc right? In that case, can we replace the server with subscription_type or something similar? Please ignore if not valid

cephci/utils/configs.py
```diff
@@ -0,0 +1,99 @@
...
+        _dict["password"] = _server["password"]
+        _dict["timeout"] = int(_server.get("timeout", 60))
+        _dict["retry"] = int(_server.get("retry", 6))
+    except KeyError:
+        raise ConfigError(f"Insufficient config for '{server}' in subscription")
+
+    return _dict
+
+
+def get_registry_credentials(server, build):
```
pranavprakash20, line 48: same here, the server variable has ibm or rh as the value, so can it be renamed as well?

cephci/utils/configs.py
```diff
@@ -0,0 +1,99 @@
+import yaml
```
pranavprakash20, line 1: also, can we have doc strings for these methods? Mentioning the params, possible values etc

vamahaja: Done

cephci/utils/configs.py
vamahaja: `cdn`, `stage`  are based on the `serverurl` that's why `server` variable is used here.

cephci/utils/configs.py
vamahaja: here server will have the values like `cdn`, `stage`, `cp` etc and build is `rh` & `ibm`

pranavprakash20 (approved):
LGTM

PR #2578, opened 2023-05-19
[Fib] Update OSP libraries to support multiple environments

vamahaja:


pranavprakash20 (dismissed):
LGTM

vamahaja:
Testrun - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-2mu2l/

compute/openstack.py
```diff
@@ -78,7 +78,7 @@ def get_openstack_driver(
 class CephVMNodeV2:
     """Represent the VMNode required for cephci."""
 
-    default_network_names = [
+    default_rhosd_network_names = [
         "provider_net_cci_12",
         "provider_net_cci_11",
         "provider_net_cci_9",
```
AmarnatReddy, line 84: Hi Vaibhav,
Is there any restriction to use only this network pools.
if not can we get this dynamically using driver.ex_get_network_pool()

vamahaja: @AmarnatReddy if that works, then we can get it dynamically

vamahaja: @AmarnatReddy we are redesigning cloud libraries and this includes these openstack operations. This PR is to unblock us for now to use both the OSP projects, Will consider this enhancement as part of redesigning.

pranavprakash20 (approved):
lgtm

AmarnatReddy (approved):
LGTM

PR #2581, opened 2023-05-22
CEPH-11032: Modification of test module to wait for host reboot

harshkumarRH:
Tier-2 test case [CEPH-11032](https://polarion.engineering.redhat.com/polarion/#/project/CEPH/workitem?id=CEPH-11032) which covers the failure recovery scenario has been observed to be failing due to an edge condition where OSDs to be restarted are already down as the OSD hosts were rebooted in the previous workflow and are yet to come up.

Failure log:
RHCS 5.3 - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-JNSM1N/

Pass log:
RHCS 5.3 - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-S7UN52
RHCS 6.1 - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-MDD3KZ

### DIFF
--- a/ceph/rados/core_workflows.py
+++ b/ceph/rados/core_workflows.py
@@ -1487,3 +1487,17 @@ class RadosOrchestrator:
         for service in orch_ls_op:
             service_name_ls.append(service["service_name"])
         return service_name_ls
+
+    def check_host_status(self, hostname) -> bool:
+        """
+        Checks the status of host(offline or online) using
+        ceph orch host ls and return boolean
+        Args:
+            hostname: hostname of host to be checked
+        Returns:
+            (bool) True -> online | False -> offline
+        """
+        host_cmd = f"ceph orch host ls --host_pattern {hostname} -f json"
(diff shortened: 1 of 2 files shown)


ceph/rados/core_workflows.py
```diff
@@ -1487,3 +1487,17 @@ def list_orch_services(self, service_type=None):
         for service in orch_ls_op:
             service_name_ls.append(service["service_name"])
         return service_name_ls
+
+    def check_host_status(self, hostname) -> bool:
```
pdhiran, line 1491: if the host being rebooted is the node, where Cephadm shell resides, this would fail.

Do we want to configure this command to be run on Client? Client reboots are seldom done. So should be safe.

Let me know what you think.

harshkumarRH: Thanks for pointing that out, I have changed it to client node. Please review, RHCS 5.3 logs are with the latest changes.

pdhiran (approved):
Looks good!

PR #2593, opened 2023-05-25
fixing the upstream rados suite

pdhiran:
In the test suite, we have RBD sanity tests performed as well. recently, test coverage was added for the rbd test to include EC pools. Since Stretch Cluster does not support creation of EC pools, the tests were failing.

Updated the test to include RE pools only.

Pass logs : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-W7TASA


neha-gangadhar:
Looks good.

neha-gangadhar (approved):

PR #2597, opened 2023-05-26
Quincy tests: Adding missing tests from pacific to Quincy suites.

pdhiran:
Quincy tests: Adding missing tests from pacific to Quincy suites.

Pass logs : 
tier-4_rados_test-pool-osd-recovery.yaml : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-279L7W 
tier-2_rados_test-pool-functionalities.yaml: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-6W4ZI6 
Bugzilla for the omap issue : https://bugzilla.redhat.com/show_bug.cgi?id=2210278

PR #2641, opened 2023-06-14
Optimize wait time required for OMAP entries to get reflected | Temporarily exclude pool data migration tests

harshkumarRH:
### Part-1 : OMAP entries wait time optimization
Reference Jira story - [RHCEPHQE-9423](https://issues.redhat.com/browse/RHCEPHQE-9423)
Originally OMAP entries used to get reflected in the cluster statistic in these three cases:
- OMAP entries made on a pool show up automatically in the ceph df stats if enough number of omaps are written to the pool (observed previously and is still true)
Explanation for this is available here and is still true - https://bugzilla.redhat.com/show_bug.cgi?id=2139021#c36
- Ideally, OMAP stats should show up if deep-scrub is performed on a pool
- Even without deep-scrub, if OSDs which are part of the pool's acting pg set were restarted, OMAP entries were observed to recognized and displayed in the ceph df stats.

Our automation was following case number 3 as the standard workflow to verify OMAP entries on a cluster, now, lately case 2 and case 3 were no longer working and the only working case was found to be case 1
Upon further trial and error it had been observed that now combination of case 2 and case 3 was required for OMAP data to get reflected in the cluster stats.

With the PR #2597, we modified the verification of OMAP to include a combination of case 2 and case 3, for this we performed deep-scrub, waited for deep-scrub to complete and then restart the concerned OSDs and further checked for existence of OMAPs
This change in workflow resulted in test execution time increasing from 50 mins to 4 hours.
Old logs - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-CNPDR5/

With the current PR, workflow has been changed to start deep-scrub on the concerned pool and perform OSD restart immediately after, instead of waiting for deep-scrub to complete. This whole process has been encapsulated in a smart wait _while loop_ which looks for OMAP entries to become available and restarts the OSDs otherwise.
As soon as OMAP entries get reflected, we are no longer waiting for deep-scrub to get completed, even though it is still running in the background. Rest of the test workflow remains the same.

Test modules modified:
- ceph/rados/pool_workflows.py
- ceph/rados/core_workflows.py
- tests/rados/test_omap_entries.py

Test suites modified:
- suites/pacific/rados/tier-2_rados_test-pool-functionalities.yaml
- suites/quincy/rados/tier-2_rados_test-pool-functionalities.yaml

Logs:
RHCS 6.1 (mClock) - http://magna002.ceph.redhat.com/ceph-qe-logs/harsh/migrate-data-pools-mclock/
RHCS 6.1 (WPQ) - http://magna002.ceph.redhat.com/ceph-qe-logs/harsh/migrate-data-pools-wpq/

As seen from the above logs, the test execution time for the test case [Omap creations on objects](https://polarion.engineering.redhat.com/polarion/#/project/CEPH/workitem?id=CEPH-83571702) is back to ~50 mins.

Execution of compete test suite _**tier-2_rados_test-pool-functionalities.yaml**_ is in-progress and logs will be attached shortly

### Part-2 : Migration of data between pools takes huge amount of time
The pool functionalities test suite contains 4 test cases verifying the different scenarios for migration of data between pools, out these 4, 3 were observed to be taking 3-4 hours each for completion.
Initial review indicated that this degradation could be due to the introduction of mClock Scheduler as the default OSD OP Queue which wasn't part of RHCS 6.0 where these test were working fine.
However, it is pretty evident that changing the OSD OP Queue back to WPQ on a 6.1 build had no impact on this degradation and test execution time was still 2-4 hours each. The results for reference are available in the pass logs attached above.

As RCA for this degradation is still under way, the 3 concerned test cases have been commented out for the time being.
The test cases will be included again once a automation fix and / or code fix is available depending on RCA

### DIFF
--- a/ceph/rados/pool_workflows.py
+++ b/ceph/rados/pool_workflows.py
@@ -91,6 +91,27 @@ class PoolFunctions:
 
         Returns: True -> pass, False -> fail
         """
+
+        def check_omap_entries() -> bool:
+            pool_stats = self.rados_obj.get_cephdf_stats(
+                pool_name=pool_name, detail=True
+            )
+            if pool_stats["stats"]["stored_omap"] <= 0:
+                # OMAP entries are not readily available,
+                # restarting primary osd as metadata does not get updated automatically.
+                # TBD: bug raised for the issue: https://bugzilla.redhat.com/show_bug.cgi?id=2210278
+                primary_osd = self.rados_obj.get_pg_acting_set(pool_name=pool_name)[0]
+                self.rados_obj.change_osd_state(action="restart", target=primary_osd)
+                pool_stats = self.rados_obj.get_cephdf_stats(
+                    pool_name=pool_name, detail=True
+                )
+                if pool_stats["stats"]["stored_omap"] <= 0:
+                    log.error(
+                        f"No omap entries available for the pool {pool_name} even after restarting primary OSD"
+                    )
+                    return False
+            return True
+
         # Getting the client node to perform the operations
         client_node = self.rados_obj.ceph_cluster.get_nodes(role="client")[0]
         obj_start = kwargs.get("obj_start", 0)
@@ -134,12 +155,14 @@ class PoolFunctions:
             deep_scrub_status = self.scrub_obj.verify_scrub_deepscrub(
                 before_deep_scrub_log, after_deep_scrub_log, "deepscrub"
             )
+
             if deep_scrub_status:
-                log.debug(
-                    "deep scrubbing has not been completed on the pool"
-                    "Sleeping for 10 seconds and checking again"
-                )
-                time.sleep(10)
+                log.info("deep scrubbing has not been completed on the pool")
+                if check_omap_entries():
+                    log.info("OMAP entries were available post OSD restart")
+                    break
+                log.info("Sleeping for 120 seconds and checking again")
+                time.sleep(120)
                 flag = 1
                 continue
             flag = 0
@@ -152,42 +175,19 @@ class PoolFunctions:
             )
             return False
 
-        pool_stats = self.rados_obj.run_ceph_command(cmd="ceph df detail")["pools"]
-        for detail in pool_stats:
-            if detail["name"] == pool_name:
-                pool_1_stats = detail["stats"]
-                if pool_1_stats["stored_omap"] <= 0:
-                    log.error(
-                        "No omap entries written into pool, Performing the workaround of restarting primary OSD"
-                    )
-                    # restarting primary osd as metadata does not get updated automatically.
-                    # TBD: bug raised for the issue:
-                    primary_osd = self.rados_obj.get_pg_acting_set(pool_name=pool_name)[
-                        0
-                    ]
-                    self.rados_obj.change_osd_state(
-                        action="restart", target=primary_osd
-                    )
-                else:
-                    log.info(
-                        f"Wrote {pool_1_stats['stored_omap']} bytes of omap data on the pool."
-                    )
-                    return True
-        time.sleep(10)  # blind sleep to let omap metadata get updated
         log.debug("Checking the amount of omap entries created on the pool")
-        pool_stats = self.rados_obj.run_ceph_command(cmd="ceph df detail")["pools"]
-        for detail in pool_stats:
-            if detail["name"] == pool_name:
-                pool_1_stats = detail["stats"]
-                total_omap_data = pool_1_stats["omap_bytes_used"]
-                omap_data = pool_1_stats["stored_omap"]
-                break
-        if omap_data <= 0:
-            log.error("No omap entries written into pool")
+        if not check_omap_entries():
+            log.error(
+                "No omap entries available for the pool after complete deep-scrub and OSD restart"
+            )
             return False
+        pool_stats = self.rados_obj.get_cephdf_stats(pool_name=pool_name, detail=True)
         log.info(
-            f"Wrote {omap_data} bytes of omap data on the pool."
-            f"Total stored omap data on pool : {total_omap_data}"
+            f"Wrote {pool_stats['stats']['stored_omap']} bytes of omap data on the pool."
+        )
+        log.info(
+            f"Wrote {pool_stats['stats']['stored_omap']} bytes of omap data on the pool."
+            f"Total stored omap data on pool : {pool_stats['stats']['omap_bytes_used']}"
         )
         return True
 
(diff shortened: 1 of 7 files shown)


suites/pacific/rados/tier-2_rados_test-pool-functionalities.yaml
```diff
@@ -255,16 +244,18 @@ tests:
...
+#      module: test_data_migration_bw_pools.py
+#      polarion-id: CEPH-83574768
+#      config:
+#        pool-1-type: replicated
+#        pool-2-type: replicated
+#        pool-1-conf: sample-pool-1
+#        pool-2-conf: sample-pool-2
+#      desc: Migrating data between different pools. Scenario-1. RE -> RE
 
   - test:
```
pdhiran, line 260: Even this can be commented. The tests are passing. But it's slow. Let's add all tests post RCA itself.

suites/quincy/rados/tier-2_rados_test-pool-functionalities.yaml
```diff
@@ -252,16 +252,18 @@ tests:
...
+#      module: test_data_migration_bw_pools.py
+#      polarion-id: CEPH-83574768
+#      config:
+#        pool-1-type: replicated
+#        pool-2-type: replicated
+#        pool-1-conf: sample-pool-1
+#        pool-2-conf: sample-pool-2
+#      desc: Migrating data between different pools. Scenario-1. RE -> RE
 
   - test:
```
pdhiran, line 268: Same as above.

ceph/rados/pool_workflows.py
```diff
@@ -134,12 +150,14 @@ def fill_omap_entries(self, pool_name, **kwargs):
...
             )
+
             if deep_scrub_status:
-                log.debug(
-                    "deep scrubbing has not been completed on the pool"
-                    "Sleeping for 10 seconds and checking again"
-                )
-                time.sleep(10)
+                log.info("deep scrubbing has not been completed on the pool")
+                if check_omap_entires():
```
pdhiran, line 156: Is it intended that every time we call ` check_omap_entires `, the primary OSD should be restarted?

pdhiran: Asking as if the Omap was not found, we sleep and continue. So the OSD could be restarted multiple times during the duration of sleep here.

pdhiran, line 156: If omap entries are found here, we are breaking out of the loop. Then again , line : 174 check seems redundant. We are rebooting the primary again post confirmation. can be removed.

tests/rados/test_omap_entries.py
```diff
@@ -25,82 +25,81 @@ def run(ceph_cluster, **kw):
...
 
-                # Waiting for cluster to get clean state after OSD stopped
-                if not wait_for_clean_pg_sets(rados_obj):
-                    log.error("PG's in cluster are not active + Clean state.. ")
-                    raise Exception("PG's in cluster are not active + Clean state.. ")
-                rados_obj.change_osd_state(action="restart", target=osd_id)
+            if omap_config["large_warn"]:
+                # Fetching the current acting set for the pool
+                acting_set = rados_obj.get_pg_acting_set(pool_name=pool_name)
+                rados_obj.change_recover_threads(config={}, action="set")
```
pdhiran, line 72: This might only be valid in 5.x. Need to update logic for this in another PR.

pdhiran (changes requested):
Looks good with minor changes. Additional reboot of primary OSD can be removed as well.

suites/pacific/rados/tier-2_rados_test-pool-functionalities.yaml
harshkumarRH: All the tests are passing, the other 3 are slow, this one isn't slow, so was not removed. However, will comment it out as discussed

neha-gangadhar:
could you please add the suite "tier-2_rados_test_omap.yaml" to metadata files as part of this PR ?

suites/pacific/rados/tier-2_rados_test-pool-functionalities.yaml
harshkumarRH: Correction: This test is also slow now
`Migrate data bw pools.           Migrating data between different pools. Scenario-2. RE -> EC   2:10:02.742211                   Pass`
Disabling it for now

tests/rados/test_omap_entries.py
pdhiran: Yes, Correct. Let's have a backlog task for this as well.

ceph/rados/pool_workflows.py
harshkumarRH: Modified to logic to first check if omap entries are available, and restart primary OSDs only in case of non-availability
Yes, agreed, Primary OSD will be restarted multiple times, this is because we do not know at which point deep-scrubbing actually reads and recognizes the OMAPs

harshkumarRH: The check at line 174 is not entirely redundant as it takes care of this edge case:
Consider that deep-scrub got completed, and while it was in progress, we restarted the primary OSD multiple times but OMAP never showed up, so now we will be out of the while loop which checks for completion of deep-scrub and also OMAP entries were never found, so this check just covers this worst possible edge case

In a happy path workflow, yes this is redundant

So I would request we let it be there, it won't do any harm as the code has been modified to ensure OSD is restarted only if OMAPs are not available

harshkumarRH:
Resolved conflicts and addressed PR comments.
Adding DNM tag as pass logs are yet to be updated

suites/pacific/rados/tier-2_rados_test-pool-functionalities.yaml
```diff
@@ -255,49 +244,53 @@ tests:
...
-      name: Migrate data bw pools.
-      module: test_data_migration_bw_pools.py
-      polarion-id: CEPH-83574768
-      config:
-        pool-1-type: replicated
-        pool-2-type: erasure
-        pool-1-conf: sample-pool-1
-        pool-2-conf: sample-pool-3
-      desc: Migrating data between different pools. Scenario-2. RE -> EC
+# Following tests has been commented till RCA is complete for massive increase
```
pdhiran, line 247: Added backlog task : RHCEPHQE-9494, to track the changes required here.

harshkumarRH:
@neha-gangadhar Metadata files have been updated

harshkumarRH:
Pass logs on RHCS 5.3 - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-NSJ4EA
Removing DNM tag
@pdhiran please review

ceph/rados/pool_workflows.py
pdhiran: Sure. Just add a debug statement somewhere in next PR, so that we can check, in how many cases, was this actually required.

pdhiran (approved):
Looks Good!

PR #2660, opened 2023-06-22
[Lib] Fix disconnected lib

adityaramteke:
Lib fix for the changes required for disconnected configuration

### DIFF
--- a/cephci/prereq.py
+++ b/cephci/prereq.py
@@ -18,10 +18,10 @@ from cephci.utils.configure import (
     create_registry_directories,
     create_self_signed_certificate,
     exec_cephadm_preflight,
+    get_private_registry_image,
     set_registry_credentials,
     setup_ssh_keys,
     start_local_private_registry,
-    update_ca_trust,
     validate_trusted_list,
 )
 from cli.exceptions import NodeConfigError
@@ -36,8 +36,6 @@ from utility.log import Log
 log = Log(__name__)
 
 CEPHADM_ANSIBLE = "cephadm-ansible"
-REGISTRY_USERNAME = "Test"
-REGISTRY_PASSWORD = "Test"
 
 doc = """
 Utility to configure prerequisites for deployed cluster
@@ -169,9 +167,32 @@ def enable_rhel_repos(node, server, distro):
     return True
 
 
-def setup_private_container_registry(installer, nodes, registry, build_type):
+def setup_private_container_registry(
+    installer,
+    nodes,
+    registry,
+    reg_username,
+    reg_password,
+    build_type,
+    private_reg_username,
+    private_reg_password,
+    detach,
+    images=None,
+):
     """
     Performs the pre-reqs required for the disconnected install
+
+    Args:
+        installer (ceph.ceph.Ceph): Installer node
+        nodes (ceph.ceph.Ceph): List of nodes
+        registry (str): registery name
+        reg_username (str): registry username
+        reg_password (str): registry password
+        build_type (str): build type
+        private_reg_username (str): private registry username
+        private_reg_password (str): private registry password
+        detach (str): detach/docker-registry image name
+        images (str): list of images
     """
 
     # Step 1: Create folders for the private registry
@@ -179,7 +200,9 @@ def setup_private_container_registry(installer, nodes, registry, build_type):
         return False
 
     # Step 2: Create credentials for accessing the private registry
-    if not set_registry_credentials(installer, REGISTRY_USERNAME, REGISTRY_PASSWORD):
+    if not set_registry_credentials(
+        installer, private_reg_username, private_reg_password
+    ):
         return False
 
     # Step 3: Create a self-signed certificate
@@ -201,24 +224,35 @@ def setup_private_container_registry(installer, nodes, registry, build_type):
 
     # Step 6: Copy the certificate to any nodes that will access the private registry for installation and update the
     # trusted list
-    if not copy_cert_to_secondary_node(installer, nodes[1]):
+    if not copy_cert_to_secondary_node(installer, nodes):
         return False
 
-    # Update ca-trust in secondary node
-    if not update_ca_trust(nodes[1]):
+    # Step 7: Login to the registry
+    for node in nodes:
+        Registry(node).login(
+            registry=registry, username=reg_username, password=reg_password
+        )
+
+    # Step 8: Start the local secure private registry
+    if not start_local_private_registry(installer, detach):
         return False
 
-    # Verify trust list is updated
-    if not validate_trusted_list(nodes[1]):
-        return False
-
-    # Step 7: Start the local secure private registry
-    if not start_local_private_registry(installer):
-        return False
-
-    # Step 8: Add images to the private registry
+    # Step 9: Add images to the private registry
     if not add_images_to_private_registry(
-        installer, REGISTRY_USERNAME, REGISTRY_PASSWORD, registry, build_type
+        installer,
+        reg_username,
+        reg_password,
+        private_reg_username,
+        private_reg_password,
+        registry,
+        build_type,
+        images,
+    ):
+        return False
+
+    # Step 10: List down private registry images
+    if not get_private_registry_image(
+        installer, private_reg_username, private_reg_password
     ):
         return False
 
--- a/cephci/utils/configure.py
+++ b/cephci/utils/configure.py
@@ -1,4 +1,5 @@
-from cephci.utils.configs import get_images
+from json import loads
+
 from cli.cephadm.ansible import Ansible
 from cli.exceptions import AnsiblePlaybookExecutionError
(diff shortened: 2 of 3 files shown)


cephci/utils/configure.py
```diff
@@ -277,15 +296,24 @@ def add_images_to_private_registry(node, username, password, registry, build_typ
...
+    con = get_cephci_config()
+    src_username = con.get("cdn_credentials", {}).get("username")
+    src_password = con.get("cdn_credentials", {}).get("password")
+    src_creds = f"{src_username}:{src_password}"
     dest_creds = f"{username}:{password}"
     dst_creds_dir = "./certs/"
 
-    # Get the required images here
-    images = get_images(build_type)
+    # # Get the required images here
```
vamahaja, line 307: ????

cli/utilities/containers.py
```diff
@@ -53,72 +53,85 @@ def __init__(self, nodes, package="podman"):
...
+        """Pull container image
+
+        Args:
+            image (str): Container image url
+        """
+
+        cmd = f"{self.base_cmd} pull"
+        if image:
+            cmd += f" {image}"
+        else:
```
vamahaja, line 133: This else part is not required as name is mandatory parameter

cli/utilities/containers.py
```diff
@@ -53,72 +53,85 @@ def __init__(self, nodes, package="podman"):
...
             raise ContainerRegistryError(f"Failed to run {cmd} on container")
+
+    def pull(self, image):
+        """Pull container image
+
+        Args:
+            image (str): Container image url
+        """
+
+        cmd = f"{self.base_cmd} pull"
```
vamahaja, line 130: this should be `cmd = f"{self.base_cmd} pull name"`

vamahaja (changes requested):

cephci/prereq.py
```diff
@@ -32,12 +32,13 @@
...
 from cli.utilities.waiter import WaitUntil
 from utility.log import Log
+from utility.utils import get_cephci_config
 
 log = Log(__name__)
 
 CEPHADM_ANSIBLE = "cephadm-ansible"
-REGISTRY_USERNAME = "Test"
-REGISTRY_PASSWORD = "Test"
+REGISTRY_USERNAME = "myregistryusername"
```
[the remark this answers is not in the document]
adityaramteke: Better to have any name than simply saying test with starting upper case letter.

cli/utilities/containers.py
adityaramteke: Sure, I'll remove that.

cephci/prereq.py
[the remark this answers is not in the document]
vamahaja: Default passwords can be anything. Either provide an option to pass these details to library or revert these changes.

cephci/prereq.py
```diff
@@ -201,27 +204,32 @@ def setup_private_container_registry(installer, nodes, registry, build_type):
...
-    # Verify trust list is updated
-    if not validate_trusted_list(nodes[1]):
-        return False
+    # Login to registry
+    con = get_cephci_config()
+    username = con.get("cdn_credentials", {}).get("username")
+    password = con.get("cdn_credentials", {}).get("password")
+    for node in nodes:
+        Registry(node).login(
+            registry="registry.redhat.io", username=username, password=password
```
vamahaja, line 216: Why hard coded registry ? Will it work for downstream, stage or ibm builds ?

cephci/utils/configure.py
```diff
@@ -277,15 +296,25 @@ def add_images_to_private_registry(node, username, password, registry, build_typ
         f"{DOMAIN_CERT_PATH_CERT}:/certs/domain.cert:Z",
     ]
 
-    src_creds = f"{username}:{password}"
+    # Get login credentials
+    con = get_cephci_config()
+    src_username = con.get("cdn_credentials", {}).get("username")
+    src_password = con.get("cdn_credentials", {}).get("password")
```
vamahaja, line 302: Same as above, this is restricting library to live server. Pass all these details through function parameter

cephci/utils/configure.py
```diff
@@ -277,15 +296,25 @@ def add_images_to_private_registry(node, username, password, registry, build_typ
...
+    src_creds = f"{src_username}:{src_password}"
     dest_creds = f"{username}:{password}"
     dst_creds_dir = "./certs/"
 
     # Get the required images here
-    images = get_images(build_type)
+    images = list(images)
+    if images and type(images) not in (tuple, list):
+        images = [images]
+    else:
```
vamahaja, line 311: Remove this one, keep image as mandatory parameter.

vamahaja (changes requested):

cephci/utils/configure.py
adityaramteke: Sure, I'll pass these through a test case.

cli/utilities/containers.py
```diff
@@ -53,72 +53,81 @@ def __init__(self, nodes, package="podman"):
...
+            cmd += f" --name {name}"
+
         if restart:
             cmd += f" --restart={restart}"
 
-        if rm:
-            cmd += f" --rm {rm}"
-        else:
-            cmd += f" {image}"  # Add image only if not an rm operation
+        if name:
```
vamahaja, line 89: repeated `name` parameter check

cephci/utils/configure.py
```diff
@@ -225,19 +226,36 @@ def validate_trusted_list(node):
...
+    """Copy the certificate to another nodes in the cluster"""
+    for secondary_node in secondary_nodes:
+        cmd = f"scp {DOMAIN_CERT_PATH} root@{secondary_node.hostname}:{CA_TRUST_ANCHORS_PATH}"
+        out, _ = node.exec_command(cmd=cmd, sudo=True, timeout=800)
+        if out:
+            log.error(f"Failed to copy the certificate to {secondary_node} node")
+            return False
+
+        # Update the ca-trust
+        if not update_ca_trust(node):
```
vamahaja, line 239: Add log error

cephci/utils/configure.py
```diff
@@ -225,19 +226,36 @@ def validate_trusted_list(node):
...
+            return False
+
+        # Validate the trusted list
+        if not validate_trusted_list(node):
```
vamahaja, line 243: Add log error

vamahaja (changes requested):

cli/utilities/containers.py
```diff
@@ -53,72 +53,78 @@ def __init__(self, nodes, package="podman"):
...
             volume (list): List of volumes
             ports (list): List of ports
             rm (list): List of entries to be removed
-            detach (bool): Detach flag
+            detach (str): Detach flag
             restart (str): Restart flag
             cmds (str): Other commands to be executed
         """
-        if not image and not rm:
+        if not name and not rm:
```
vamahaja, line 78: Why this change ? which image is mandatory, not name

adityaramteke: Yes, you are correct, based on the general run command, IMAGE is a mandatory parameter:
``` podman run [options] IMAGE [COMMAND [ARG...]] ```

I've some queries regarding options we are currently using for podman run command for disconnected installation:

We are passing image with different parameters for e.g.  --detach IMAGE (Basically to run the container in background) and --rm IMAGE (To remove container when stopped)

 How should we handle such scenarios with the current implementation?

Comment (vamahaja) [reply cli/utilities/containers.py:78]: @adityaramteke as per documentation [1]. 

`name` is not a mandatory parameter. `image` is mandatory and `rm` is optional parameter which requires `image` to be passed.
Also in case  `detach` parameter doesn't have any value then it works as `-d` (which will run container in background) but here `detach` is passed as list or string as `--detach` it accepts multiple string, here we `IMAGE` is used because it's passed at the end.

[1] https://docs.podman.io/en/latest/markdown/podman-run.1.html

Comment (adityaramteke) [reply cli/utilities/containers.py:78]: Thanks, Vaibhav! I've made the changes.

File: cli/utilities/containers.py [hunk 12]
```diff
@@ -53,72 +53,79 @@ def __init__(self, nodes, package="podman"):
...
-            rm (list): List of entries to be removed
-            detach (bool): Detach flag
+            rm (str): Image name to remove from background
+            detach (str): Image name to detach
             restart (str): Restart flag
             cmds (str): Other commands to be executed
         """
-        if not image and not rm:
-            raise NotSupportedError("Image name or rm option needs to be provided")
+        if not image:
```
vamahaja, line 78: This will never be executed.

adityaramteke: Resolved.

vamahaja (commented):
LGTM


@pranavprakash20 please review

pranavprakash20 (approved):
LGTM

vamahaja (approved):

PR #2672, opened 2023-06-26
[FibFix] Fix CloudProvider libraries

vamahaja:
Fix consists of -

- Spelling correction for `CloudProviders` directory.
- Convert node methods to properties.
- Separate class `Volume` for volume operations.
- Move cloud provider config read operation to separate class 'CloudProvider'.
- Additional steps to validate resource cleanup


cli/cloudproviders/openstack.py
```diff
@@ -0,0 +1,426 @@
...
+                _state = self.get_node_state_by_id(id)
+            except Exception as e:
+                log.error(e)
+                raise CloudProviderError(e)
+
+            if _state == state:
+                log.info(f"Node '{id}' is in exptected '{_state}' state")
+                break
+
+            msg = f"Node '{id}' is in expected '{_state}' state, retry after {interval} sec"
```
tintumathew10, line 151: I think this msg should be 
msg = f"Node '{id}' is not in expected '{_state}' state, retry after {interval} sec"

cli/cloudproviders/openstack.py
```diff
@@ -0,0 +1,426 @@
...
+                    _state = STATE_DESTROY
+                else:
+                    log.error(e)
+                    raise CloudProviderError(e)
+
+            if _state == state:
+                log.info(f"Volume '{id}' is in expected '{_state}' state")
+                break
+
+            msg = f"Volume '{id}' is in '{_state}' state, retry after {interval} sec"
```
tintumathew10, line 322: same as above.

MohitBis (approved):
LGTM

PR #2688, opened 2023-06-28
Scale testing - NVMeOF Single GW components

rahullepakshi:
Adding test suite for scaling nvmeof single GW components - namespaces and subsystems as per 7.0 TP requirement

Test scenario - Test as per below design and cleanup the environment
<html>
<body>
<b style="font-weight:normal;" id="docs-internal-guid-66fbd557-7fff-d988-0162-36d1515ffb35"><div dir="ltr" style="margin-left:33pt;" align="left">

  | Subsystems | Namespaces | Image count | Image size
-- | -- | -- | -- | --
Scale -1 | 1 | 64 | 64 | 50G
Scale -2 | 2 | 128 | 128 | 50G
Scale -3 | 3 | 192 | 192 | 50G
scale-4 | 4 | 256 | 256 | 50G

</div></b>
</body>
</html>

Test logs at http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-X634J1/

### DIFF
--- a/suites/quincy/nvmeof/ceph_nvmeof_baremetal_scale.yaml
+++ b/suites/quincy/nvmeof/ceph_nvmeof_baremetal_scale.yaml
@@ -1,0 +1,235 @@
+# Test module to scale Single GW components - scale factor being 1 subsystem and 64 namespaces with 4 scale steps
+# Test conf at conf/quincy/upi/octo-9-node-nvmeof.yaml
+# Test attributes
+  #  Single ceph-nvmeof GW colocated with osd on node5, node 6,7,8 and 9 are nvmeof initiators
+  #  nvmeof GW - at end of each scale step/ test below is the configuration
+     # Scale-1 : 1 subsystem, 64 namespaces,  64 RBD images of 50G size each
+     # Scale-2 : 2 subsystem, 128 namespaces, 128 RBD images of 50G size each
+     # Scale-3 : 3 subsystem, 192 namespaces, 192 RBD images of 50G size each
+     # Scale-4 : 4 subsystem, 256 namespaces, 256 RBD images of 50G size each
+  #  nvmeof initiator - Each initiator/ client connects to a subsystem ( 4 initiator : 4 subsystems)
+  #  io test (no performance tests)
+     # Tool : fio
+     # io is run on all nvme volumes listed on all initiators configured to that point
+     # io type combination : For volumes listed on an initiator a write in first test followed by a read in next test
+     # io size
+        # write : 40% on all volumes at end of test suite (incremental 20% writes at every write test for initiator 1, 2)
+        # Read : 5% on all volumes at every read test
+  # Check ceph health and rbd image usage at end of each test
+tests:
+  - test:
+      abort-on-fail: true
+      module: install_prereq.py
+      name: install ceph pre-requisites
+
+  - test:
+      abort-on-fail: true
+      config:
+        verify_cluster_health: true
+        steps:
+          - config:
+              command: bootstrap
+              service: cephadm
+              args:
+                mon-ip: node1
+                registry-url: registry.redhat.io
+                allow-fqdn-hostname: true
+          - config:
+              command: add_hosts
+              service: host
+              args:
+                attach_ip_address: true
+                labels: apply-all-labels
+          - config:
+              command: apply
+              service: mgr
+              args:
+                placement:
+                  label: mgr
+          - config:
+              command: apply
+              service: mon
+              args:
+                placement:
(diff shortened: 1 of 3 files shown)


suites/quincy/nvmeof/ceph_nvmeof_baremetal_scale.yaml
```diff
@@ -0,0 +1,282 @@
...
+          - node: node7
+            size: 20%
+            io_type: write
+      desc: test with 2 subsystems and 128 namespaces in Single GW
+      destroy-cluster: false
+      module: test_ceph_nvmeof_gateway_scale.py
+      name: Scale subsystems and namespaces in Single NVMeOF GW
+      polarion-id:
+
+  - test:
```
preethisn, line 153: There are multiple occurrences of the same test cases for checking Ceph image usage and health. Can we consolidate and have them in the same testss instead of caling them seapartely each tim. We can inlcude repeated test cases into a single instance

suites/quincy/nvmeof/ceph_nvmeof_baremetal_scale.yaml
```diff
@@ -0,0 +1,282 @@
...
+              size: 50G
+            listener_port: 5002
+            allow_host: "*"
+        initiators:
+          - subnqn: nqn.2016-06.io.spdk:cnode2
+            listener_port: 5002
+            node: node7
+        run_io:
+          - node: node6
+            size: 5%
```
preethisn, line 142: Can we add description for size and what it means and whererever applicable.This will make it easier for others to understand the purpose of each test case.

suites/quincy/nvmeof/ceph_nvmeof_baremetal_scale.yaml
```diff
@@ -0,0 +1,282 @@
...
+      rbd_pool: rbd
+      do_not_create_image: true
+      rep-pool-only: true
+      rep_pool_config:
+        pool: rbd
+      install: true
+      subsystems:
+        - nqn: nqn.2016-06.io.spdk:cnode1
+          serial: 1
+          max_ns: 64
```
preethisn, line 82: Can we have 256 namespace support for single subysystem as per TP requirenment instead of having 64.

suites/quincy/nvmeof/ceph_nvmeof_baremetal_scale.yaml
rahullepakshi: Thanks for the review. 
I think we cannot consolidate these checks to main tests as they use separate test modules. But can you please share me a link or example how we can achieve this?

suites/quincy/nvmeof/ceph_nvmeof_baremetal_scale.yaml
rahullepakshi: 1 subsystem 256 namespace for subsystem - This would be my next test suite I will be adding 
4 subsystem 64 namespace for each subsystem - covered in this test suite
Above 2 are two extreme scale test scenario from TP and -  the below 2 has coverage from above 2 scenarios and this need not necessarily need separate test scenario
2 subsystem 128 namespace for each subsystem
3 subsystem 86 namespace for each subsystem

suites/quincy/nvmeof/ceph_nvmeof_baremetal_scale.yaml
preethisn: Ceph health check and ceph usage are part of the test case as validation check points. We should have the code flow in the same module. We can define a method and call the function in the same module.

suites/quincy/nvmeof/ceph_nvmeof_baremetal_scale.yaml
preethisn: IMO, 64 is not required for test coverage, 256 max is across the gateway. So we should ideally design as per TP and try to cover more combination from that. No need of separate test again for 256.

suites/quincy/nvmeof/ceph_nvmeof_baremetal_scale.yaml
rahullepakshi: addressed this

suites/quincy/nvmeof/ceph_nvmeof_baremetal_scale.yaml
rahullepakshi: addressed this

rahullepakshi:
@sunilkumarn417 please review

sunilkumarn417 (approved):

PR #2694, opened 2023-06-30
Adding tests to perform cache flush with IOs stopped abruptly

manasagowri:
Automating test case - https://polarion.engineering.redhat.com/polarion/#/project/CEPH/workitem?id=CEPH-83574893

Verify cache flush command when IO is stopped abruptly
with persistent cache enables via same or different IO tools in SSD mode

Steps : 
1) Setup persistent write cache for an image
2) Write data to the image using fio jobs
3) Stop fio abruplty and verify that cache flush command

Success Logs - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-NSAV96/Validate_cache_flush_with_persistent_cache_enabled_0.log

### DIFF
--- a/suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
+++ b/suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
@@ -312,3 +312,28 @@ tests:
       module: test_rbd_persistent_write_back_cache.py
       name: PWL cache creation with exclusive lock
       polarion-id: CEPH-83574719
+
+  - test:
+      abort-on-fail: true
+      config:
+        level: client                        # PWL at client
+        cache_file_size: 1073741824          # 1 GB
+        rbd_persistent_cache_mode: ssd       # "ssd" or "rwl" on pmem device
+        client: node5
+        drive: /dev/nvme0n1
+        cleanup: true
+        # validate_exclusive_lock: true
+        rep-pool-only: True
+        rep_pool_config:
+          pool: pool2
+          image: image2
+          size: 10G
+        fio:
+          image_name: image2
+          pool_name: pool2
+          runtime: 120
+      desc: Validate cache flush with persistent cache enabled
+      destroy-cluster: false
+      module: test_rbd_persistent_writeback_cache_flush.py
+      name: Validate cache flush with persistent cache enabled
+      polarion-id: CEPH-83574893
--- a/tests/rbd/rbd_peristent_writeback_cache.py
+++ b/tests/rbd/rbd_peristent_writeback_cache.py
@@ -182,8 +182,14 @@ class PersistentWriteAheadLog:
                 f"{self.client.hostname}:{self.pwl_path} cache file did not found!!!"
             )
 
-    def flush(self, image):
-        pass
+    def flush(self, image_spec):
+        """Perform cache flush on the image
+        Args:
+            image_spec: <pool>/<image> where cache flush needs to be performed.
+        """
+        log.info(f"Perform cache flush on image {image_spec}....")
+        cmd = f"rbd persistent-cache flush {image_spec}"
+        return self.client.exec_command(cmd=cmd, sudo=True)
 
     def invalidate(self, image):
         pass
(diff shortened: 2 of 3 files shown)


tests/rbd/rbd_peristent_writeback_cache.py
```diff
@@ -182,8 +182,11 @@ def check_cache_file_exists(self, image, timeout=120, **kw):
                 f"{self.client.hostname}:{self.pwl_path} cache file did not found!!!"
             )
 
-    def flush(self, image):
-        pass
+    def flush(self, image_spec):
+        """Perform cache flush on the image"""
```
sunilangadi2, line 186: please add `args` string in docs.

tests/rbd/test_rbd_persistent_writeback_cache_flush.py
```diff
@@ -0,0 +1,124 @@
...
+from utility.utils import run_fio
+
+log = Log(__name__)
+
+
+def kill_fio(rbd):
+    """
+    Kill the fio process running on the client
+    """
+    sleep(60)
```
sunilangadi2, line 42: why we need to sleep here?

tests/rbd/test_rbd_persistent_writeback_cache_flush.py
```diff
@@ -0,0 +1,124 @@
...
+    Args:
+        ceph_cluster: ceph cluster object
+        **kw: test parameters
+
+    Pre-requisites :
+        - need client node with ceph-common package, conf and keyring files
+        - FIO should be installed on the client.
+
+    """
+    log.info("Running PWL....")
```
sunilangadi2, line 95: looks like this info is redundant.

suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
```diff
@@ -312,3 +312,28 @@ tests:
...
+        level: client                        # PWL at client
+        cache_file_size: 1073741824          # 1 GB
+        rbd_persistent_cache_mode: ssd       # "ssd" or "rwl" on pmem device
+        client: node5
+        drive: /dev/nvme0n1
+        cleanup: true
+        # validate_exclusive_lock: true
+        rep-pool-only: True
+        rep_pool_config:
+          pool: pool2
```
sunilangadi2, line 328: can we take `pool name `and ` image name` from utils default config?

suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
```diff
@@ -312,3 +312,28 @@ tests:
       module: test_rbd_persistent_write_back_cache.py
       name: PWL cache creation with exclusive lock
       polarion-id: CEPH-83574719
+
+  - test:
```
sunilangadi2, line 316: Does it not applicable to `pacific`?

tests/rbd/test_rbd_persistent_writeback_cache_flush.py
manasagowri: So the idea here is that we start writing IOs and kill the fio process abruptly in between and check the operation of cache flush command.
Since we need to stop IOs after a few seconds after starting them I have added the sleep here. Hope that's clear? @sunilangadi2

suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
manasagowri: This is a standard we have followed for all persistent cache test cases till now, I would like to retain this for now. Maybe when we do SDK implementation for these we can consider these optimisations.

suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
manasagowri: None of the persistent cache tests have been added to pacific, so I would like to take that up as part of a different PR since the scope of this PR is not to add all persistent cache tests to pacific.

tests/rbd/test_rbd_persistent_writeback_cache_flush.py
manasagowri: Removed.

tests/rbd/rbd_peristent_writeback_cache.py
manasagowri: Done.

sunilangadi2 (approved):
Chnages LGTM.

PR #2709, opened 2023-07-07
Add Basic IO perf test to CI suite

sunilkumarn417:
- Include IO Perf test case to CI pipeline
- Works for Upstream pipeline as well.


Manohar-Murthy (approved):
lgtm

PR #2710, opened 2023-07-10
[LibFix] Fix `Container.run` library

vamahaja:
Fix consists of -
  1. Add missing parameter for `detach_key`.
  2. Fix library definition.
  3. Fix `image` and `rm` parameter condition

### DIFF
--- a/cli/utilities/containers.py
+++ b/cli/utilities/containers.py
@@ -54,28 +54,30 @@ class Container(Cli):
     def run(
         self,
         image=None,
-        name=None,
         rm=None,
+        name=None,
         env=None,
         volume=None,
         ports=None,
         restart=None,
-        detach=None,
+        detach_key=None,
+        detach=False,
         cmds=None,
     ):
         """Executes the provided command using podman
         Args
             image (str): Image name
+            rm (str): Image name to remove from background
             name (str): Container name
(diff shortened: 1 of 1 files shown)


adityaramteke (approved):

adityaramteke:
LGTM

cli/utilities/containers.py
```diff
@@ -54,25 +54,27 @@ def __init__(self, nodes, package="podman"):
...
             volume (list): List of volumes
             ports (list): List of ports
-            rm (str): Image name to remove from background
-            detach (str): Image name to detach
             restart (str): Restart flag
+            detach_key (list): List of detach operation
+            detach (str): Image name to detach
             cmds (str): Other commands to be executed
         """
         if not image or not rm:
```
adityaramteke, line 80:
```suggestion
        if not image and not rm:
```
The condition will not be fulfilled when passing only the image.

adityaramteke (changes requested):

cli/utilities/containers.py
vamahaja: Fixed now

adityaramteke (approved):

MohitBis (approved):
LGTM

PR #2789, opened 2023-07-31
Fix tox issue with value type comparison

sunilkumarn417:
Fix tox issue, This is because newer update in the flake8 version

https://github.com/red-hat-storage/cephci/actions/runs/5712253957/job/15475385088
```
./tests/rados/test_osd_args.py:54:12: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
```


harshkumarRH (approved):
Minor change, LGTM
There is a small mistake from my end in line#56 of tests/rados/test_osd_args.py, which I will fix with another PR

AmarnatReddy (approved):

PR #2810, opened 2023-08-07
[Reef] Add checks for supporting reef

vamahaja:
Tests are getting aborted for reef as support for RHCS 7 is missing from test workflow. Add checks for reef builds.


tintumathew10 (commented):
LGTM

tintumathew10 (approved):

PR #2811, opened 2023-08-08
[Utility] Update reports utility to support different options

vamahaja:
Update `cephci/reports.py` -

1. Add check for service
2. Mandatory parameter `testrun-template`
3. Optional parameter `testrun-id` & `testrun-title`
4. Add logging utility for logging in log dir


MohitBis (approved):
LGTM

PR #2834, opened 2023-08-14
core_workflows: Updated command to be executed on client node.

pdhiran:
In Stretch Clusters, while we are testing Arbiter node down scenarios, all commands need to be run on Client node. changed the command execution path. 

Pass logs : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-D00F05


neha-gangadhar (approved):
looks good

PR #2841, opened 2023-08-21
Enhance NVMeOF scale test modules

rahullepakshi:
Enhance NVMeOF scale test modules as below
1) Handle FIO failures
2) Fetch NVMeOF GW log at end of each test
3) Log rbd image usage, total storage capacity usage and ceph health at end of each test


Manohar-Murthy (approved):
lgtm.

PR #2844, opened 2023-08-22
Updating RGW reef sanity suite for both single and multisite

anrao19:
# Description
Gone through all suites that are running as a part of quincy and identified testcase from each suite to cover all features.

</details>


suites/reef/rgw/sanity_rgw_multisite.yaml
```diff
@@ -277,3 +277,278 @@ tests:
...
+      clusters:
+        ceph-pri:
+          config:
+            script-name: test_Mbuckets_with_Nobjects.py
+            verify-io-on-site: ["ceph-sec"]
+            config-file-name: test_Mbuckets_with_Nobjects_multipart.yaml
+            timeout: 300
+  - test:
+      name: LargeObjGet_GC on primary
+      desc: test_LargeObjGet_GC on primary
```
hmaheswa, line 379: @anrao19 this test might fail as we are including monitoring services that slows down large object upload and it fails with timeout error. refer https://issues.redhat.com/browse/RHCEPHQE-9210. could you please recheck once

anrao19: @hmaheswa
Gone through given jira ticket, here issue seen in large gc testcase when suite contains monitoring service activation tesecase. In given suite we have not included monitoring testcase, so ideally its should not cause any issue

suites/reef/rgw/sanity_rgw_multisite.yaml
```diff
@@ -107,68 +107,6 @@ tests:
...
-                      placement:
-                        nodes:
-                          - node1
-                    - service_type: alertmanager
-                      placement:
-                        count: 1
-                    - service_type: node-exporter
-                      placement:
-                        host_pattern: "*"
-                    - service_type: crash
```
hmaheswa, line 164: I have an observation previously that crashes were not reported without crash service, need to test if it is true and add back only crash service if not dependant on other monitoring services
approving this PR for now

hmaheswa (approved):

mkasturi18 (approved):

PR #2848, opened 2023-08-23
Added IO perf tests

sunilkumarn417:
Added IO perf comparison tests for 16k block size.

**Results:**
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-4H9LYA/

### DIFF
--- a/suites/reef/nvmeof/tier-1_nvmeof_librbd_vs_nvmeof_16k_block_size.yaml
+++ b/suites/reef/nvmeof/tier-1_nvmeof_librbd_vs_nvmeof_16k_block_size.yaml
@@ -1,0 +1,173 @@
+# Basic IO Perf comparison test LibRBD vs NVMeoF protocols
+# cluster configuration: conf/quincy/nvmeof/ceph_nvmeof_sanity.yaml
+
+tests:
+
+# Set up the cluster
+  - test:
+      abort-on-fail: true
+      module: install_prereq.py
+      name: install ceph pre-requisites
+  - test:
+      abort-on-fail: true
+      config:
+        verify_cluster_health: true
+        steps:
+          - config:
+              command: bootstrap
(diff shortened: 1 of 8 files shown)


suites/reef/nvmeof/tier-1_nvmeof_librbd_vs_nvmeof_16k_block_size.yaml
```diff
@@ -0,0 +1,173 @@
...
+  - test:
+      name: libRBD VS NVMeoF - Read and Write IOtype
+      description: Write and Read IO comparison on 10G image with 16kb-block-size
+      module: test_io_perf.py
+      config:
+        iterations: 1               # number of iterations to find out average
+        io_profiles:
+          - FIO_WRITE_BS_4k_IODepth8_LIBAIO
+          - FIO_READ_BS_4k_IODepth8_LIBAIO
+        io_overrides:
```
rahullepakshi, line 80: @sunilkumarn417 considering the combinations we had/ going to have on io_type, block_size and any unqiue io args for fio, IMO we can pass all of them as overrides, say like here we can pass io_type as override. 

This will reduce number of entries at https://github.com/red-hat-storage/cephci/blob/master/utility/io/fio_profiles.py and have only 3 entries - read, writes and readwrite(replace every parameter in each entry through overrides) or 2 entries if we combine for read and write. 

This also  provides flexibility to specify them at test suite itself. Thoughts?

sunilkumarn417: @rahullepakshi , yes in a way its makes sense. lets discuss more and we could take this in next PR as an improvement.

Manohar-Murthy (approved):
lgtm.

PR #2900, opened 2023-09-06
Automating negative scenario: snapshot mirroring on pool level

manasagowri:
This PR automates the test case - CEPH-83573617 - [negative scenario] Verify snap mirroring on pool level

Success Logs - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-U4PPNH/

### DIFF
--- a/ceph/rbd/initial_config.py
+++ b/ceph/rbd/initial_config.py
@@ -549,12 +549,21 @@ def initial_mirror_config(**kw):
     kw["ceph_cluster"] = ceph_cluster
     config = kw.get("config")
 
+    ret_config["output"] = list()
     for pool_type in primary_config.get("pool_types"):
         pool_config = getdict(config.get(pool_type))
         kw["do_not_enable_mirror_on_image"] = config.get(pool_type).get(
             "do_not_enable_mirror_on_image"
         )
-        config_mirror_multi_pool(primary_config, secondary_config, pool_config, **kw)
+
+        out = config_mirror_multi_pool(
+            primary_config, secondary_config, pool_config, **kw
+        )
+
+        if "Snapshot based mirroring cannot be enabled in pool mode" in out:
+            ret_config["output"].append(f"{out} for {pool_type}")
+            continue
+
         config_mirror_multi_pool(
             secondary_config, primary_config, pool_config, is_secondary=True, **kw
         )
--- a/ceph/rbd/workflows/rbd_mirror.py
+++ b/ceph/rbd/workflows/rbd_mirror.py
@@ -377,7 +377,11 @@ def enable_image_mirroring(primary_config, secondary_config, **kw):
     mirrormode = kw.get("mirrormode")
     io_total = kw.get("io_total")
 
-    rbd_primary.mirror.image.enable(pool=pool, image=image, mode=mirrormode)
+    out = rbd_primary.mirror.image.enable(pool=pool, image=image, mode=mirrormode)
+
+    if "cannot enable mirroring: pool is not in image mirror mode" in out[1].strip():
+        return out[1]
+
     wait_for_status(
         rbd=rbd_primary,
         cluster_name=primary_cluster.name,
@@ -438,6 +442,9 @@ def config_mirror_multi_pool(
 
     config = kw.get("config")
     pool_test_config = multi_pool_config.pop("test_config", None)
+
+    output = ""
+
     for pool, pool_config in multi_pool_config.items():
         # If any pool level test config is present, pop it out
         # so that it does not get mistaken as another image configuration
@@ -463,6 +470,9 @@ def config_mirror_multi_pool(
             if (
                 not kw.get("do_not_enable_mirror_on_image")
                 and pool_config.get("mode") == "image"
+            ) or (
+                pool_config.get("mode") == "pool"
+                and pool_config.get("mirrormode") == "snapshot"
             ):
                 mirrormode = pool_config.get("mirrormode", "")
                 multi_image_config = getdict(pool_config)
@@ -479,10 +489,27 @@ def config_mirror_multi_pool(
                         "mirrormode": mirrormode,
                         "io_total": image_config_val.get("io_total", None),
                     }
-                    enable_image_mirroring(
+                    out = enable_image_mirroring(
                         primary_config, secondary_config, **image_enable_config
                     )
+
+                    if (
+                        pool_config.get("mode") == "pool"
+                        and pool_config.get("mirrormode") == "snapshot"
+                    ):
+                        if (
+                            out
+                            and "cannot enable mirroring: pool is not in image mirror mode"
+                            in out.strip()
+                        ):
+                            output = "Snapshot based mirroring cannot be enabled in pool mode"
+                        elif not is_secondary:
+                            output = (
+                                "Snapshot based mirroring did not fail in pool mode"
+                            )
 
     # Add back the popped pool test config once configuration is complete
     if pool_test_config:
         pool_config["test_config"] = pool_test_config
+
+    return output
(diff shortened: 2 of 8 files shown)


ceph/rbd/initial_config.py
```diff
@@ -554,9 +555,16 @@ def initial_mirror_config(**kw):
         kw["do_not_enable_mirror_on_image"] = config.get(pool_type).get(
             "do_not_enable_mirror_on_image"
         )
-        config_mirror_multi_pool(primary_config, secondary_config, pool_config, **kw)
+
+        out = config_mirror_multi_pool(
+            primary_config, secondary_config, pool_config, **kw
+        )
+
+        # out_config.update({"cluster_1": config_mirror_multi_pool(primary_config, secondary_config, pool_config, **kw)})
```
sunilangadi2, line 563: Please remove the commented code.

suites/quincy/rbd/tier-2_rbd_mirror_snapshot_regression.yaml
```diff
@@ -8,183 +8,202 @@
...
       clusters:
         ceph-rbd1:
           config:
-            imagesize: 2G
-            ec_pool_config:
-              mirrormode: snapshot
-              mode: image
             rep_pool_config:
+              num_pools: 1
+              num_images: 1 # one each of below encryptions will be applied on each image
```
sunilangadi2, line 202: please remove comments it's not applicable here.

suites/quincy/rbd/tier-2_rbd_mirror_snapshot_regression.yaml
```diff
@@ -8,183 +8,202 @@
...
+  #             mode: image
+  #           snapshot_schedule_level: "cluster"
+  #     polarion-id: CEPH-83575376
+  #     desc: Create snapshot based RBD mirrored pools, schedule snapshots at cluster level and verify
   - test:
-      name: test_rbd_mirror_snapshot_pool
-      module: test_rbd_mirror_snapshot.py
+      abort-on-fail: True
+      desc: Verify snapshot based mirroring cannot be enabled at pool mode
+      name: Verify snapshot based mirroring cannot be enabled at pool mode
```
sunilangadi2, line 194: please provide different statement for test `name` and `desc`
name can be `Test to validate negative scenario with pool based snapshot mirroring`

ceph/rbd/workflows/rbd_mirror.py
```diff
@@ -463,6 +470,9 @@ def config_mirror_multi_pool(
             if (
                 not kw.get("do_not_enable_mirror_on_image")
                 and pool_config.get("mode") == "image"
+            ) or (
+                pool_config.get("mode") == "pool"
```
sunilangadi2, line 474: `snapshot` based applicable only for `image` based Is it?

suites/quincy/rbd/tier-2_rbd_mirror_snapshot_regression.yaml
```diff
@@ -188,3 +188,22 @@ tests:
             snapshot_schedule_level: "cluster"
       polarion-id: CEPH-83575376
       desc: Create snapshot based RBD mirrored pools, schedule snapshots at cluster level and verify
+  - test:
```
sunilangadi2, line 191: please add one line space before new test.

tests/rbd_mirror/test_snapshot_mirroring_on_pool_negative.py
```diff
@@ -0,0 +1,61 @@
...
+        "Running negative scenario of enabling snapshot based mirroring at pool level"
+    )
+    try:
+        ret_val = 0
+        kw.get("config", {})["do_not_run_io"] = True
+        mirror_obj = initial_mirror_config(**kw)
+        if not mirror_obj.get("output"):
+            log.error("Snapshot based mirroring did not fail in pool mode")
+            ret_val = 1
+        # for output in mirror_obj.get("output", {}).values():
```
sunilangadi2, line 45: please remove these commented code.

tests/rbd_mirror/test_snapshot_mirroring_on_pool_negative.py
```diff
@@ -0,0 +1,61 @@
...
+    1. Bootstrap two CEPH clusters and create a pool and an image each for replicated and ec pools
+    2. Setup snapshot based mirroring in between these clusters at pool level
+    3. Verify that enabling snapshot based mirroring at pool level errors out appropriately
+    """
+    pool_types = ["rep_pool_config", "ec_pool_config"]
+    log.info(
+        "Running negative scenario of enabling snapshot based mirroring at pool level"
+    )
+    try:
+        ret_val = 0
```
sunilangadi2, line 39: instead of this variable, we can directly return 0 or 1 Is it?

tests/rbd_mirror/test_snapshot_mirroring_on_pool_negative.py
```diff
@@ -0,0 +1,61 @@
...
+            ret_val = 1
+        # for output in mirror_obj.get("output", {}).values():
+        for output in mirror_obj.get("output"):
+            if output:
+                log.info(output)
+            else:
+                log.error("Snapshot based mirroring did not fail in pool mode")
+                ret_val = 1
+
+        # log.info("Initial configuration complete")
```
sunilangadi2, line 53: please remove this commented code.

tests/rbd_mirror/test_snapshot_mirroring_on_pool_negative.py
```diff
@@ -0,0 +1,61 @@
...
+    3. Verify that enabling snapshot based mirroring at pool level errors out appropriately
+    """
+    pool_types = ["rep_pool_config", "ec_pool_config"]
+    log.info(
+        "Running negative scenario of enabling snapshot based mirroring at pool level"
+    )
+    try:
+        ret_val = 0
+        kw.get("config", {})["do_not_run_io"] = True
+        mirror_obj = initial_mirror_config(**kw)
```
sunilangadi2, line 41: can we mention `log.info` for executing test on Replicated pool and EC pool how we did other test?

sunilangadi2 (changes requested):
@manasagowri please cover this test in `pacific` and `reef` as well?

ceph/rbd/workflows/rbd_mirror.py
manasagowri: Yes, that's why this is a negative scenario test.

tests/rbd_mirror/test_snapshot_mirroring_on_pool_negative.py
manasagowri: Since the complete test is being executed in one go inside the initial_mirror_config method itself, that log can't be written. However when the test is complete, there is a log which says Snapshot mirroring cannot be enabled on pool mode for both rep pool and ec pool separately. I think this log should be sufficient.

tests/rbd_mirror/test_snapshot_mirroring_on_pool_negative.py
manasagowri: Since we are checking in loop for whether the test worked appropriately for both rep pool and ec pool, we are assigning one variable with 0 or 1 based on pass or failure. At the end so that even if one fails, it will return back 1. So the variable is required here.

tests/rbd_mirror/test_snapshot_mirroring_on_pool_negative.py
sunilangadi2: i am saying like this way https://github.com/red-hat-storage/cephci/blob/2faeceac9b6611f023511c0bfef12d9db9880928/tests/rbd_mirror/test_mirror_move_secondary_trash.py#L75

sunilangadi2 (approved):
LGTM.

PR #2919, opened 2023-09-12
[LibFix] Add container rmi and inspect lib

adityaramteke:
Add support for container rmi and inspect libraries:

- rmi lib will help to remove container image from node
- inspect lib will help inspecting the container image


adityaramteke:
Closing the pull request as created duplicate #2920

PR #2946, opened 2023-09-26
Updated quincy upgrade suites to use correct tags from recipe file

tintumathew10:
# Description

Updated quincy upgrade suites to use correct tags from recipe file


MohitBis (approved):
LGTM

vamahaja (approved):

PR #2952, opened 2023-09-26
Adding baremetal conf file for nvmeof test execution

rahullepakshi:
Adding Cali systems conf file for nvmeof scale test execution


Manohar-Murthy (approved):
approving temporary barmetal execution config.

PR #2970, opened 2023-09-28
Fixing client evict loop

AmarnatReddy:
# Description
Fixing client evict loop

**Problem :** 
PID which has been listed as part of `ceph tell mds.0 client ls | grep '"pid":'` pid was not found in any of the clients 
the expectation was the pid will be present in any one of the clients.

**Solution:**
We are trying on all the clients which are present and raising exception if the pid is not present even after looping over all the clients

<details>

</details>

### DIFF
--- a/tests/cephfs/cephfs_utilsV1.py
+++ b/tests/cephfs/cephfs_utilsV1.py
@@ -358,8 +358,10 @@ class FsUtils(object):
         grep_pid_cmd = """sudo ceph tell mds.%d client ls | grep '"pid":'"""
         out, rc = client_node.exec_command(cmd=grep_pid_cmd % rank)
         client_pid = re.findall(r"\d+", out)
+        successful_clients = 0
         while True:
             for client in clients:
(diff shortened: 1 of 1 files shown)


tests/cephfs/cephfs_utilsV1.py
```diff
@@ -369,6 +371,10 @@ def auto_evict(client_node, clients, rank):
                 except Exception as e:
                     print(e)
                     pass
+            if successful_clients == len(clients):
```
sumabai, line 374: Please add test logs with fix

AmarnatReddy: Logs : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-7UU1JT/

These are failed logs for different reasons but with this fix, it will not get stuck in an infinite loop if it does not find pid

neha-gangadhar (approved):
lgtm

PR #3018, opened 2023-10-12
CEPH-83576084-Delete and recreate a LUN with same name

sunilkumarn417:
CEPH-83576084-Delete and recreate a LUN with same name and remap it on same initiator


Manohar-Murthy (approved):
lgtm.

PR #3045, opened 2023-10-19
Install rbd-nbd before running device map in krbd

manasagowri:
Baremetal pipeline fix - 

http://magna002.ceph.redhat.com/cephci-jenkins/test-runs/baremetal/18.2.0-44/rados/2/tier-2_rbd_encryption/ - 
Since we create cluster and configure client only once, we would not have installed rbd-nbd in the previous tests. We need to make sure it is installed before running any of the encryption tests. 

Success Log - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-WYVBE4/

### DIFF
--- a/ceph/rbd/workflows/krbd_io_handler.py
+++ b/ceph/rbd/workflows/krbd_io_handler.py
@@ -80,6 +80,12 @@ def krbd_io_handler(**kw):
                 device_names.append(rbd.map(pool=pool_name, image=image_name)[:-1])
 
             if operations.get("device_map"):
+                out = exec_cmd(
+                    cmd="rpm -qa|grep rbd-nbd", node=client, sudo=True, output=True
+                )
+                if not out or out == 1:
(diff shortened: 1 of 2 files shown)


ceph/rbd/workflows/krbd_io_handler.py
```diff
@@ -80,6 +80,10 @@ def krbd_io_handler(**kw):
                 device_names.append(rbd.map(pool=pool_name, image=image_name)[:-1])
 
             if operations.get("device_map"):
+                out = rbd.exec_cmd(cmd="rpm -qa|grep rbd-nbd", sudo=True)
```
sunilangadi2, line 83: what is the reason of not checking `output=True` here like below code?

manasagowri: That was by mistake, thanks for pointing it out Sunil I have made the changes now.

sunilangadi2 (approved):
changes LGTM.

PR #3077, opened 2023-10-31
[LibFix] Fix cephadm ansible wrapper tests

vamahaja:


cli/utilities/operations.py
```diff
@@ -24,3 +27,30 @@ def wait_for_cluster_health(node, status, timeout=300, interval=20):
     if w.expired:
         log.error(f"Cluster is not in {status} state even after {timeout} secs")
         return False
+
+
+def wait_for_osd_daemon_state(client, id, state):
```
pranavprakash20, line 32: Can this be a generic one? I mean wait_for_daemon_state(daemon, client, id, state) ?

Not an urgent change, but just a thought

pranavprakash20 (approved):
LGTM

cli/utilities/operations.py
vamahaja: Different daemons has different commands to get status, which will make the function more complex, hence created separate method.

vdas-redhat (approved):
LGTM

PR #3091, opened 2023-11-02
Discovery port updates and CEPH-83576093 Reboot Gateway node

sunilkumarn417:
update new test case to use 8009 discovery port


Manohar-Murthy (approved):
lgtm.

PR #3119, opened 2023-11-09
Automation of running IOs on encrypted images from multiple clients

manasagowri:
Automation of test case - CEPH-83575259 - Have multiple clients and start IOs to the encrypted images and verify

    Test Case Flow
    1. Create a pool and an image
    2. Apply encryption format to the images Luks1/Luks2
    3. Load the encryption
    4. Have multiple clients and run IOs to the encrypted images

Success Logs - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-G5KFZN/


manasagowri:
Running IOs from multiple clients is not supported for block device images. Polarion test case marked as inactive.

PR #3121, opened 2023-11-09
Adding nvme disk to persistent cache suite

manasagowri:
Using nvme disk for persistent cache suite since baremetal configuration has nvme disks.


sunilangadi2 (approved):
LGTM.

PR #3132, opened 2023-11-15
add do-not-skip-test-case option to framework

sunilkumarn417:
To override the `--skip-tc` option from cephCI Command line, introducing the overriding paramter at the test case level.

Currently,

|Condition | Mod_name not in Skip Test list | Do Not Skip Test-DNST(Default: False) | Result |
|-|-|-|-|
|cli skip test provided | False | False | Test skipped |
|cli skip test provided + DNST=True | False | True | Test executed by overriding CLI option | 
|cli skip test not provided + DNST=False(by default) | True | False | Normal Test execution |

### DIFF
--- a/run.py
+++ b/run.py
@@ -684,6 +684,7 @@ def run(args):
     for test in tests:
         test = test.get("test")
         tc = fetch_test_details(test)
+        do_not_skip_test = test.get("do-not-skip-tc", False)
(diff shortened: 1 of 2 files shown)


AmarnatReddy (commented):
LGTM

AmarnatReddy (approved):

PR #3151, opened 2023-11-20
Changes after analysing baremetal pipeline failures

manasagowri:
This PR is to make changes to existing test suites for RBD so that they can be executed in both rhosd and baremetal pipelines.

The changes done to accomodate this are.

1. Since baremetal setup, doesn't install rbd-nbd as part of adding client node, we have added code in test suites and also in krbd_io_handler module to install rbd-nbd if its not already installed.
2. By default we were creating clones with v2 format as that is the default being used while creating clones, now we have added a new configuration option supported by CLI rbd-default-clone-format while creating clones, so that user can use this to add clones with required format.
3. Removed dm-cache test from quincy and reef, since it is applicable only for rhel-8
4. Added tests to remove all unnecessary epel packages from baremetal client node, since these come in the way of installing few packages required by RBD tests.
5. tier-2_rbd_regression_Image_migration.yaml was using same pool name for both replicated and ec pools causing errors, so changed the name of ec pool
6. In persistent cache tests, made changes to execute client, pool and image level tests sequentially since our baremetal client has only one nvme node and we can execute only one test at a time
7. tier-4_rbd_negative_scenario_regression.yaml - added tests to be performed for both v1 clone and v2 clone formats
8. Added code in rbd_utils to fetch client node from calling method, if passed.
9. test_concurrent_write_pwl_cache.py - In this module added logic to check for image watchers before cleaning up persistent cache settings, since clean up fails if image still has watchers.
10. test_rbd.py - Added code to create rbd pool if it doesn't exist.
11. test_rbd_deep_flatten_negative.py - Modified steps for verification based on clone format whether it is v1 or v2.
12. test_rbd_dm_cache.py and test_rbd_immutable_cache.py - Added logic to get client node from rbd object instead of test suite config
13. test_rbd_migration_image_s3_object.py - Added logic to deploy rgw daemon if it doesn't exist in the cluster.
14. All rbd_persistent_cache tests - Changed login to execute client, pool and image level sequentially.

Success Logs for the same is present in http://magna002.ceph.redhat.com/cephci-jenkins/test-runs/baremetal/18.2.0-128/rbd/9/

### DIFF
--- a/suites/reef/rbd/tier-2_rbd_cache_scenarios.yaml
+++ b/suites/reef/rbd/tier-2_rbd_cache_scenarios.yaml
@@ -84,28 +84,18 @@ tests:
       polarion-id: CEPH-83573758
 
   - test:
-      abort-on-fail: true
+      desc: Install rbd-nbd and remove any epel packages
+      module: exec.py
+      name: Install rbd-nbd
       config:
-        client: node4
-        rep-pool-only: True
-        rep_pool_config:
-          pool: pool_test
-          image1: image1
-          image2: image2
-          image3: image3
-          size: 20G
-        fio:
-          runtime: 120
-      desc: Validate dm-cache and dm-write cache for RBD based image disk
-      destroy-cluster: false
-      module: test_rbd_dm_cache.py
-      name: DM cache and DM write cache creation
-      polarion-id: CEPH-83575581
+        sudo: true
+        commands:
+          - "rm -rf /etc/yum.repos.d/epel*"
+          - "dnf install rbd-nbd"
 
   - test:
       abort-on-fail: true
       config:
-        client: node4
         rep-pool-only: True
         rep_pool_config:
           pool: test_pool
--- a/tests/rbd/test_rbd.py
+++ b/tests/rbd/test_rbd.py
@@ -44,6 +44,13 @@ def one_time_setup(node, rhbuild, branch: str) -> None:
         branch: The branch that needs to be cloned
         rhbuild: specification of rhbuild. ex: 4.3-rhel-7
     """
+    node.exec_command(cmd="sudo rm -rf /etc/yum.repos.d/epel*")
+    out, _ = node.exec_command(cmd="ceph osd pool ls")
+    if out:
+        pool_list = out.split("\n")
+        if "rbd" not in pool_list:
+            node.exec_command(cmd="ceph osd pool create rbd")
+
     node.exec_command(
         cmd=f"sudo rm -rf ceph && git clone --branch {branch} --single-branch --depth 1 {TEST_REPO}"
     )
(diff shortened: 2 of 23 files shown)


suites/reef/rbd/tier-2_rbd_cache_scenarios.yaml
```diff
@@ -84,28 +84,18 @@ tests:
       polarion-id: CEPH-83573758
 
   - test:
-      abort-on-fail: true
+      desc: Install rbd-nbd and remove any epel packages
+      module: exec.py
+      name: Install rbd-nbd
```
sunilangadi2, line 89: why is this test needed?

tests/rbd/test_rbd.py
```diff
@@ -44,6 +44,13 @@ def one_time_setup(node, rhbuild, branch: str) -> None:
         branch: The branch that needs to be cloned
         rhbuild: specification of rhbuild. ex: 4.3-rhel-7
     """
+    node.exec_command(cmd="sudo rm -rf /etc/yum.repos.d/epel*")
+    out, _ = node.exec_command(cmd="ceph osd pool ls")
+    if out:
+        pool_list = out.split("\n")
+        if "rbd" not in pool_list:
+            node.exec_command(cmd="ceph osd pool create rbd")
```
sunilangadi2, line 52: how `rbd` pool is associated with `epel` package?

tests/rbd/test_rbd_deep_flatten_negative.py
```diff
@@ -9,14 +9,71 @@
 log = Log(__name__)
 
 
+def verify_deep_flatten(clone_format, rbd, pool, image, snap_spec):
+    """ """
```
sunilangadi2, line 13: why empty doc string?

sunilangadi2 (changes requested):

suites/reef/rbd/tier-2_rbd_cache_scenarios.yaml
manasagowri: Because baremetal setup does not install rbd-nbd package. I have already mentioned this in the description for this PR.

tests/rbd/test_rbd.py
manasagowri: it's not related. why do you think it's related?

tests/rbd/test_rbd_deep_flatten_negative.py
manasagowri: that was how it was when that test was automated. Changing that is not part of this PR. I can do it separately in a different PR.

sunilangadi2 (approved):
discussed and clarified PR looks good along with test result.

PR #3161, opened 2023-11-22
[Tier2] OSD redeployment creates collocated OSD

adityaramteke:
Test case validates redeployment of OSD does not create a non- collocated OSDs

- Create non-collocated OSDs with 15G data_size and 100G db_size
- Set an unmanaged flag to OSD creation
- Zap one OSD
- Set back unmanaged flag to OSD creation
- Create new OSD

http://magna002.ceph.redhat.com/cephci-jenkins/aramteke/automation-83575600/cephci-run-MZ2584/


vamahaja (commented):
LGTM

PR #3165, opened 2023-11-23
Remove client from NS limitations conf file

rahullepakshi:
Remove client from NS limitations conf file


Manohar-Murthy (approved):
approving the fix.

PR #3166, opened 2023-11-23
Adding rbd python tests to cephci

manasagowri:
Automation of Test Case CEPH-83574791 - Create random different data and perform writes using rbd python module which provides file access to rbd images and verify

Test Steps
1. create RBD pool and image
2. create random different data and perform writes using rbd python module which provides file access to rbd images and verify

Success Logs - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-67LCPF/

### DIFF
--- a/ceph/rbd/workflows/rbd_python.py
+++ b/ceph/rbd/workflows/rbd_python.py
@@ -1,0 +1,65 @@
+import sys
+
+import rados
+import rbd
+from docopt import docopt
+
+doc = """
+A simple test suite wrapper that executes tests based on yaml test configuration
+
+ Usage:
+  rbd_python.py --pool <name>
+        (--image <name>)
+        (--image-size <size>)
+        [--conf-file FILE]
+        [--data-pool <name>]
+
+Options:
+  -h --help                         show this screen
+  --pool <pool>                     pool name
+  --image <image>                   image name
+  --image-size <size>               image size
+  --conf-file <file>                conf file path
+  --data-pool <name>                data pool for ec pool
+"""
+
+
+def rbd_python(args):
+    """
+    Create image and write random data to it, read data written and verify
+    """
+    conf_file = args.get("--conf-file", "/etc/ceph/ceph.conf")
+    pool = args.get("--pool")
+    image_name = args.get("--image")
+    image_size = int(args.get("--image-size"))
+    data_pool = args.get("--data-pool")
+
+    # Create a rados cluster instance
+    with rados.Rados(conffile=conf_file) as cluster:
+        # Open the pool object
+        with cluster.open_ioctx(pool) as ioctx:
+            # Create rbd instance
+            rbd_inst = rbd.RBD()
+            # Create rbd image
+            if data_pool:
+                rbd_inst.create(ioctx, image_name, image_size, data_pool=data_pool)
+            else:
+                rbd_inst.create(ioctx, image_name, image_size)
+            # Open rbd image for write and read
+            with rbd.Image(ioctx, image_name) as image:
+                data = str.encode("bar" * 20000)
+                image.write(data, 0)
+                mydata = image.read(0, 60000)
+                if data != mydata:
+                    print(
+                        f"Data written and read back don't match for image {image_name}"
+                    )
+                    return 1
+    print(f"Data written matches with the data read back for image {image_name}")
+    return 0
+
+
+if __name__ == "__main__":
+    args = docopt(doc, help=True)
+    rc = rbd_python(args)
+    sys.exit(rc)
--- a/suites/reef/rbd/tier-2_rbd_regression.yaml
+++ b/suites/reef/rbd/tier-2_rbd_regression.yaml
@@ -199,3 +199,17 @@ tests:
       module: test_rbd_compression.py
       name: Test to verify data compression on global, pool and image level
       polarion-id: CEPH-83574644
+
+  - test:
+      desc: Verify python rbd module
+      config:
+        do_not_create_image: true
+        rep_pool_config:
+          num_pools: 1
+          num_images: 1
+        ec_pool_config:
+          num_pools: 1
+          num_images: 1
+      module: test_rbd_python_module.py
+      polarion-id: CEPH-83574791
+      name: Test image creation, write and read data using python rbd module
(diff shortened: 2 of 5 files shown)


suites/reef/rbd/tier-2_rbd_regression.yaml
```diff
@@ -199,3 +199,17 @@ tests:
       module: test_rbd_compression.py
       name: Test to verify data compression on global, pool and image level
       polarion-id: CEPH-83574644
+
+  - test:
```
sunilangadi2, line 203: This test would be applicable to pacific and quincy, as well Is it?

tests/rbd/test_rbd_python_module.py
```diff
@@ -0,0 +1,83 @@
...
+    which provides file access to rbd images and verify
+
+    """
+    log.info(
+        "Running python rbd module image creation, write and read data - CEPH-83574791"
+    )
+
+    try:
+        client = kw.get("ceph_cluster").get_ceph_object(role="client").node
+        client.exec_command(cmd="sudo pip install docopt")
```
sunilangadi2, line 63: `docopt` is already there here https://github.com/red-hat-storage/cephci/blob/5c76b84270023f8a9b159921108190b618bc94c8/setup.py#L19
why to install again, also other test are already using it.

tests/rbd/test_rbd_python_module.py
```diff
@@ -0,0 +1,83 @@
...
+                    cmd += f" --data-pool {data_pool}"
+                out, err = client.exec_command(cmd=cmd)
+                if err or "don't match" in out:
+                    log.error(out)
+                    log.error(
+                        f"Executing rbd python to create image, read and\
+                               write data failed for image {pool}/{image}"
+                    )
+                    return 1
+                log.info(out)
```
sunilangadi2, line 35: please add `log.info` for success message as well.

ceph/rbd/workflows/rbd_python.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def rbd_python(args):
+    """
+    Create image and write random data to it, read data written and verify
+    """
+    conf_file = args.get("--conf-file", "/etc/ceph/ceph.conf")
+    pool = args.get("--pool")
+    image_name = args.get("--image")
+    image_size = int(args.get("--image-size"))
```
sunilangadi2, line 34: Image size should add with data byte parameters as well right like `Mb` `G` `T`

ceph/rbd/workflows/rbd_python.py
```diff
@@ -0,0 +1,60 @@
...
+    data_pool = args.get("--data-pool")
+
+    with rados.Rados(conffile=conf_file) as cluster:
```
sunilangadi2, line 37: please add comment for what we are doing here?

ceph/rbd/workflows/rbd_python.py
```diff
@@ -0,0 +1,60 @@
...
+        with cluster.open_ioctx(pool) as ioctx:
+            rbd_inst = rbd.RBD()
+            if data_pool:
+                rbd_inst.create(ioctx, image_name, image_size, data_pool=data_pool)
+            else:
+                rbd_inst.create(ioctx, image_name, image_size)
+            with rbd.Image(ioctx, image_name) as image:
+                data = str.encode("bar" * 20000)
+                image.write(data, 0)
```
sunilangadi2, line 46: how much size data gets writes here?

ceph/rbd/workflows/rbd_python.py
```diff
@@ -0,0 +1,60 @@
...
+                mydata = image.read(0, 60000)
+                if data != mydata:
```
sunilangadi2, line 48: not cleared about data validation why `60000` for read? Does it bar 3*20000?

ceph/rbd/workflows/rbd_python.py
```diff
@@ -0,0 +1,60 @@
...
+                    print(
+                        f"Data written and read back don't match for image {image_name}"
+                    )
+                    return 1
+    print(f"Data written matches with the data read back for image {image_name}")
+    return 0
+
+
+if __name__ == "__main__":
+    args = docopt(doc)
```
sunilangadi2, line 58: make `help=True` in beginning itself so it's easy for users.
Ref: https://github.com/red-hat-storage/cephci/blob/5c76b84270023f8a9b159921108190b618bc94c8/pipeline/scripts/ci/upstream_cli.py#L206

sunilangadi2 (changes requested):

tests/rbd/test_rbd_python_module.py
manasagowri: docopt will be installed in jenkins execution node as part of this. But I am executing rbd python tests inside the client node where again I need docopt, that's why this command will install docopt inside our cluster's client node.

tests/rbd/test_rbd_python_module.py
manasagowri: this log.info out will actually print the success message itself clearly if you check the logs for this execution.

ceph/rbd/workflows/rbd_python.py
manasagowri: No in rbd python we need to specify size in bytes as an integer itself. This is different from the way we pass in CLI command to create image.

ceph/rbd/workflows/rbd_python.py
manasagowri: around 4MB data is written

ceph/rbd/workflows/rbd_python.py
manasagowri: bar is 3 letters and that multipied by 20000 will actually be 60000 letters written that's why we are reading 60000 back to compare

suites/reef/rbd/tier-2_rbd_regression.yaml
manasagowri: Added test to pacific and quincy suite as well.

ceph/rbd/workflows/rbd_python.py
manasagowri: Done.

ceph/rbd/workflows/rbd_python.py
manasagowri: Done.

sunilangadi2 (approved):
changes LGTM.

PR #3179, opened 2023-11-27
Adding test - Concurrent writes with exclusive lock. persistent write cache

manasagowri:
Automation of test case - https://polarion.engineering.redhat.com/polarion/#/project/CEPH/workitem?id=CEPH-83574720 - Concurrent writes to check exclusive lock feature with persistent cache enables via same or different IO tools in SSD mode

Test Steps :

Enable the flags
write data to the same image concurrently
Make sure 2 or more writes to the image at the same time
E.g. 2 long-running "fio" jobs competing with each other on the same image.

Success Logs - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-BY7541/

### DIFF
--- a/suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
+++ b/suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
@@ -62,6 +62,7 @@ tests:
         install_packages:
           - ceph-common
           - fio
+          - rbd-nbd
         copy_admin_keyring: true
       desc: Configure client node
       destroy-cluster: false
@@ -410,3 +411,28 @@ tests:
           desc: RBD Persistent Cache invalidate image level
           module: test_rbd_persistent_writeback_cache_invalidate.py
           name: PWL cache path validation at image level
+
+  - test:
+      abort-on-fail: true
+      config:
+        level: client                        # PWL at client
+        cache_file_size: 1073741824          # 1 GB
+        rbd_persistent_cache_mode: ssd       # "ssd" or "rwl" on pmem device
+        client: node10
+        drive: /dev/nvme0n1
+        cleanup: true
+        validate_exclusive_lock: true
+        rep-pool-only: True
+        rep_pool_config:
+          pool: pool1
+          image: image1
+          size: 10G
+        fio:
+          image_name: image1
+          pool_name: pool1
+          runtime: 120
+      desc: Validate concurrent writes to exclusive lock feature with persistent cache enabled
+      destroy-cluster: false
+      module: test_concurrent_write_pwl_cache.py
+      name: PWL cache creation with exclusive lock and concurrent writes
+      polarion-id: CEPH-83574720
--- a/suites/reef/rbd/tier-3_rbd_persistent_write_back_cache.yaml
+++ b/suites/reef/rbd/tier-3_rbd_persistent_write_back_cache.yaml
@@ -58,10 +58,11 @@ tests:
       config:
         command: add
         id: client.1
-        node: node6
+        node: node4
         install_packages:
           - ceph-common
           - fio
+          - rbd-nbd
         copy_admin_keyring: true
       desc: Configure client node
       destroy-cluster: false
@@ -410,3 +411,31 @@ tests:
           desc: RBD Persistent Cache invalidate image level
           module: test_rbd_persistent_writeback_cache_invalidate.py
           name: PWL cache path validation at image level
+
+  - test:
+      abort-on-fail: true
+      config:
+        levels:
+          - client                           # PWL at client
+          - pool
+          - image
+        cache_file_size: 1073741824          # 1 GB
+        rbd_persistent_cache_mode: ssd       # "ssd" or "rwl" on pmem device
+        client: node10
+        drive: /dev/nvme0n1
+        cleanup: true
+        validate_exclusive_lock: true
+        rep-pool-only: True
+        rep_pool_config:
+          pool: pool1
+          image: image1
+          size: 10G
+        fio:
+          image_name: image1
+          pool_name: pool1
+          runtime: 120
+      desc: Validate concurrent writes to exclusive lock feature with persistent cache enabled
+      destroy-cluster: false
+      module: test_concurrent_write_pwl_cache.py
+      name: PWL cache creation with exclusive lock and concurrent writes
+      polarion-id: CEPH-83574720
--- a/tests/rbd/krbd_io_handler.py
+++ b/tests/rbd/krbd_io_handler.py
@@ -88,6 +88,8 @@ def krbd_io_handler(**kw):
             image_index = config["image_spec"].index(image_spec)
 
             if operations.get("mount"):
+                if not device_names:
+                    device_names = config.get("device_names")
                 if not config.get("skip_mkfs"):
                     run_mkfs(
                         client_node=rbd.ceph_client,
@@ -120,11 +122,16 @@ def krbd_io_handler(**kw):
                         size=config.get("file_size", "100M"),
                     )
                 else:
-                    run_fio(
-                        client_node=rbd.ceph_client,
-                        device_name=device_names[-1],
-                        runtime=config.get("runtime", 30),
-                    )
+                    fio_args = {
+                        "client_node": rbd.ceph_client,
+                        "device_name": device_names[-1],
+                    }
+                    if config.get("io_size"):
+                        fio_args.update({"size": config.get("io_size")})
+                    else:
+                        fio_args.update({"runtime": config.get("runtime", 30)})
+
+                    run_fio(**fio_args)
 
             if operations.get("fsck"):
                 if rbd.exec_cmd(cmd=f"fsck -n {device_names[-1]}"):
@@ -143,7 +150,7 @@ def krbd_io_handler(**kw):
                         )
                         return_flag = 1
                     if operations.get("device_map"):
-                        if rbd.device_unmap(
+                        if rbd.device_map(
                             "unmap",
                             f"{pool_name}/{image_name}",
                             config.get("device_type", "nbd"),
--- a/tests/rbd/rbd_peristent_writeback_cache.py
+++ b/tests/rbd/rbd_peristent_writeback_cache.py
@@ -218,9 +218,59 @@ def get_entity_level(config):
     return config_level, entity
 
 
-def fio_ready(config, client):
+def fio_ready(config, client, **kw):
     """Method to prepare FIO config args."""
     fio_args = config["fio"]
     fio_args["client_node"] = client
     fio_args["long_running"] = True
+
+    if kw:
+        fio_args.update(kw)
     return fio_args
+
+
+def unmount(rbd, mount_point):
+    """Unmount the mount point
+
+    Args:
+        rbd: RBD object.
+        mount_point: directory where image is mounted
+    """
+    flag = 0
+    umount_cmd = f"umount -f {mount_point}"
+    if rbd.exec_cmd(cmd=umount_cmd, sudo=True):
+        log.error(f"Umount failed for {mount_point}")
+        flag = 1
+
+    if rbd.exec_cmd(cmd=f"rm -rf {mount_point}", sudo=True):
+        log.error(f"Remove dir failed for {mount_point}")
+        flag = 1
+
+    return flag
+
+
+def device_cleanup(rbd):
+    """Unmount and unmap the device.
+
+    Args:
+        rbd: RBD object
+    """
+    cmd = "lsblk --include 43 --json"
+    out = rbd.exec_cmd(cmd=cmd, sudo=True, output=True)
+    if out:
+        rbd_devices = loads(out)
+
+        for devices in rbd_devices.get("blockdevices"):
+            device_name = f"/dev/{devices.get('name')}"
+            mount_points = devices.get("mountpoints")
+            for mount_point in mount_points:
+                unmount(rbd=rbd, mount_point=mount_point)
+            out_2 = rbd.device_map(
+                operation="unmap",
+                image_spec=f"{device_name}",
+                device_type="nbd",
+                encryption_config=[],
+                long_running=True,
+            )
+            if out_2:
+                log.error(f"RBD unmap failed for {device_name} ")
--- a/tests/rbd/rbd_utils.py
+++ b/tests/rbd/rbd_utils.py
@@ -89,6 +89,7 @@ class Rbd:
                 )
                 return out
 
+            err = ""
             out, err = node.exec_command(
                 sudo=True,
                 cmd=cmd,
@@ -814,7 +815,7 @@ class Rbd:
         cmd = f"rbd encryption format {image_spec} {encryption_type} {passphrase}"
         return self.exec_cmd(cmd=cmd)
 
-    def device_map(self, operation, image_spec, device_type, encryption_config):
+    def device_map(self, operation, image_spec, device_type, encryption_config, **kw):
         """Perform rbd device map operation using given parameters
 
         Args:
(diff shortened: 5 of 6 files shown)


conf/reef/rbd/4-node-cluster-with-1-client.yaml
```diff
@@ -39,3 +39,5 @@ globals:
       node4:
         role:
           - client
+        no-of-volumes: 3
```
sunilangadi2, line 42: why we need volumes on client node?
please add comment.

suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
```diff
@@ -410,3 +411,28 @@ tests:
...
           module: test_rbd_persistent_writeback_cache_invalidate.py
           name: PWL cache path validation at image level
+
+  - test:
+      abort-on-fail: true
+      config:
+        level: client                        # PWL at client
+        cache_file_size: 1073741824          # 1 GB
+        rbd_persistent_cache_mode: ssd       # "ssd" or "rwl" on pmem device
+        client: node5
```
sunilangadi2, line 421: as we face some issue with passing `client` node explicitly, it's good to get that node through module code.

suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
```diff
@@ -410,3 +411,28 @@ tests:
...
+        drive: /dev/nvme0n1
+        cleanup: true
+        validate_exclusive_lock: true
+        rep-pool-only: True
+        rep_pool_config:
+          pool: pool1
+          image: image1
+          size: 10G
+        fio:
```
sunilangadi2, line 430: already provided config for `pool name` and `image name` in above rep_pool_config can we use the same.

suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
```diff
@@ -410,3 +411,28 @@ tests:
...
+          image_name: image1
+          pool_name: pool1
+          runtime: 120
+      desc: Validate concurrent writes to exclusive lock feature with persistent cache enabled
+      destroy-cluster: false
+      module: test_concurrent_writes_rbd_persistent_writeback_cache.py
```
sunilangadi2, line 436: module name is very big, it's good to keep small e.g
`test_concurrent_write_pwl_cache.py`

suites/reef/rbd/tier-3_rbd_persistent_write_back_cache.yaml
```diff
@@ -58,10 +58,11 @@ tests:
       config:
         command: add
         id: client.1
-        node: node6
+        node: node4
```
sunilangadi2, line 61: please get the client node directly in module code wise.

suites/reef/rbd/tier-3_rbd_persistent_write_back_cache.yaml
```diff
@@ -410,3 +411,31 @@ tests:
...
+  - test:
+      abort-on-fail: true
+      config:
+        levels: 
+          - client                           # PWL at client
+          - pool
+          - image
+        cache_file_size: 1073741824          # 1 GB
+        rbd_persistent_cache_mode: ssd       # "ssd" or "rwl" on pmem device
+        client: node4
```
sunilangadi2, line 424: same here.

suites/reef/rbd/tier-3_rbd_persistent_write_back_cache.yaml
```diff
@@ -410,3 +411,31 @@ tests:
...
+        drive: /dev/vdd
+        cleanup: true
+        validate_exclusive_lock: true
+        rep-pool-only: True
+        rep_pool_config:
+          pool: pool1
+          image: image1
+          size: 10G
+        fio:
+          image_name: image1
```
sunilangadi2, line 434: same here.

tests/rbd/krbd_io_handler.py
```diff
@@ -143,7 +150,7 @@ def krbd_io_handler(**kw):
                         )
                         return_flag = 1
                     if operations.get("device_map"):
-                        if rbd.device_unmap(
+                        if rbd.device_map(
```
sunilangadi2, line 153: but still command here is `unmap`?

tests/rbd/rbd_peristent_writeback_cache.py
```diff
@@ -218,9 +218,16 @@ def get_entity_level(config):
...
-def fio_ready(config, client):
+def fio_ready(config, client, **kw):
     """Method to prepare FIO config args."""
     fio_args = config["fio"]
     fio_args["client_node"] = client
     fio_args["long_running"] = True
+
+    if kw:
+        fio_args.update(kw)
+    # if kw.get("test_name"):
```
sunilangadi2, line 229: why commented code here?

tests/rbd/rbd_utils.py
```diff
@@ -89,6 +89,7 @@ def exec_cmd(self, **kw):
                 )
                 return out
 
+            err = ""
```
sunilangadi2, line 92: we can directly assign Is it?

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
```diff
@@ -0,0 +1,231 @@
+"""RBD Persistent write back cache, Test concurrent writes to same image
+
+Test Case Covered:
+CEPH-83574720 - 
+Concurrent writes to check exclusive lock feature with persistent cache enables via same or different IO tools in SSD mode
+
+Steps : 
+1) Setup persistent write cache for an image
+2) Write data to the same image concurrently using two or more long running fio jobs
```
sunilangadi2, line 10: Does it on same mount point or different one?

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
```diff
@@ -0,0 +1,231 @@
...
+)
+from tests.rbd.rbd_utils import initial_rbd_config
+from utility.log import Log
+from utility.utils import run_fio
+
+log = Log(__name__)
+
+
+def unmount(rbd, mount_point):
+    """Unmount the mount point"""
```
sunilangadi2, line 40: module `args` details missing, It's good to add this method in `finally` block.

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
```diff
@@ -0,0 +1,231 @@
...
+
+    if rbd.exec_cmd(cmd=f"rm -rf {mount_point}", sudo=True):
+        log.error(f"Remove dir failed for {mount_point}")
+        flag = 1
+
+    return flag
+
+
+def device_cleanup(rbd):
+    """Unmount and unmap the device."""
```
sunilangadi2, line 55: same here. also return values.

sunilangadi2 (changes requested):

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
```diff
@@ -0,0 +1,232 @@
...
+                log.error(
+                    "Files in mount point 1 do not match with files in mount point 2"
+                )
+                return 1
+        else:
+            log.error("Error while fetching files in the mount points")
+            return 1
+    except Exception as e:
+        log.error(f"Validation of concurrent writes failed with error: {e}")
+        raise Exception(f"Validation of concurrent writes failed with error: {e}")
```
sunilangadi2, line 183: Does it redundant `exception` twice?

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
```diff
@@ -0,0 +1,232 @@
...
+        out, err = cache.rbd.exec_cmd(
+            cmd=f"ls -l {mount_path_1.rsplit('/', 1)[0]}", all=True
+        )
+        out_2, err_2 = cache.rbd.exec_cmd(
+            cmd=f"ls -l {mount_path_2.rsplit('/', 1)[0]}", all=True
+        )
+        if not (err or err_2):
+            log.info(f"Files in first mount point: {out}")
+            log.info(f"Files in second mount point: {out_2}")
+            if out == out_2 and "file_1" in out and "file_2" in out:
```
sunilangadi2, line 169: `file_2` should in `out_2` Is it?

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
```diff
@@ -0,0 +1,232 @@
...
+                log.info(
+                    "All files written by both mount points are present in both the mount points"
+                )
+            else:
+                log.error(
+                    "Files in mount point 1 do not match with files in mount point 2"
```
sunilangadi2, line 175: It need not to match right Does `error` message correct?

sunilangadi2 (changes requested):

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
```diff
@@ -0,0 +1,232 @@
...
+
+Test Case Covered:
+CEPH-83574720 -
+Concurrent writes to check exclusive lock feature with persistent cache enables
+via same or different IO tools in SSD mode
+
+Steps :
+1) Setup persistent write cache for an image
+2) Write data to the same image concurrently using two or more long running fio jobs
+3) Verify that data written is not corrupted using rbd status command
```
sunilangadi2, line 11: How are we validating data corruption here?
can we use this command
`fio --name=mytest --ioengine=sync --rw=randwrite --bs=4k --size=1G --verify=crc32 --verify_fatal=1`

sunilangadi2 (changes requested):

conf/reef/rbd/4-node-cluster-with-1-client.yaml
manasagowri: Added that for testing purpose. Removing it now.

suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
manasagowri: Since persistent cache tests are running only in baremetal mero config, its fine to retain it here for now. We can change it when we are doing SDK implementation for persistent cache if required. Since this is unnecessary here for this test case. 
In your test case, dm cache since it was running on both cephci and baremetal we had to do the change.

suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
manasagowri: Its just to avoid few lines of code in copying the pool and image to fio parameters. No harm in retaining it this way I feel

suites/quincy/rbd/tier-3_rbd_persistent_write_back_cache.yaml
manasagowri: done.

suites/reef/rbd/tier-3_rbd_persistent_write_back_cache.yaml
manasagowri: Since persistent cache tests are running only in baremetal mero config, its fine to retain it here for now. We can change it when we are doing SDK implementation for persistent cache if required. Since this is unnecessary here for this test case.
In your test case, dm cache since it was running on both cephci and baremetal we had to do the change.

suites/reef/rbd/tier-3_rbd_persistent_write_back_cache.yaml
manasagowri: Since persistent cache tests are running only in baremetal mero config, its fine to retain it here for now. We can change it when we are doing SDK implementation for persistent cache if required. Since this is unnecessary here for this test case.
In your test case, dm cache since it was running on both cephci and baremetal we had to do the change.

suites/reef/rbd/tier-3_rbd_persistent_write_back_cache.yaml
manasagowri: Since persistent cache tests are running only in baremetal mero config, its fine to retain it here for now. We can change it when we are doing SDK implementation for persistent cache if required. Since this is unnecessary here for this test case.
In your test case, dm cache since it was running on both cephci and baremetal we had to do the change.

tests/rbd/krbd_io_handler.py
manasagowri: yes, you can see i am passing parameter  "unmap" to input of method device_map. 
Based on parameter passed "map" or "unmap" it will do the required action.

tests/rbd/rbd_peristent_writeback_cache.py
manasagowri: removed.

tests/rbd/rbd_utils.py
manasagowri: When it is a long running command only out is returned, and if user gives all=True we are returning out and err variables both. In this case it will throw error that err variable is not defined. That's why i am just assigning it empty value.

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
manasagowri: mentioned in detail now.

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
manasagowri: args added, I want code to be non clumsy so created a different method. No harm I guess.

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
manasagowri: both methods don't return anything. I have updated args.

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
manasagowri: In line 181 I am catching the exception and writing meaningful log.error and then raising the exception in line 183. This is implemented this way because all persistent cache tests are raising exceptions instead of returning 1 or 0. That's how its designed and I don't want to change the design now. 

We already discussed the same review comment as part of cache invalidate PR - https://github.com/red-hat-storage/cephci/pull/3141#discussion_r1395737404

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
manasagowri: No, both mount points should contain both the files, that's the verification to say that data written in both mount points are in sync with each other and also that they have been stored in the corresponding rbd image.
This is what we discussed right?

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
manasagowri: It should match, see above comment

tests/rbd/test_concurrent_writes_rbd_persistent_writeback_cache.py
manasagowri: I will change the sentence here, we are only validating that data is written to both mount points and both data are present. Data corruption verification is not required afaik.

tests/rbd/test_concurrent_write_pwl_cache.py
```diff
@@ -0,0 +1,243 @@
...
+    get_entity_level,
+)
+from tests.rbd.rbd_utils import initial_rbd_config
+from utility.log import Log
+from utility.utils import run_fio
+
+log = Log(__name__)
+
+
+def unmount(rbd, mount_point):
```
sunilangadi2, line 41: clean up code is already there here https://github.com/red-hat-storage/cephci/blob/6692c56f9e02a83a84527e24b922e8a02e45d296/tests/rbd/rbd_peristent_writeback_cache.py#L33
can we reuse the same.

manasagowri: That cleanup does only persistent cache settings cleanup. It doesn't unmount. So it can't be used.

sunilangadi2 (approved):
changes LGTM.

PR #3191, opened 2023-11-29
Adding test to compare performance with or without pwl cache

manasagowri:
Automation of test - CEPH-83574711 - verify the performance with and without write back cache in SSD mode
Steps :
1) Create a pool and an image.
2) Enable client, pool and image level pwl cache on the pool.
3) Mount image and write some amount of data
4) Note time taken for data write
5) Disable pwl cache setting
6) Write same amount of data as before
7) Note time taken for data write
8) Verify that pwl cache enabled image takes lesser time to write data

Success Logs - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-2JCDAC/

### DIFF
--- a/tests/rbd/test_performance_pwl_cache.py
+++ b/tests/rbd/test_performance_pwl_cache.py
@@ -1,0 +1,180 @@
+"""RBD Persistent write back cache, Test performance with/without cache enabled
+Test Case Covered:
+CEPH-83574711 - verify the performance with and without write back cache in SSD mode
+Steps :
+1) Create a pool and an image.
+2) Enable client, pool and image level pwl cache on the pool.
+3) Mount image and write some amount of data
+4) Note time taken for data write
+5) Disable pwl cache setting
+6) Write same amount of data as before
+7) Note time taken for data write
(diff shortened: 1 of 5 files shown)


tests/rbd/test_performance_pwl_cache.py
```diff
@@ -0,0 +1,175 @@
...
+    Args:
+        ceph_cluster: ceph cluster object
+        **kw: test parameters
+    Pre-requisites :
+        - need client node with ceph-common package, conf and keyring files
+        - FIO should be installed on the client.
+    """
+    log.info("Running PWL....")
+    log.info(
+        "Running test - Test performance with/without cache enabled. persistent write cache"
```
sunilangadi2, line 148: please add polarion id here in `info`

sunilangadi2 (changes requested):
Since it is performance testing can we pass FIO data to fille in test suite config, so in future we can increase this value based on cluster config,
also right now it's writing only 1GB can we increase it to more than 4GB?
as this much size VM's also support.

sunilangadi2 (approved):
changes LGTM.

PR #3194, opened 2023-11-29
[test-fix] Add fix for OSD deployed with limit filter spec test failure

vinpapnoi:
# Description

Moved the test case "Verify OSD deployed with limit filter spec" from the suite "tier-1_service_apply_spec.yaml" to "tier-1_cephadm-clients.yaml"

Additional changes have also been made to "tier-1_service_apply_spec.yaml" to fix MDS deployment failures.

### DIFF
--- a/suites/quincy/cephadm/tier-1_cephadm-clients.yaml
+++ b/suites/quincy/cephadm/tier-1_cephadm-clients.yaml
@@ -134,6 +134,7 @@ tests:
                 data_devices:
                   all: "true"                         # boolean as string
                 encrypted: "true"                     # boolean as string
+                unmanaged: "true"                     # boolean as string
         - config:
             command: shell
             args:                 # sleep to get all services deployed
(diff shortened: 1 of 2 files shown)


vinpapnoi:
Logs : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-WRMG1P

suites/quincy/cephadm/tier-1_cephadm-clients.yaml
```diff
@@ -134,6 +134,7 @@ tests:
                 data_devices:
                   all: "true"                         # boolean as string
                 encrypted: "true"                     # boolean as string
+                unmanaged: "true"                     # boolean as string
```
vamahaja, line 137: Why this change required?

vamahaja (changes requested):

suites/quincy/cephadm/tier-1_cephadm-clients.yaml
vinpapnoi: So that when any new osd spec is deployed we would not have to manually zap devices which is a lengthy time consuming process. Instead it will automatically assign the OSD's according to the new spec file added below.

vinpapnoi: Also, if we do not apply the unmanaged flag to true then any OSD's if removed would get automatically redeployed for the same service. And any other spec file would not deploy OSD's specific to itself.

tintumathew10 (approved):
LGTM

PR #3200, opened 2023-11-30
[Lib] Add temp yaml file creation through config

adityaramteke:
Add a method to create a .yaml temp file from suite config

### DIFF
--- a/cli/utilities/utils.py
+++ b/cli/utilities/utils.py
@@ -2,6 +2,7 @@ import os
 import random
 import re
 import string
+import tempfile
 from datetime import datetime
 from json import loads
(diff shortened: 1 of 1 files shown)


cli/utilities/utils.py
```diff
@@ -855,3 +856,20 @@ def get_pid_limit(node, service):
         return None
     pid_limit = re.findall(r"-?\d+", pid_limit[0])[0]
     return pid_limit
+
+
+def create_yaml_file(node, config):
```
vamahaja, line 861: `create_yaml_config`

adityaramteke: Done, updated.

PR #3202, opened 2023-12-04
Stretch_cluster.py: Commenting test due to active bug: 2249962

pdhiran:
1. commenting out the test to verify access to cluster during netsplits in stretch mode. bug :: https://bugzilla.redhat.com/show_bug.cgi?id=2249962 
2. Increasing the number of retries to add host into maintenance mode.
3. Increased timeout for objects to be shown in the ceph df status post IO's.
4. Updated the method for setting the noautoscale flag on the cluster. There exists a bug in 6.1z3 : https://bugzilla.redhat.com/show_bug.cgi?id=2252788

### DIFF
--- a/ceph/rados/core_workflows.py
+++ b/ceph/rados/core_workflows.py
@@ -2828,3 +2828,52 @@ class RadosOrchestrator:
             return None
         log.info(f"The inconsistent object is created in the pg{pg_id}")
         return pg_id
+
+    def set_noautoscale_flag(self, retry: int = 15) -> bool:
+        """
+        sets/Unsets the noautoscale flag on the cluster
+        Args:
+            retry: max number of retries to set the noautoscale flag on the cluster
+        Returns:
+            True -> noautoscale flag set successfully
+            False -> noautoscale flag not set
+        """
+        log.debug("Setting the noautoscale flag on the cluster")
+        iteration = 0
+        flag = True
+
+        while iteration <= retry:
+            iteration += 1
+            cmd = "ceph osd pool set noautoscale"
+            try:
+                out, _ = self.client.exec_command(cmd=cmd, sudo=True, timeout=600)
+                log.debug(f"o/p of maintenance enter cmd : {out}")
+            except Exception as e:
+                log.debug(f"Exception hit, but was expected; {e}")
+
+            log.debug(
+                "sleeping for 60 seconds before checking status of autoscale flag"
+            )
+            time.sleep(60)
+
+            cmd = "ceph osd pool autoscale-status"
+            pool_status = self.run_ceph_command(cmd=cmd, timeout=600)
+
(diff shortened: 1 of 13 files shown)


ceph/rados/core_workflows.py
```diff
@@ -2740,3 +2740,52 @@ def get_host_object(self, hostname):
 
         log.error(f"Could not find host object for host {hostname}")
         return None
+
+    def set_autoscale_flag(self, retry: int = 15) -> bool:
```
harshkumarRH, line 2744: Can we please consider changing the function name to `set_noautoscale_flag`?

harshkumarRH, line 2744: There exists a module to take care of pg_autoscaler `configure_pg_autoscaler(self, **kwargs)`, any possibility that both can be merged?

tests/rados/test_pg_autoscale_flag.py
```diff
@@ -51,13 +51,9 @@ def run(ceph_cluster, **kw):
 
     # Getting the autoscale configurations after setting the flag
     # all the pools should have autoscale set to off
-    cmd = "ceph osd pool autoscale-status"
-    pool_status = rados_obj.run_ceph_command(cmd=cmd)
-
-    for entry in pool_status:
-        if entry["pg_autoscale_mode"] == "on":
-            log.error(f"Pg autoscaler not turned off for pool : {entry['pool_name']}")
-            return 1
```
harshkumarRH, line 60: Line 45-53 can also be removed as it is now part of `set_autoscale_flag` common module

ceph/rados/core_workflows.py
pdhiran: the functionality of both the modules is different. 
`configure_pg_autoscaler` deals with the pool level settings to change the autoscale state on the pool.

This method is a global level flag, like noout, noscrub etc, that acts on entire cluster. I do not feel that the two should be merged into one function.

harshkumarRH: Any reason as to why we are going with retries instead of direct timeouts?

SrinivasaBharath: +1

pdhiran: just time does not give a correct view on how many times a command was executed to deem something does not work. Count gives much better idea. 

Nothing too technical about the choice of retries vs just timeout.

tests/rados/test_pg_autoscale_flag.py
pdhiran: done. removed all invalid lines

SrinivasaBharath:
Apart from the re-base issue, PR is good to merge. Please re-base the code before merging.

neha-gangadhar (approved):
lgtm

PR #3212, opened 2023-12-07
[test-fix] Add fix for cephadm log spam failure

vinpapnoi:
# Description

Before the bug fix, the spam logs would be printed in cephadm logs every time the devices get refreshed. So if the command `ceph orch device ls --refresh` was ran repeatedly, the spam logs would keep getting filled without the devices actually changing.

With the bug fix, the logs would be triggered only once if the devices have changed.

The change here is to first run the device refresh command to populate the logs and get the count of those logs. Then running the device refresh command again and comparing the count of logs.

### DIFF
--- a/suites/reef/cephadm/tier1-service-apply-spec.yaml
+++ b/suites/reef/cephadm/tier1-service-apply-spec.yaml
@@ -263,3 +263,17 @@ tests:
       config:
         nodes:
           - node2
+  - test:
+      name: Verify spam not present in cephadm log file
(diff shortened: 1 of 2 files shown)


vinpapnoi:
Test Run Logs:
Quincy : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-WFHU1D
Reef: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-F5SVAW

vamahaja (commented):
LGTM

tintumathew10 (approved):

PR #3219, opened 2023-12-10
Test to cover immutable object cache test with cluster operations

sunilangadi2:
# Description
Automated test case:
https://polarion.engineering.redhat.com/polarion/#/project/CEPH/workitem?id=CEPH-83574132

**Test Case Flow :**
    1)Create multiple rbd pool and multiple images using rbd commands
    2)Write some data to the images using FIO or RBD bench
    3)Create multiple clones images in different pools from the multiple parent image created
    4)Read the cloned images of different pools in parallel and check the cache status
    5) Restart the Mon, OSD, and cluster target and parallelly run the IO and cache status
    6) Remove mon and add mon back to cluster with parallelly run the IO and cache status
    8)  Check ceph health 
    
  Test Result:
  http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-KL6LY5/
    
<details>

</details>

### DIFF
--- a/suites/reef/rbd/tier-2_rbd_cache_scenarios.yaml
+++ b/suites/reef/rbd/tier-2_rbd_cache_scenarios.yaml
@@ -108,3 +108,22 @@ tests:
       module: test_rbd_immutable_cache.py
       name: Immutable object cache creation
       polarion-id: CEPH-83574134
+
+  - test:
+      abort-on-fail: true
+      config:
+        rep_pool_config:
+          num_pools: 2
+          num_images: 5
+          size: 4G
+        ec_pool_config:
+          num_pools: 2
+          num_images: 5
+          size: 4G
+        fio:
+          size: 100M
+      desc: validate immutable object cache feature with cluster operation parallel
+      destroy-cluster: false
+      module: test_rbd_immutable_cache_cluster_operations.py
+      name: Immutable object cache creation with cluster operation
+      polarion-id: CEPH-83574132
--- a/tests/rbd/test_rbd_immutable_cache_cluster_operations.py
+++ b/tests/rbd/test_rbd_immutable_cache_cluster_operations.py
@@ -1,0 +1,273 @@
+"""Perform cluster-related operations along with
+    immutable object cache test parallel.
+
+    Pre-requisites :
+    We need atleast one client node with ceph-common and fio packages,
+    conf and keyring files
+
+    Test cases covered -
+    CEPH-83574132 - Immutable object cache with cluster operations.
+
+    Test Case Flow -
+
+    1)Create multiple rbd pool and multiple images using rbd commands
+    2)Write some data to the images using FIO
+    3)Create multiple clones images in different pools from the multiple parent image created
+    4)Read the cloned images of different pools in parallel and check the cache status
+    5) Restart the Mon, OSD, and cluster target and parallelly run the IO and cache status
+    6) Remove mon and add mon back to cluster with parallelly run the IO and cache status
+    8) check ceph health status
+    9) Perform test on both Replicated and EC pool
+"""
+import time
+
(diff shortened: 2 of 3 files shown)


suites/reef/rbd/tier-2_rbd_cache_scenarios.yaml
```diff
@@ -108,3 +108,20 @@ tests:
       module: test_rbd_immutable_cache.py
       name: Immutable object cache creation
       polarion-id: CEPH-83574134
+
+  - test:
+      abort-on-fail: true
+      config:
+        rep_pool_config:
+          num_pools: 2
+          num_images: 5
```
manasagowri, line 117: default size of images created is 1GB is that enough for scale testing?

tests/rbd/test_rbd_immutable_cache_cluster_operations.py
```diff
@@ -0,0 +1,273 @@
...
+        "io_type": "write",
+        "verify": "crc32",
+        "verify_fatal": 1,
+    }
+
+    if kw:
+        fio_args.update(kw)
+
+    try:
+        run_fio(**fio_args)
```
manasagowri, line 78: don't you think its better to run IOs on mounted path instead of directly on images?

tests/rbd/test_rbd_immutable_cache_cluster_operations.py
```diff
@@ -0,0 +1,273 @@
...
+        if out or err and "100% complete" not in err:
+            log.error(f"Snapshot creation failed for {clone_snap_spec}")
+            return 1
+
+        # Read IO from cloned images to get cache files
+        clone = clone_spec.split("/")[1]
+        fio_args["image_name"] = clone
+        fio_args["io_type"] = "read"
+
+        run_fio(**fio_args)
```
manasagowri, line 100: same comment here.

manasagowri (changes requested):

suites/reef/rbd/tier-2_rbd_cache_scenarios.yaml
sunilangadi2: Added `size to 4G`

tests/rbd/test_rbd_immutable_cache_cluster_operations.py
sunilangadi2: No, Immutable cache files are getting created on when user performs IO operation on cloned images so we need to run IO on cloned images. hence running IO's directly on images here.

tests/rbd/test_rbd_immutable_cache_cluster_operations.py
sunilangadi2: same.

tests/rbd/test_rbd_immutable_cache_cluster_operations.py
manasagowri: But when would a customer run IOs directly on an image? It's always by mounting right? So how does one use immutable cache in customer scenario?

manasagowri (changes requested):

tests/rbd/test_rbd_immutable_cache_cluster_operations.py
sunilangadi2: yes but user can use multiple ways of using image to write data mounting is oneway,
even if we mount and run also internally it's rbd image itself
also in test case it didn't mentioned about it may be we can take this further 
but for users ways are more only so in our testing i feel it's good to test with IO on image directly.

sunilangadi2: @manasagowri https://issues.redhat.com/browse/RHCEPHQE-12402
created task to explore customer-based test scenarios to cover more test coverage.

manasagowri (approved):

PR #3227, opened 2023-12-13
[tfa-fix] Add fix for replace OSD test case failure

vinpapnoi:
# Description

This fix is for the failure observed in TFA for "tier-2-osd-scenarios" suite.
The error observed was :
`json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)`


tintumathew10 (commented):
LGTM

vamahaja (approved):

PR #3245, opened 2023-12-19
Adding privatekey argument so that we can utilise the ssh key 

shivamdurgbuns:
# Description

Adding private-key argument so that we can utilize the ssh key option of paramiko to connect to the server using the ssh key

### DIFF
--- a/ceph/ceph.py
+++ b/ceph/ceph.py
@@ -1236,6 +1236,7 @@ class SSHConnectionManager(object):
         self.__transport = None
         self.__outage_start_time = None
         self.outage_timeout = datetime.timedelta(seconds=outage_timeout)
+        self.__private_key_file_path = private_key_file_path
 
     @property
     def client(self):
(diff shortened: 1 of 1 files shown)


ceph/ceph.py
```diff
@@ -1253,13 +1254,22 @@ def __connect(self):
...
-                    look_for_keys=self.look_for_keys,
-                    allow_agent=False,
-                )
+                if self.look_for_keys:
+                    self.__client.connect(
+                        self.ip_address,
+                        username=self.username,
+                        look_for_keys=self.look_for_keys,
+                        allow_agent=False,
+                        key_file_name=self.__private_key_file_path
```
vamahaja, line 1263: By default `__private_key_file_path` set to `None`. In this case don't need this if.. else block. first block should work.

shivamdurgbuns: I have removed the password argument as well in the first block.

shivamdurgbuns: Updated the changes, works without passing a password as well (when just ssh key is passed)

vamahaja (approved):

PR #3302, opened 2024-01-12
[NFS-Ganesha][Setup][Core-Dump]addition

MohitBis:
[NFS-Ganesha][Setup][Core-Dump]addition

Problem:

To add coredump feature, the node details are required in the methods defined under nfs_operations. But currently, the node ips alone are being passed to these methods, hence any operations on nodes won't be possible with current implementation

Solution:
Pass the ceph_cluster variable to the setup_nfs_cluster method. This will enable the node specific and other related operations to be carried out, thus resolving the problem.

The current pr is specific to the tests and all the existing tests are modified in passing the ceph_cluster variable to the setup_nfs_cluster method

### DIFF
--- a/cli/utilities/utils.py
+++ b/cli/utilities/utils.py
@@ -627,8 +627,12 @@ def check_coredump_generated(node, coredump_path, created_after):
     cmd = f"ls -Art {coredump_path} | tail -n 1"
     file_name, _ = node.exec_command(cmd=cmd, sudo=True)
 
+    # If the path is empty / not created, return False as no coredump has been generated
+    if file_name == "":
+        return False
+
     # Get the file creation time
-    cmd = f"stat -c '%w' {file_name}"
+    cmd = f"stat -c '%w' {coredump_path}/{file_name}"
     created_time, _ = node.exec_command(cmd=cmd, sudo=True)
 
     # Remove timezone and fractional seconds from time
@@ -638,8 +642,8 @@ def check_coredump_generated(node, coredump_path, created_after):
 
     # Verify if the file is created after the given time
     if created_time > created_after:
-        return False
-    return True
+        return True
+    return False
 
 
 def create_files(client, mount_point, file_count):
--- a/tests/nfs/nfs_operations.py
+++ b/tests/nfs/nfs_operations.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from threading import Thread
 from time import sleep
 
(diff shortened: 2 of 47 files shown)


MohitBis:
Logs: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-3H0Y65/

tests/nfs/nfs_operations.py
```diff
@@ -88,6 +104,16 @@ def cleanup_cluster(clients, nfs_mount, nfs_name, nfs_export):
     if not isinstance(clients, list):
         clients = [clients]
 
+    # Check nfs coredump
+    if ceph_cluster_obj:
+        nfs_nodes = ceph_cluster_obj.get_nodes("nfs")
+        coredump_path = "/var/lib/systemd/coredump"
+        for nfs_node in nfs_nodes:
+            if not check_coredump_generated(nfs_node, coredump_path, setup_start_time):
```
pranavprakash20, line 112: It should be """ if check_coredump_generated """ right @MohitBis ?
Beacuse the method `check_coredump_generated` returns False or True based on whether the coredump is generated. Here, as per this logic, the execption is raised when coredump is not generated

tests/nfs/nfs_operations.py
```diff
@@ -88,6 +104,16 @@ def cleanup_cluster(clients, nfs_mount, nfs_name, nfs_export):
...
+                raise NfsCleanupFailed(
+                    f"Coredump grenerated for node {nfs_node.hostname} after steup"
```
pranavprakash20, line 114: rephrase to 
"Coredump generated post execution of the current test case"

cli/utilities/utils.py
```diff
@@ -627,8 +627,12 @@ def check_coredump_generated(node, coredump_path, created_after):
     cmd = f"ls -Art {coredump_path} | tail -n 1"
     file_name, _ = node.exec_command(cmd=cmd, sudo=True)
 
+    # If no coredump path empty
```
pranavprakash20, line 630: Rephrase comment as

" If the path is empty / not created, return True as no coredump has been generated"

pranavprakash20 (changes requested):
Please have a look at the comments.

tests/nfs/nfs_operations.py
MohitBis: Thanks @pranavprakash20 
Updated

tests/nfs/nfs_operations.py
MohitBis: Thanks @pranavprakash20 
Updated

cli/utilities/utils.py
MohitBis: Thanks @pranavprakash20 
Updated

pranavprakash20 (approved):
LGTM

MohitBis:
Latest Log: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-UXR6FX/

msainiRedhat (approved):
LGTM

vdas-redhat (approved):
LGTM

PR #3303, opened 2024-01-12
Modified methods to create pools and images parallely/sequentially as per suite input

manasagowri:
This PR is to support scaling of rbd operations. Currently our workflow method supports creation of pools and images only in sequential fashion.
With this change, we will be able to specify in the test suite config whether the pools and images should be created parallely/sequentially.

Logs for parallel pool and image creation - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-Q7HKZJ/

Logs for sequential pool and image creation - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1HP0XT/

PFB a sample test suite config to use this feature.

> test:
>       desc: Run all image operations in scale
>       module: test_rbd_image_operations.py
>       name: RBD image operations in scale
>       config:
>         rep_pool_config:
>           num_pools: 3
>           num_images: 5
>           size: 4G
>           create_pool_parallely: true
>           create_image_parallely: true
>         ec_pool_config:
>           num_pools: 3
>           num_images: 5
>           size: 4G
>           create_pool_parallely: true
>           create_image_parallely: true

### DIFF
--- a/ceph/rbd/workflows/rbd.py
+++ b/ceph/rbd/workflows/rbd.py
@@ -1,3 +1,4 @@
+from ceph.parallel import parallel
 from ceph.rbd.utils import getdict
 from ceph.rbd.workflows.pool import create_ecpool, create_pool
 from utility.log import Log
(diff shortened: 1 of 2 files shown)


ceph/rbd/workflows/rbd.py
```diff
@@ -36,7 +37,179 @@ def create_snap_and_clone(rbd, snap_spec, clone_spec, **kw):
...
+                        cluster,
+                        rbd,
+                        pool,
+                        pool_config,
+                        image,
+                        image_config_val,
+                        is_ec_pool,
+                        raise_exception=True,
+                    )
+        except Exception:
```
sunilangadi2, line 127: does it good to raise exception here instead of return 1

ceph/rbd/workflows/rbd.py
```diff
@@ -36,7 +37,179 @@ def create_snap_and_clone(rbd, snap_spec, clone_spec, **kw):
...
+        pool_init_conf = {"pool-name": pool, "cluster": cluster}
+        rbd.pool.init(**pool_init_conf)
+
+    if is_ec_pool and create_ecpool(
+        pool=pool_config.get("data_pool"),
+        k_m=pool_config.get("ec-pool-k-m"),
+        profile=pool_config.get("ec_profile", "use_default"),
+        pg_num=pool_config.get("ec_pg_num", 32),
+        pgp_num=pool_config.get("ec_pgp_num", 32),
+        failure_domain=pool_config.get("failure_domain", ""),
```
sunilangadi2, line 170: what is this parameter `failure_domain`?

ceph/rbd/workflows/rbd.py
```diff
@@ -36,7 +37,179 @@ def create_snap_and_clone(rbd, snap_spec, clone_spec, **kw):
...
+        }
+        rbd.pool.init(**pool_init_conf)
+
+    multi_image_config = getdict(pool_config)
+    image_config = {
+        k: v
+        for k, v in multi_image_config.items()
+        if v.get("is_secondary", False) == is_secondary
+    }
+    rc = create_images(
```
sunilangadi2, line 194: why are we creating images inside a single pool?

sunilangadi2: Method name should be `create_single_pool_and_images`

ceph/rbd/workflows/rbd.py
```diff
@@ -45,82 +218,59 @@ def config_rbd_multi_pool(
     client,
     is_secondary=False,
     cluster="ceph",
+    create_pool_parallely=False,
+    create_image_parallely=False,
 ):
-    """ """
+    """
+    Create number of pools as specified in parallel/sequential
```
sunilangadi2, line 225: return and args are missing, please add these in all methods along with examples config.

sunilangadi2 (changes requested):

sunilangadi2:
It's good to add `log.info` at each method where images are creating parallely or sequentially.

ceph/rbd/workflows/rbd.py
manasagowri: You remember in one of your parallel executions as well, we did the same? Because p.spawn does not return any return value? So we have to raise exception?

ceph/rbd/workflows/rbd.py
manasagowri: It's one of the parameters used in ec pool. Please check doc.

ceph/rbd/workflows/rbd.py
manasagowri: config example? i didn't understand what you mean by that?

ceph/rbd/workflows/rbd.py
sunilangadi2: ok got it thanks.

ceph/rbd/workflows/rbd.py
manasagowri: The input parameters are all self explanatory. We don't need to explain it I feel.

manasagowri:
> It's good to add `log.info` at each method where images are creating parallely or sequentially.

Already if you see the logs, it is saying clearly what operations its doing. So log.info is simply redundant. And this PR is just an enhancement to already existing methods to support parallel execution. So we don't need to do these granular changes now.

sunilangadi2 (approved):
some of the comments will be addressed in image operations PR,
changes LGTM.

PR #3325, opened 2024-01-19
Adding this to allow longer fqdn names

shivamdurgbuns:
# Description

Allowing larger fqdn hostnames to bootstrap script after discussion.


AmarnatReddy (approved):
LGTM

PR #3346, opened 2024-01-24
Performing image operations at scale

manasagowri:
This PR is to automate image operations in scale for baremetal configuration. Image operations covered as part of this test are - image list, info, rename, resize, copy, feature enable, disable, run io and check rbd status and remove image and verify the same for multiple pools and images at scale.

Success Logs for sequential execution - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-U7IYOT/ 

Success Logs for parallel execution - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-N78BXY/

### DIFF
--- a/ceph/rbd/workflows/rbd.py
+++ b/ceph/rbd/workflows/rbd.py
@@ -1,5 +1,11 @@
+import datetime
+import json
+from importlib import import_module
+from time import sleep
+
 from ceph.parallel import parallel
-from ceph.rbd.utils import getdict
+from ceph.rbd.utils import convert_size, getdict, random_string
+from ceph.rbd.workflows.krbd_io_handler import krbd_io_handler
 from ceph.rbd.workflows.pool import create_ecpool, create_pool
 from utility.log import Log
 
@@ -139,7 +145,8 @@ def create_single_pool_and_images(
     ceph_version,
     is_ec_pool,
     is_secondary,
-    create_image_parallely,
+    create_image_parallely=False,
+    do_not_create_image=False,
     raise_exception=False,
 ):
     """
@@ -185,27 +192,28 @@ def create_single_pool_and_images(
         }
         rbd.pool.init(**pool_init_conf)
 
-    multi_image_config = getdict(pool_config)
-    image_config = {
-        k: v
-        for k, v in multi_image_config.items()
-        if v.get("is_secondary", False) == is_secondary
-    }
-    rc = create_images(
-        config,
-        cluster,
-        rbd,
-        pool,
-        pool_config,
-        image_config,
-        is_ec_pool,
-        create_image_parallely,
-    )
-    if rc:
-        log.error(f"Error while creating images for pool {pool}")
-        if raise_exception:
-            raise Exception(f"Error while creating images for pool {pool}")
-        return 1
+    if not do_not_create_image:
+        multi_image_config = getdict(pool_config)
+        image_config = {
+            k: v
+            for k, v in multi_image_config.items()
+            if v.get("is_secondary", False) == is_secondary
+        }
+        rc = create_images(
+            config,
+            cluster,
+            rbd,
+            pool,
+            pool_config,
+            image_config,
+            is_ec_pool,
+            create_image_parallely,
+        )
+        if rc:
+            log.error(f"Error while creating images for pool {pool}")
+            if raise_exception:
+                raise Exception(f"Error while creating images for pool {pool}")
+            return 1
     return 0
 
 
@@ -244,6 +252,7 @@ def create_pools_and_images(
                 create_image_parallely,
             )
             if rc:
+                log.error("Multi pool and multi image config failed")
                 return rc
     else:
         try:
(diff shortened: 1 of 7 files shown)


sunilangadi2 (changes requested):
logs URL not working,
Please cover these test suites in `quincy` as well.
will recheck once it's ready to review.

The existing test with `number of pools` and `number of image` config test
we can ignore those test here as it's already exist we just need to increase there
pool and image count in config.

ceph/rbd/workflows/rbd.py
```diff
@@ -264,13 +273,500 @@ def create_pools_and_images(
...
+            if test_ops_parallely:
+                raise Exception(
+                    f"Image info for parameter {k} in not same as input config value {v}"
+                )
+            return 1
+    log.info(f"Image info matches the conf provided in input for image {pool}/{image}")
+    return 0
+
+
+def resize_single_image_and_verify(**kw):
```
sunilangadi2, line 369: we already have some existing methods for resizing single image is it?

ceph/rbd/workflows/rbd.py
```diff
@@ -264,13 +273,500 @@ def create_pools_and_images(
...
+    if test_ops_parallely:
+        with parallel() as p:
+            for image, image_conf in image_config.items():
+                p.spawn(method_obj, image=image, image_conf=image_conf, **kw)
+    else:
+        for image, image_conf in image_config.items():
+            rc = method_obj(image=image, image_conf=image_conf, **kw)
+            if rc:
+                log.error(
+                    f"Test image feature enable and disable failed for {pool}/{image}"
```
sunilangadi2, line 769: error message needs to be image operation better we can log what operation is it as well.

tests/rbd/test_rbd_image_operations.py
```diff
@@ -0,0 +1,295 @@
...
+    return 0
+
+
+def run(**kw):
+    """RBD image operations testing.
+
+    Args:
+        **kw: test data
+    """
+    log.info("Running test <polarion_id> - Testing RBD image operations.")
```
sunilangadi2, line 276: please add poarion id?

tests/rbd/test_rbd_image_operations.py
```diff
@@ -0,0 +1,295 @@
...
+            node=client,
+        )
+        if rc:
+            log.error(f"Error while deleting new pool {new_pool}")
+            if test_ops_parallely:
+                raise Exception(f"Error while deleting new pool {new_pool}")
+            return 1
+
+        if is_ec_pool:
+            rc = exec_cmd(
```
sunilangadi2, line 163: can we create one pool deletion method instead of adding same command to delete multiple pool?

suites/reef/rbd/tier-3_rbd_operations.yaml
```diff
@@ -0,0 +1,95 @@
...
+      module: exec.py
+      name: Install rbd-nbd
+      config:
+        sudo: true
+        commands:
+          - "rm -rf /etc/yum.repos.d/epel*"
+          - "dnf install rbd-nbd -y"
+
+  - test:
+      desc: Run all image operations in scale
```
sunilangadi2, line 76: Please add poarion id for the test?

sunilangadi2 (changes requested):
if we are using same method for multiple images as well, then i think we can take out `single` name used for each method 
we can keep just image related operations as method names
also `suites/quincy/rbd/tier-3_rbd_operations.yaml` can we rename file name to `suites/quincy/rbd/tier-3_rbd_image_operations.yaml`

ceph/rbd/workflows/rbd.py
```diff
@@ -264,13 +273,500 @@ def create_pools_and_images(
...
+                f"All parameters in {image_conf} are not present in {image_info}"
+            )
+        return 1
+    for k, v in req_image_conf.items():
+        value = image_info[k]
+        if "size" in k:
+            value = convert_size(value)
+        if value != v:
+            log.error(
+                f"Image info for parameter {k} in not same as input config value {v}"
```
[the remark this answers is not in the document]
manasagowri: Depending on whether we are running parallely or serially we should raise exception.
If its sequential, then we can just return value instead of exception.

ceph/rbd/workflows/rbd.py
manasagowri: not for multiple pools. and not to verify the resize. This is the only method.

ceph/rbd/workflows/rbd.py
manasagowri: failed.

tests/rbd/test_rbd_image_operations.py
manasagowri: I'm still thinking whether we can create new polarion test case or use any of the existing ones. Once I get clarity, I will add this. Since baremetal pipeline won't run anytime soon, we can merge this for now, later will add polarion id to this. before adding it to pipeline.

tests/rbd/test_rbd_image_operations.py
manasagowri: it's just one function exec_cmd call right. I don't think further customisation is necessary here.

suites/reef/rbd/tier-3_rbd_operations.yaml
manasagowri: see above comment.

ceph/rbd/workflows/rbd.py
```diff
@@ -139,7 +145,8 @@ def create_single_pool_and_images(
     ceph_version,
     is_ec_pool,
     is_secondary,
-    create_image_parallely,
+    create_image_parallely=False,
+    do_not_create_image=False,
```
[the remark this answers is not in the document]
manasagowri: what better name do you suggest? This name felt as relevant as it can get for me.
That's how it works. As you can see its default value is false. If its specified as true while calling then it won't create images.

sunilangadi2 (approved):
changes LGTM further suggestion will be addrssed in some other PR

PR #3353, opened 2024-01-25
[Test] Deploy non-collocated OSDs with wrong db

adityaramteke:
# Description

1. Add couple of new disks for the non-collocated OSD scenario
2. Specify few of the disks as data disk and couple of them as db device in service configuration file without specifying the block_db_size
3. Now apply the specs file
4. You will notice the db devices created with the size as size of single disk/total number of data disks instead of total size of all the db devices divided by number of total data disks

Full logs: http://magna002.ceph.redhat.com/cephci-jenkins/aramteke/automation/cephci-run-COY05Y/

### DIFF
--- a/tests/ceph_volume/test_ceph_volume_wrong_db_size_compute.py
+++ b/tests/ceph_volume/test_ceph_volume_wrong_db_size_compute.py
@@ -1,0 +1,98 @@
+from cephci.utils.configs import get_cloud_credentials, get_configs
+from cli.cephadm.cephadm import CephAdm
+from cli.cloudproviders import CloudProvider
+from cli.cluster.node import Node
+from cli.cluster.volume import Volume
+from cli.exceptions import OperationFailedError
+from cli.ops.devices import get_node_disks
+from cli.utilities.operations import wait_for_osd_daemon_state
+from cli.utilities.utils import WaitUntil, create_yaml_config
+
+
+def _add_node_volumes(nodes, config, size, count):
+    """Create volume in bulk and attach to node
+    Args:
+        nodes (ceph): nodes on volumes to create
+        config (dict): ceph config
+        size (str): size of volume
+        count (int): number of volume
+    """
+    # Get cloud creadentials
+    cloud_type = config.get("cloud-type")
+    cloud_configs = get_cloud_credentials(cloud_type)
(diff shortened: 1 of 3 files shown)


tests/ceph_volume/test_ceph_volume_wrong_db_size_compute.py
```diff
@@ -0,0 +1,112 @@
...
+    # Get configs
+    get_configs()
+    config = kw.get("config")
+
+    # Get the installer and OSD nodes
+    installer = ceph_cluster.get_nodes(role="installer")[0]
+    osd_nodes = ceph_cluster.get_nodes(role="osd")
+
+    # Get device spec
+    device_spec = config.get("spec", {}).get("spec", {})
```
pranavprakash20, line 74: @adityaramteke This will fail as there is no section named `spec` in the suite. This has to be changed to `specs` as per the suite file.

pranavprakash20:
Please attach a run log @adityaramteke

adityaramteke:
@pranavprakash20 logs are attached please do a review.

MohitBis (dismissed):
LGTM

tests/ceph_volume/test_ceph_volume_wrong_db_size_compute.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # Create a spec file after updating devices
+    file = create_yaml_config(installer, specs)
+
+    # Create OSDs with spec file
+    c = {"pos_args": [], "input": file}
+    CephAdm(nodes=installer, mount=file).ceph.orch.apply(**c)
+
+    # Sleep for 60s for OSDs to be generate
+    time.sleep(60)
```
vamahaja, line 101: why constant sleep ? can we use waiter for OSD check ?

adityaramteke: I'll explain, so when we apply spec, generating OSD ids takes few seconds, if we run without sleep, it gives index out of range. 

Why waiter won't work here is, I have wait_for_osd_daemon_state to check if OSDs are up and running there we are using waiter again. 

So using time.sleep will look a good approach. Let me know if you still think differently.

tintumathew10 (commented):
LGTM

vamahaja (approved):

PR #3381, opened 2024-02-07
[tfa-fix] Rename test_osd_memory_target for cephadm and fix json error

vinpapnoi:
# Description

### Problem:

Test suite `tier-2-osd-scenarios` failed for the test `test_osd_memory_target.py`.

### Reason for Failure:

There exist two files of the same name under different directories such as :

- `tests/rados/test_osd_memory_target.py` : https://github.com/red-hat-storage/cephci/blob/0c60bf9638cee697fe179cfe7b9f322cac5ff80d/tests/rados/test_osd_memory_target.py
- `tests/cephadm/test_osd_memory_target.py` : https://github.com/red-hat-storage/cephci/blob/0c60bf9638cee697fe179cfe7b9f322cac5ff80d/tests/cephadm/test_osd_memory_target.py

Both test files differ in contents. 

The suite `tier-2-osd-scenarios` must use the file under `cephadm` directory but it is instead using the file under the `rados` directory.

Also, when using the correct test file, the test fails with error :  `JSONDecodeError: Expecting property name enclosed in double quotes`

### Solution:

Rename `tests/cephadm/test_osd_memory_target.py` to `tests/cephadm/test_osd_memory_target_cephadm.py` and update all suite files which use this test.
Also, fix json decode error.

### DIFF
--- a/tests/cephadm/test_osd_memory_target_cephadm.py
+++ b/tests/cephadm/test_osd_memory_target_cephadm.py
@@ -38,7 +38,7 @@ def run(ceph_cluster, **kw):
     log.info(f"Memory target is set on host '{hostname}' : '{value}'")
 
     # Verify that the option is set in ceph config dump
-    kw = {"format": "json-pretty"}
+    kw = {"format": "json"}
     out = cephadm.ceph.config.dump(**kw)
     data = json.loads(out[0])
     found = False
@@ -56,7 +56,7 @@ def run(ceph_cluster, **kw):
 
     # Verify osd memory target for osd on the host matches the host level value
     out = cephadm.ceph.osd.tree(**kw)
-    data = json.loads(out[0])
+    data = json.loads(out)
     for item in data["nodes"]:
         if item.get("type") == "host" and item.get("name") == hostname:
             osd_list = item["children"]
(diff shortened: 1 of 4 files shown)


vinpapnoi:
Test run logs : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-PY26SZ/Test_OSD_Memory_Target_0.log

tintumathew10 (commented):
LGTM

tests/cephadm/test_osd_memory_target_cephadm.py
```diff
@@ -38,7 +38,7 @@ def run(ceph_cluster, **kw):
     log.info(f"Memory target is set on host '{hostname}' : '{value}'")
```
vamahaja, line 38: I see these tests are used in RADOS test suites [1] as well, if renamed it will broke these test suites, Update test suites accordingly.
[1] https://github.com/red-hat-storage/cephci/blob/master/suites/pacific/rados/tier-2_rados_test-osd-rebalance.yaml#L94

vinpapnoi: We can let the RADOS test suite files as is because they are under different directory and once the rename for the cephadm test is done then the naming will be different for both and they will execute the appropriate tests.

vamahaja: ACK

vamahaja (approved):

PR #3445, opened 2024-02-28
Fix for upgrade 7-1/2 to 7

tintumathew10:
# Description

Fix for upgrade 7-1/2 to 7
image

### DIFF
--- a/ceph/ceph_admin/bootstrap.py
+++ b/ceph/ceph_admin/bootstrap.py
@@ -363,8 +363,11 @@ class BootstrapMixin:
                 "promtail",
                 "snmp_gateway",
                 "loki",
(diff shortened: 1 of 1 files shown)


ceph/ceph_admin/bootstrap.py
```diff
@@ -365,6 +365,10 @@ def bootstrap(self: CephAdmProtocol, config: Dict):
                 "loki",
                 "nvmeof",
```
vamahaja, line 366: Don't need to remove this ?

tintumathew10: good catch.

vamahaja (approved):

PR #3450, opened 2024-02-29
mds backtrace damage repoduce

julpark-rh:
Adding MDS backtrace damage reproduction (Needs to be changed once scrub fix is merged)

http://magna006.ceph.redhat.com/cephci-jenkins/cephci-run-8MXFZ7/

    Test steps:
    1. mount a directory in specific client using fuse in a pool
    2. mount a directory in specific client using kernel in the same pool
    3. create directories in the mounted directory for both clients
    4. create files in the mounted directory for both clients
    5. list out all the objects in the pool
    6. find all the objects in the pool ending with "\.00000000"
    7. If the objects have attribute "parent" then it is a directory
    8. delete the "parent" attribute from the objects
    9. run scrub with no option
    10. check if the cluster is damaged state after the scrub.

# Description

<details>

</details>

### DIFF
--- a/tests/cephfs/mds_backtrace_damage.py
+++ b/tests/cephfs/mds_backtrace_damage.py
@@ -1,0 +1,154 @@
+import random
+import string
+import time
+import traceback
+
+from tests.cephfs.cephfs_basic_tests import mount_test_case
+from tests.cephfs.cephfs_utilsV1 import FsUtils
+from utility.log import Log
+
+log = Log(__name__)
+
+
+def run(ceph_cluster, **kw):
+    """
+    Test steps:
+    1. mount a directory in specific client using fuse in a pool
+    2. mount a directory in specific client using kernel in the same pool
+    3. create directories in the mounted directory for both clients
+    4. create files in the mounted directory for both clients
+    5. list out all the objects in the pool
+    6. find all the objects in the pool ending with ".00000000"
+    7. If the objects have attribute "parent" then it is a directory
+    8. delete the "parent" attribute from the objects
+    9. run scrub with no option
+    10. check if the cluster is damaged state after the scrub.
+    """
+    try:
+        fs_util = FsUtils(ceph_cluster)
+        clients = ceph_cluster.get_ceph_objects("client")
+        client1 = clients[0]
+        num_dir = 20
+        num_files = 10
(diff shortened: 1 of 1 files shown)


tests/cephfs/mds_backtrace_damage.py
```diff
@@ -0,0 +1,154 @@
...
+        # run scrub
+        client1.exec_command(sudo=True, cmd=f"ceph tell mds.{fs_name}:0 scrub start /")
+        # check if the cluster is in damaged state
+        time.sleep(30)
+        out1, _ = client1.exec_command(sudo=True, cmd="ceph -s")
+        print(out1)
+        out2, _ = client1.exec_command(sudo=True, cmd="ceph fs status")
+        print(out2)
+        if "HEALTH_ERR" in out1:
+            log.error("Cluster is in damaged state")
```
sumabai, line 123: Cluster being in damaged state with given workflow is Expected right? Then why should this be reported as ERROR, can be it a INFO, and string can be appended with 'as Expected' ?

julpark-rh: oh you are right. let me fix that.

sumabai (approved):
lgtm

tests/cephfs/mds_backtrace_damage.py
```diff
@@ -0,0 +1,154 @@
...
+        time.sleep(30)
+        out1, _ = client1.exec_command(sudo=True, cmd="ceph -s")
+        print(out1)
+        out2, _ = client1.exec_command(sudo=True, cmd="ceph fs status")
+        print(out2)
+        if "HEALTH_ERR" in out1:
+            log.info("Cluster is in damaged state")
+            return 0
+        else:
+            log.info("Cluster is not in damaged state")
```
neha-gangadhar, line 126: The else part should be reported as log.error if the cluster does not reach damaged state?

neha-gangadhar:
looks good

tests/cephfs/mds_backtrace_damage.py
julpark-rh: I have changed it to `log.error()`
thanks

neha-gangadhar (approved):
lgtm

PR #3467, opened 2024-03-05
Clone scale operations for a snapshot of an image

pavankumarag:
This test the clone operations at scale for a given snapshot. All details of testcase is covered in https://polarion.engineering.redhat.com/polarion/#/project/CEPH/workitem?id=CEPH-83584436.

**Steps**:

1. Create snap Eg rbd snap create   --pool rbd_scale_snap_pool --image rbd_scale_snap_image --snap snap_0
2. Perform snap clone parallely Eg rbd clone rbd_scale_snap_pool/rbd_scale_snap_image@snap_0 rbd_scale_snap_pool/clone_SfiDL
3. 3. Remove snap Eg rbd snap rm --pool rbd_scale_snap_pool --image rbd_scale_snap_image --snap snap_0

**Run Summary:**
image

**Magna logs location:**

http://magna002.ceph.redhat.com:8083/clone_scale/

### DIFF
--- a/ceph/rbd/workflows/snap_clone_operations.py
+++ b/ceph/rbd/workflows/snap_clone_operations.py
@@ -278,6 +278,21 @@ def clone_ops(**kw):
                 return 1
             clones = json.loads(out)
             for clone in clones:
+                # due to parallel support, have to check if clone already flattened or not
+                du_out, du_err = rbd.du(pool=clone["pool"], format="json")
+                if du_err:
+                    log.error(f"rbd du failed for {pool} with error {du_err}")
+                    if test_ops_parallely:
+                        raise Exception(f"rbd du failed for {pool} with error {du_err}")
+                    return 1
+                du_out_json = json.loads(du_out)
+                for _image in du_out_json["images"]:
+                    if _image["name"] == clone["image"]:
+                        if _image["used_size"] != 0:
+                            log.info(
+                                f"{clone['image']} is already flattened so skipping"
+                            )
+                            break
                 _, err = rbd.flatten(pool=clone["pool"], image=clone["image"])
                 if "100% complete...done" not in out + err:
                     log.error(
@@ -435,11 +450,14 @@ def remove_snap_and_verify(**kw):
     image = kw.get("image")
     test_ops_parallely = kw.get("test_ops_parallely", False)
     snap_name = kw.get("snap_name", None)
+    protected = kw.get("protected", "true")
     user_defined_snap = dict()
     if snap_name is None:
         user_defined_snap = get_user_defined_snaps(rbd=rbd, pool=pool, image=image)[0]
     else:
         user_defined_snap.update({"name": snap_name})
+        if protected == "true":
+            user_defined_snap.update({"protected": protected})
     if not is_secondary and user_defined_snap.get("protected", "false") == "true":
         _, err = rbd.snap.unprotect(
             pool=pool, image=image, snap=user_defined_snap["name"]
@@ -565,11 +583,11 @@ def wrapper_for_image_snap_ops(**kw):
             "test_ops_parallely": <>
         }
     """
-    rbd = kw.get("rbd")
-    pool = kw.get("pool")
+    rbd = kw.pop("rbd")
+    pool = kw.pop("pool")
     image = kw.pop("image", {})
     snap_names = kw.pop("snap_names", [])
-    test_ops_parallely = kw.get("test_ops_parallely", False)
+    test_ops_parallely = kw.pop("test_ops_parallely", False)
     ops_module = kw.pop("ops_module", {})
     mod_obj = import_module(ops_module)
     ops_method = kw.pop("ops_method", {})
@@ -586,10 +604,11 @@ def wrapper_for_image_snap_ops(**kw):
                     image=image,
                     snap_name=snap,
                     test_ops_parallely=True,
+                    **kw,
                 )
     else:
         for snap in snap_names:
-            rc = method_obj(rbd=rbd, pool=pool, image=image, snap_name=snap)
+            rc = method_obj(rbd=rbd, pool=pool, image=image, snap_name=snap, **kw)
             if rc:
                 log.error(
                     f"Test image snap operation {ops_method} failed for {pool}/{image}/{snap}"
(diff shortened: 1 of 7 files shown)


Manohar-Murthy:
Include IO during snap and clone operations.

pavankumarag:
New logs here: http://magna002.ceph.redhat.com:8083/clone_scale/

ceph/rbd/workflows/snap_clone_operations.py
```diff
@@ -278,6 +278,21 @@ def clone_ops(**kw):
...
+                du_out, du_err = rbd.du(pool=clone["pool"], format="json")
+                if du_err:
+                    log.error(f"rbd du failed for {pool} with error {du_err}")
+                    if test_ops_parallely:
+                        raise Exception(f"rbd du failed for {pool} with error {du_err}")
+                    return 1
+                du_out_json = json.loads(du_out)
+                for _image in du_out_json["images"]:
+                    if _image["name"] == clone["image"]:
+                        if _image["used_size"] != 0:
```
sunilangadi2, line 291: does used bytes depends on clone is flattened or not?

ceph/rbd/workflows/snap_clone_operations.py
```diff
@@ -565,11 +583,11 @@ def wrapper_for_image_snap_ops(**kw):
             "test_ops_parallely": <>
         }
     """
-    rbd = kw.get("rbd")
-    pool = kw.get("pool")
+    rbd = kw.pop("rbd")
```
sunilangadi2, line 586: why `pop` here?

suites/quincy/rbd/tier-3_rbd_snap_clone_operations.yaml
```diff
@@ -0,0 +1,104 @@
...
+      desc: Install rbd-nbd and remove any epel packages
+      module: exec.py
+      name: Install rbd-nbd
+      config:
+        sudo: true
+        commands:
+          - "rm -rf /etc/yum.repos.d/epel*"
+          - "dnf install rbd-nbd -y"
+
+  - test:
```
sunilangadi2, line 84: Does it need separate tets suite?
Does it create one pool, one image on one snap that many clones?

ceph/rbd/workflows/snap_clone_operations.py
pavankumarag: yup, it will fail as image is already flattened since we run in parallel. The problem is "snap protect + clone + flatten + snap unprotect" is too tightly coupled so in parallel flatten gets picked up for same clone image multiple times, so added this safeguard instead of decoupling "snap protect + clone + flatten + snap unprotect"

ceph/rbd/workflows/snap_clone_operations.py
pavankumarag: pop to avoid duplicate kw arg error, same like wrapper_for_image_ops

suites/quincy/rbd/tier-3_rbd_snap_clone_operations.yaml
pavankumarag: yes, previous testsuite was about 

per pool, per image -> snap at scale eg 500

This test suite is about

per pool, per image, per snap -> clone at scale eg 500.

Since "snap protect + clone + flatten + snap unprotect" is too tightly coupled, had to do few workarounds to fit in.

pavankumarag:
> > This test the clone operations at scale for a given snapshot. All details of testcase is covered in https://polarion.engineering.redhat.com/polarion/#/project/CEPH/workitem?id=CEPH-83584436.
> > **Steps**:
> > 
> > 1. Create snap Eg rbd snap create   --pool rbd_scale_snap_pool --image rbd_scale_snap_image --snap snap_0
> > 2. Perform snap clone parallely Eg rbd clone rbd_scale_snap_pool/rbd_scale_snap_image@snap_0 rbd_scale_snap_pool/clone_SfiDL
> > 3. 3. Remove snap Eg rbd snap rm --pool rbd_scale_snap_pool --image rbd_scale_snap_image --snap snap_0
> > 
> > **Run Summary:** image
> > **Magna logs location:**
> > http://magna002.ceph.redhat.com:8083/clone_scale/
> 
> @pavankumarag Please add detailed steps in polarion, as to after creating a snap, we need to protect the snap, only then we can create a clone. Also, if you have included clone flatten option as well here, please add that in the test steps in polarion.
> 
> If you haven't included clone flatten, please include the same.
> 
> Also since this is a tier-3 scenario, the test scope could also include running IOs on the newly created clones. And further resizing, renaming, snapping and cloning the clone.
> 
> Also after running IOs on the image, we would need to take md5sum of the file written and once clone is created we need to verify that the file present in clone has the same md5sum as the parent image. Without verifying data integrity there is no way for us to make sure that the clone is the exact copy of the image at the instant snapshot was taken. So the test scenario wouldn't be complete without this verification.
> 
> Sorry to be mentioning all these larger changes in the last minute, but please include these before marking the polarion test case as automated. Since otherwise it would just be an incomplete functional regression of creating clone, not a comprehensive tier-3 scenario.
> 
> Please do think and see if there are some more things to be considered with snaps and clones. @sunilangadi2 Please check if there's anything more to add as well.

Thanks @manasagowri for the detailed feedback. Couple of my thoughts and data points, collating here as they are scattered.

Me and @manasagowri had a discussion, here are the outcomes

> Using RBD children or RBD info before clone flatten instead of rbd du - both have pass percentage issue due to design challenges  like snap-protect, clone, clone-flatten, snap-unprotect are tightly coupled.

> Md5sum - currently there are design challenges as snap-protect, clone, clone-flatten, snap-unprotect are tightly coupled, can evaluate for next sprint

> Using wrapper_for_image_ops for snap as well - Here I have isolated snap operations to wrapper_for_snap_operations due the change in paramaters format and its working well, IMHO its better to let this be in the current way 

> Adding clone IO, clone resize, clone rename etc in the same testcase - The way I have thought about it is the way customer performs bulk operations i.e
      - customer will not do bulk clone resize and clone renames
      - However customer will do bulk clones or bulk snaps of a image
Hence I have loosely couple snap and clone cases in to two cases
(1) Pool -> Image -> snap creation at scale - PR already merged, need to add IO in next OR
(2) Pool -> Image -> snap -> clone creation at scale - this PR

If we collectively think otherwise, I can extend this same testcase to clone IO + clone resize + clone rename at scale. But the scope and efforts are huge.
 `We decided to add these also as a onetime effort, additional check on clone operations`

So this PR will closed and reworked in the subsequent PR to address the design challenges.

PR #3470, opened 2024-03-05
[tfa-fix] Remove dashboard parameter from the bootstrap modules

vinpapnoi:
# Description

### Problem:
Test suite `tier2-permissive-mode-scenarios` failed to perform bootstrap with error `cephadm: error: unrecognized arguments: --dashboard-user`

### Reason for Failure:
When providing the "dashboard" parameter to the cephadm-ansible modules for bootstrap, we encounter the above error and a bug has been raised for the same : https://bugzilla.redhat.com/show_bug.cgi?id=2267865

### Solution:
The dashboard is enabled by default in deployments so the parameter can be removed from the bootstrap modules.
The test suites which use the modules do not involve any dashboard specific testing so removing the parameter will not affect the tests.

### DIFF
--- a/tests/cephadm/ansible_wrapper/cephadm_bootstrap/bootstrap-with-custom-ssh.yaml
+++ b/tests/cephadm/ansible_wrapper/cephadm_bootstrap/bootstrap-with-custom-ssh.yaml
@@ -8,7 +8,6 @@
       cephadm_bootstrap:
         mon_ip: "{{ mon_ip }}"
         image: "{{ image | default(False) }}"
-        dashboard: "{{ dashboard | default(True) }}"
         monitoring: "{{ monitoring | default(False) }}"
         firewalld: "{{ firewalld | default(False) }}"
         ssh_user: "{{ ssh_user }}"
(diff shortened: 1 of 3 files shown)


tintumathew10 (commented):
LGTM

pranavprakash20 (approved):

tests/cephadm/ansible_wrapper/cephadm_bootstrap/bootstrap-with-custom-ssh.yaml
```diff
@@ -8,7 +8,6 @@
       cephadm_bootstrap:
         mon_ip: "{{ mon_ip }}"
         image: "{{ image | default(False) }}"
-        dashboard: "{{ dashboard | default(True) }}"
```
vamahaja, line 11: Why this removed ? Is this parameter unsuppported ?

If test case require to disable this then set default value to `False`

vinpapnoi: This parameter is causing failure in the tests since it requires "dashboard-user". Bug has been raised for the same (see description)
It is not necessary to provide this parameter since it enabled by default. To disable it we can use "skip-dashboard" parameter instead.
Hence for the CI we can remove this parameter completely.

vamahaja (approved):

PR #3498, opened 2024-03-11
[Lib] Update steps to get public IP for baremetal

vamahaja:
Update `cli/cloudproviders/baremetal.py` library method `get_node_public_ips` to get public IPs using `socket` connection.


tintumathew10 (commented):
LGTM

PR #3499, opened 2024-03-11
[LibFix] Update cluster-conf library to create config

vamahaja:
Update `cephci/cluster-conf.py` to generate cluster conf based on new format


tintumathew10 (approved):
lgtm

PR #3501, opened 2024-03-12
Rados TFA fixes

pdhiran:
1. Updated `create_inconsistent_object` method to check if the KW pair deleted is listed post deletion in the Object. ( lines to verify successful deletion )
2. Added 2 new methods `start_check_scrub_complete` & `start_check_deep_scrub_complete`, which triggers Scrubbing on a PG, and returns true once the scrubbing/deep-scrubbing is complete on the PG.
3. Updated the `target_size_ratio` to 0.8, to make a drastic change in the count of the PGs on pool.
4. Updated the pool tests to use bulk flag functionality, just checking PG count change for functionality verification.
5. Updated the PG autoscale mode warn tests.
6. In method `wait_for_device`, the module was failing as it was picking a non-osd container to get the disk details upon OSD addition/deletion. Updated the code to use OSD container.
7. Updated methods to check for scrub completion before checking the logs for scrub messages.

### DIFF
--- a/ceph/rados/core_workflows.py
+++ b/ceph/rados/core_workflows.py
@@ -821,6 +821,7 @@ class RadosOrchestrator:
                 2. target_size_ratio: ratio of cluster pool will utilize. Values -> 0 - 1. (float)
                 3. target_size_bytes: size the pool is assumed to utilize. eg: 10T (str)
                 4. pg_num_min: minimum pg's for a pool. (int)
+                5. pool_name: Name of the pool on which operation is to be done
         Returns:
         """
         pool_name = kwargs["pool_name"]
@@ -1504,11 +1505,11 @@ class RadosOrchestrator:
         cmd = f"ceph pg {pg_id} query"
         try:
             pg_query = self.run_ceph_command(cmd=cmd)
+            log.debug(f"The status of pg : {pg_id} is {pg_query['state']}")
+            return pg_query["state"]
         except Exception as err:
-            log.error(f"Hit exception : {err}")
+            log.error(f"Hit exception collecting PG state for : {pg_id}. Error:  {err}")
             return False
-        log.debug(f"The status of pg : {pg_id} is {pg_query['state']}")
-        return pg_query["state"]
 
     def get_osd_map(self, pool: str, obj: str, nspace: str = None) -> dict:
         """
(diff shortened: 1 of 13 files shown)


ceph/rados/core_workflows.py
```diff
@@ -2234,21 +2235,161 @@ def create_inconsistent_object(self, pool_name, object_name):
...
             if rm_key_count == 3 or not obj_del_check:
                 break
         if not self.change_osd_state(action="start", target=primary_osd):
             log.error(f"Unable to start the OSD : {primary_osd}")
             raise Exception("Execution error")
         log.info(f"Performing the deep-scrub on the pg-{pg_id}")
-        self.run_deep_scrub(pgid=pg_id)
-        # sleeping for 10 seconds, waiting for scrubbing to be over
+        if not self.start_check_scrub_complete(pg_id=pg_id):
+            log.debug(f"deep-scrubbing could not be completed on PG : {pg_id}")
```
neha-gangadhar, line 2254: The function invoked "start_check_scrub_complete" -> is performing scrub so can we modify the debug logger to "Scrubbing could not be completed on pg" instead of "deep-scrubbing"?

ceph/rados/core_workflows.py
```diff
@@ -2234,21 +2235,161 @@ def create_inconsistent_object(self, pool_name, object_name):
...
+            raise Exception("PG not deep-scrubbed error")
+
+        # sleeping for 10 seconds
         time.sleep(10)
         health_check = self.check_inconsistent_health(True)
         assert health_check
         log.info(f"The inconsistent object is created in the pg{pg_id}")
         return pg_id
 
+    def start_check_scrub_complete(self, pg_id):
```
neha-gangadhar, line 2264: we could have single module with parameter passed as "deep-scrub" if we want to check scrub or deep-scrub. However not mandatory, keeping separate is fine as well.

tests/rados/test_scrub_log.py
```diff
@@ -126,18 +127,24 @@ def check_scrub_workflow(installer, rados_obj, acting_sets, task) -> bool:
...
-            rados_obj.run_scrub(pgid=pgid)
+            log.debug(f"Running scrub on pg : {pgid}")
+            if not rados_obj.start_check_scrub_complete(pg_id=pgid):
+                log.error(f"Could not complete scrub on pg : {pgid}")
+                return False
             log_lines = [f"{pgid} scrub starts", f"{pgid} scrub ok"]
         elif task == "deep-scrub":
-            rados_obj.run_deep_scrub(pgid=pgid)
+            if not rados_obj.start_check_deep_scrub_complete(pg_id=pgid):
+                log.error(f"Could not complete scrub on pg : {pgid}")
```
neha-gangadhar, line 140: Please modify logger to "Could not complete deep scrub on pg"

neha-gangadhar (commented):
Overall looks good. Can you please run the failed tests and share run logs.

pdhiran:
Pass log : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-20AGGV

ceph/rados/core_workflows.py
pdhiran: Thanks for the catch here. the Method called should be deep-scrub itself. The inconsistent Obj is only identified upon a deep-scrub on the PG. Updated the method call.

ceph/rados/core_workflows.py
pdhiran: Since the params checked are different, I thought it would be good to have two methods instead of if-else statements. Having two methods would also satisfy SOLID principles. 

I am in favour of 2 separate methods itself. If needed, can be merged into 1.

tests/rados/test_scrub_log.py
pdhiran: Updated the logging.

tests/rados/pool_tests.py
```diff
@@ -255,46 +255,55 @@ def run(ceph_cluster, **kw):
...
+        log.debug("Set the bulk flag on the pool to False")
+        time.sleep(120)
 
-        if config.get("delete_pool"):
-            rados_obj.detete_pool(pool=pool_name)
+        pg_count_bulk_false = rados_obj.get_pool_details(pool=pool_name)[
+            "pg_num_target"
+        ]
+        log.debug(
+            f"PG count on pool {pool_name} post addition of bulk flag : {pg_count_bulk_true}"
```
pdhiran, line 299: Update comment

SrinivasaBharath (approved):
LGTM

harshkumarRH (approved):
Looks good!

PR #3557, opened 2024-03-27
Changes in vault install script to accomodate failures for baremetal

TejasC88:
- Take the OS version if needed from the config.
- Defaults to rhel 9, only needs to be given for rhel 8

Pass log : http://magna002.ceph.redhat.com/ceph-qe-logs/tejas/cephci-run-2S38QX/


viduship (approved):
looks good

mkasturi18 (approved):

PR #3601, opened 2024-04-15
test_pg_split.py: Fixed the finally method as part of TFA

pdhiran:
In the finally block, we were hitting exceptions. Fixed the issue.

Checking if the variable was set in the workflow, before trying to access it.

File "/home/jenkins/ceph-builds/18.2.1-119/Regression/rados/97/cephci/run.py", line 820, in run
rc = test_mod.run(
File "/home/jenkins/ceph-builds/18.2.1-119/Regression/rados/97/cephci/tests/rados/test_pg_split.py", line 308, in run
if target_osd not in active_osd_list:
UnboundLocalError: local variable 'target_osd' referenced before assignment

Pass log : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-YWCSXC

### DIFF
--- a/tests/rados/test_pg_split.py
+++ b/tests/rados/test_pg_split.py
@@ -164,9 +164,6 @@ def run(ceph_cluster, **kw):
         log.debug(
             f" waiting for PGs to settle down on pool {pool['pool_name']} after applying bulk flag"
         )
-        method_should_succeed(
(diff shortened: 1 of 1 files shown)


neha-gangadhar (approved):
lgtm

PR #3611, opened 2024-04-16
RBD package support for pydocs

pavankumarag:
RBD package support for pydocs

python -m pydoc -p 1234

image


sunilangadi2 (approved):
if possible please cover RBD docs in **readthedocs** as well
https://cephci.readthedocs.io/en/latest/

PR #3617, opened 2024-04-18
Fix Rados Inconsistent Objects test cases

pdhiran:
1. Modified the test_osd_inconsistency.py file, where the method to check the health warnings was updated.
2. Test case : CEPH-83571453 marked inactive and Tests from CephCI removed. instead the Inconsistent Object test for EC pools will be run in test : CEPH-83586175

Pass logs : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-K01ZN8

### DIFF
--- a/ceph/rados/core_workflows.py
+++ b/ceph/rados/core_workflows.py
@@ -2206,11 +2206,14 @@ class RadosOrchestrator:
         start_time = time.time()
         while time.time() - start_time < time_check:
             health_detail = self.node.shell(args=["ceph health detail"])
-            log.info(f"Health warning: \n {health_detail}")
-            if "inconsistent" not in health_detail and inconsistent_present is False:
+            log.info(f"Health warning on cluster: \n {health_detail} \n\n")
+            status_report = self.run_ceph_command(cmd="ceph report", client_exec=True)
+            ceph_health_status = list(status_report["health"]["checks"].keys())
+            health_warn = "PG_DAMAGED"
+            if health_warn not in ceph_health_status and inconsistent_present is False:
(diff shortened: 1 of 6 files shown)


ceph/rados/core_workflows.py
```diff
@@ -2206,11 +2206,14 @@ def check_inconsistent_health(self, inconsistent_present: bool = True):
...
-            if "inconsistent" not in health_detail and inconsistent_present is False:
+            log.info(f"Health warning on cluster: \n {health_detail} \n\n")
+            status_report = self.run_ceph_command(cmd="ceph report", client_exec=True)
+            ceph_health_status = list(status_report["health"]["checks"].keys())
+            health_warn = "PG_DAMAGED"
+            if health_warn not in ceph_health_status and inconsistent_present is False:
                 log.info("pg inconsistent not exists in the cluster")
                 return True
-            elif "inconsistent" in health_detail and inconsistent_present is True:
+            elif health_warn in ceph_health_status and inconsistent_present is True:
```
harshkumarRH, line 2216: Can we keep all the the conditions? 
`elif "inconsistent" in health_detail and health_warn in ceph_health_status and inconsistent_present is True`

harshkumarRH (approved):
Looks good to me

ceph/rados/core_workflows.py
pdhiran: The last runs were failing for string check. But i believe that checking the health warning name should be sufficient.

neha-gangadhar (approved):
Approved

PR #3633, opened 2024-04-23
Rados mute alerts: Added adutomation for bug automation : 2247232

pdhiran:
Added -ve scenarios where we execute commands with wrong syntax and check for crashes

Pass log : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-J2AUAR/


neha-gangadhar (approved):
lgtm

PR #3723, opened 2024-05-27
PG Autoscaler tests: Added coverage for bug : 2136766

pdhiran:
Now checking if the PG autoscaler states are maintained post removal of autoscaler flags.

Pass log:
7.1 : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-JXX1HM


harshkumarRH (approved):
Looks good

neha-gangadhar (approved):
Approving based on 1 review.

PR #3738, opened 2024-05-30
[Test] Add ceph-volume starting and stopping OSD

adityaramteke:
Create ceph-volume based OSD.
Start the OSD using ceph-osd command.
Perform OSD in and out operation.
Validate OSDs are getting up and running.

Logs: http://magna002.ceph.redhat.com/cephci-jenkins/aramteke/automation/cephci-run-NGUN98/

### DIFF
--- a/cli/ceph/osd/osd.py
+++ b/cli/ceph/osd/osd.py
@@ -67,3 +67,21 @@ class Osd(Cli):
         if isinstance(out, tuple):
             return out[0].strip()
         return out
+
+    def out(self, id):
+        """Set osd out
+
+        Args:
+            id (str): osd id
+        """
+        cmd = f"{self.base_cmd} out {id}"
+        return self.execute(sudo=True, long_running=True, cmd=cmd)
+
+    def _in(self, id):
+        """Set osd in
+
+        Args:
+            id (str): osd id
+        """
+        cmd = f"{self.base_cmd} in {id}"
+        return self.execute(sudo=True, long_running=True, cmd=cmd)
--- a/cli/cephadm/cephadm.py
+++ b/cli/cephadm/cephadm.py
@@ -105,3 +105,19 @@ class CephAdm(Cli):
         if isinstance(out, tuple):
             return out[0].strip()
         return out
+
+    def ceph_osd(self, id, **kw):
+        """Run ceph-osd command
+
+        Args:
+            id (str): OSD id
+
+            Supported Keys:
+                conf (str): read configuration from the given confirguration file
+                keyring (str): keyring file name
+        """
+        cmd = f"{self.base_cmd} ceph-osd --id {id} {build_cmd_from_args(**kw)}"
+        out = self.execute(sudo=True, cmd=cmd)
+        if isinstance(out, tuple):
+            return out[0].strip()
+        return out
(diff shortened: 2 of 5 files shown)


cli/ceph/osd/osd.py
```diff
@@ -67,3 +67,21 @@ def tree(self, epoch=None, states=None, **kw):
...
+    def out(self, id):
+        """Set osd out
+
+        Args:
+            id (str): osd id
+        """
+        cmd = f"{self.base_cmd} out {id}"
+        return self.execute(sudo=True, long_running=True, cmd=cmd)
+
+    def _in(self, id):
```
pranavprakash20, line 80: Can we remove the _ here? Or any specific reason why its added Aditya?

cli/cephadm/cephadm.py
```diff
@@ -105,3 +105,17 @@ def logs(self, fsid, name):
...
+
+    def ceph_osd(self, id, conf, keyring):
+        """Run ceph-osd command
+
+        Args:
+            id (str): OSD id
+            conf (str): read configuration from the given confirguration file
+            keyring (str): keyring file name
+        """
+        cmd = f"{self.base_cmd} --id {id} --conf {conf} --keyring {keyring}"
```
pranavprakash20, line 117: The command to be executed is ceph-osd right? But i dont see that in the cmd here
The self.base_cmd is cephadm. 
Please let me know if i missed anything

pranavprakash20 (changes requested):

cli/ceph/osd/osd.py
adityaramteke: It wasn't allowing to use python keywords for method; hence I used _.

cli/cephadm/cephadm.py
adityaramteke: @pranavprakash20 you are right, I miss that adding, thanks for catching it. 
I have corrected this now, and updated the full logs.

cli/cephadm/cephadm.py
```diff
@@ -105,3 +105,17 @@ def logs(self, fsid, name):
         if isinstance(out, tuple):
             return out[0].strip()
         return out
+
+    def ceph_osd(self, id, conf, keyring):
```
pranavprakash20, line 109: one doubt @adityaramteke here, so are these the only params to ceph-osd command?
I mean there should be more isn't it? In that case, can we have these as **args? or similar logic?
Just a thought, what do you think?

cli/ceph/osd/osd.py
pranavprakash20: Got it. @vamahaja need a help here, any other workarounds?

cli/cephadm/cephadm.py
adityaramteke: @pranavprakash20 Yes, that should work, ID is mandatory parameter and rest are additional paramters, I have made the changes accordingly.

cli/ceph/osd/osd.py
adityaramteke: @pranavprakash20, @vamahaja only suggested me to have it this method with _.

vamahaja: @pranavprakash20 as `in` is inbuild python keyword, use `_in`

vamahaja (commented):
LGTM

pranavprakash20 (approved):
Lgtm

MohitBis (approved):
LGTM

PR #3739, opened 2024-06-02
[RADOS] Alternate approach to 'wait_for_device' method

harshkumarRH:
Jira tracker: [RHCEPHQE-13984](https://issues.redhat.com/browse/RHCEPHQE-13984): I would want to make robust: "wait_for_device()" method used in OSD replacements

Scope is to enhance the existing method `wait_for_device` to make it more robust and reduce occurrences of false failure

Test modules changed:
- `osd_remove` in `ceph/rados/utils.py`
- `wait_for_device_rados` in `tests/rados/rados_test_util.py`

Test cases modified:
-`tests/rados/test_bug_fixes.py`
-`tests/rados/test_osd_inprogress_rebalance.py`

Logs:
Quincy:
Reef:

### DIFF
--- a/tests/rados/test_osd_inprogress_rebalance.py
+++ b/tests/rados/test_osd_inprogress_rebalance.py
@@ -6,7 +6,7 @@ from ceph.rados.core_workflows import RadosOrchestrator
 from tests.rados.rados_test_util import (
     create_pools,
     get_device_path,
-    wait_for_device,
+    wait_for_device_rados,
     write_to_pools,
 )
 from tests.rados.stretch_cluster import wait_for_clean_pg_sets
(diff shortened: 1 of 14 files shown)


tests/rados/test_osd_inprogress_rebalance.py
```diff
@@ -6,7 +6,7 @@
 from tests.rados.rados_test_util import (
     create_pools,
     get_device_path,
-    wait_for_device,
+    wait_for_device_rados,
```
pdhiran, line 9: is the plan to do away with older implementation and use the new method? and replace each call one by one?

harshkumarRH: That's correct.
The old implementation could have been changed if the method was only being used in rados test cases, however, few rbd and  cephfs tests are also utilizing the same module and I did not want to affect them.
If the new implementation works flawlessly for a series of runs, we can make it the default.

pdhiran (approved):
looks good.

PR #3747, opened 2024-06-04
CephFS Customer Case - CLA

hkadam134:
# Description
Customer Case CLosed Loop : 
Bug : https://bugzilla.redhat.com/show_bug.cgi?id=2255435
Logs : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-K23KLJ/

I have also included  tests/cephfs/cephfs_utilsV1.py in the same PR which is needed for Scale Testing Scripts

<details>

</details>

### DIFF
--- a/tests/cephfs/clients/mds_conf_modifying.py
+++ b/tests/cephfs/clients/mds_conf_modifying.py
@@ -1,10 +1,13 @@
+import json
 import random
 import string
 import traceback
+from distutils.version import LooseVersion
 
 from ceph.ceph import CommandFailed
 from tests.cephfs.cephfs_utilsV1 import FsUtils
 from utility.log import Log
+from utility.utils import get_ceph_version_from_cluster
 
 log = Log(__name__)
 
@@ -22,6 +25,7 @@ def run(ceph_cluster, **kw):
         config = kw.get("config")
         clients = ceph_cluster.get_ceph_objects("client")
         build = config.get("build", config.get("rhbuild"))
+        ceph_version = get_ceph_version_from_cluster(clients[0])
         fs_util.prepare_clients(clients, build)
         fs_util.auth_list(clients)
         client1 = clients[0]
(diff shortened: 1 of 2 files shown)


tests/cephfs/clients/mds_conf_modifying.py
```diff
@@ -180,6 +181,11 @@ def run(ceph_cluster, **kw):
             )
             if output[0].rstrip() != target[1]:
                 raise CommandFailed(f"Failed to set {client_conf} to {value}")
+
+        rhbuild = config.get("rhbuild")
+        if "7" in rhbuild:
```
neha-gangadhar, line 186: The config value set to "0" is applicable to both 6.1 and 7.1 series, any reason to have the build check here?

neha-gangadhar:
It is a good practice to include suite files too in the same PR when raising for review to avoid to-and-fro on updates. If the review is complete, can be merged immediately instead of waiting for the suite file to be added. 
If the test is missed to add in the suite file, we may lose the execution cycle in CI.

tests/cephfs/clients/mds_conf_modifying.py
neha-gangadhar: Noted

neha-gangadhar:
> > It is a good practice to include suite files too in the same PR when raising for review to avoid to-and-fro on updates. If the review is complete, can be merged immediately instead of waiting for the suite file to be added. If the test is missed to add in the suite file, we may lose the execution cycle in CI.
> 
> The suite file is already present Neha, I am using the existing test file and just adding one more configuration.

thanks, noted.

tests/cephfs/clients/mds_conf_modifying.py
```diff
@@ -180,6 +181,11 @@ def run(ceph_cluster, **kw):
...
+            set_balance_automate_and_validate(client1)
```
AmarnatReddy, line 187: Hi Hemanth
Should we check the return value of the set_balance_automate_and_validate method and fail the test case if it fails any validation? Are we missing this step?

hkadam134: Validation is being done in this function only Amar - 
def set_balance_automate_and_validate(client):

Should we have one more validation check in the main function?

hkadam134:
The same test file (tests/cephfs/clients/mds_conf_modifying.py) is modified as part of one more PR - https://github.com/red-hat-storage/cephci/pull/3744.

I will tag as DNM in-order to avoid conflict.

tests/cephfs/clients/mds_conf_modifying.py
hkadam134: Updated Amar.

tests/cephfs/clients/mds_conf_modifying.py
```diff
@@ -56,6 +57,15 @@ def run(ceph_cluster, **kw):
                 )
                 return 1
 
+        if LooseVersion(ceph_version) >= LooseVersion("18.2"):
+            result = set_balance_automate_and_validate(client1)
+            if result:
```
AmarnatReddy, line 62: I think we don't need this if else block as we are directly raising the exception.
This will take program control to exception block where we are returning 1

hkadam134: Ok, Removed the check

sumabai (approved):
lgtm

julpark-rh:
LGTM

AmarnatReddy (approved):
LGTM

PR #3836, opened 2024-06-28
Adding test to failover and failback using power off and power on nodes

manasagowri:
Adding automation for NVMe HA failover and failback using power off and power on nodes for test case - https://polarion.engineering.redhat.com/polarion/#/project/CEPH/workitem?id=CEPH-83589012

Success Logs - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-U5NV9Y/

### DIFF
--- a/suites/reef/nvmeof/tier-2_nvmeof_4nodes_gateway_ha_tests.yaml
+++ b/suites/reef/nvmeof/tier-2_nvmeof_4nodes_gateway_ha_tests.yaml
@@ -1,5 +1,5 @@
 # Basic Ceph-NvmeoF sanity Test suite
-# cluster configuration file: conf/reef/nvmeof/ceph_nvmeof_sanity.yaml
+# cluster configuration file: conf/reef/nvmeof/ceph_nvmeof_ha_cluster_4nodes.yaml
 # inventory: conf/inventory/rhel-9.3-server-x86_64-xlarge.yaml
 
 tests:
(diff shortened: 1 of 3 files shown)


suites/reef/nvmeof/tier-2_nvmeof_4nodes_gateway_ha_tests.yaml
```diff
@@ -1,5 +1,5 @@
 # Basic Ceph-NvmeoF sanity Test suite
-# cluster configuration file: conf/reef/nvmeof/ceph_nvmeof_sanity.yaml
+# cluster configuration file: conf/reef/nvmeof/ceph_nvmeof_ha_cluster_4nodes.yaml
```
HaruChebrolu, line 2: Please add the same config changes into squid folder aswell.

HaruChebrolu:
I noticed you mentioned using OpenStack nodes for start and stop operations. Should we also consider scenarios involving baremetal nodes? If so, would get_openstack_driver() be useful in that context as well?

suites/reef/nvmeof/tier-2_nvmeof_4nodes_gateway_ha_tests.yaml
manasagowri: Done. added all changes from reef to squid test suite as well.

manasagowri:
> @manasagowri I noticed you mentioned using OpenStack nodes for start and stop operations. Should we also consider scenarios involving baremetal nodes? If so, would get_openstack_driver() be useful in that context as well?

@HaruChebrolu  This is just another way of failover and failback, so it is enough to regress it in openstack alone as of now. Since we don't have method to power off and power on baremetal nodes, we will not be adding the provision for that as part of this PR.

This test case is specific to functional test only, so just openstack is enough

sunilkumarn417 (approved):

PR #3887, opened 2024-07-16
Addition of a new generic test suite for RADOS with all the features

pdhiran:
Generic test suite for Rados.

log : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-X32YGQ/


suites/reef/rados/test_rados_all_generic_features.yaml
```diff
@@ -0,0 +1,1800 @@
...
+      module: test_objectstoretool_workflows.py
+      polarion-id: CEPH-83581811
+      desc: Verify ceph-objectstore-tool functionalities
+
+
+  - test:
+      name: ceph-bluestore-tool utility
+      module: test_bluestoretool_workflows.py
+      polarion-id: CEPH-83571692
+      desc: Verify ceph-bluestore-tool functionalities
```
harshkumarRH, line 1466: These two tests are already added around line 1350s

suites/reef/rados/test_rados_all_generic_features.yaml
harshkumarRH, line 1: Just my opinion, I am not in favour of adding so many tests in one test suite. I thought our intention was to create a test suite for advanced/extended regression but here execution list is so long that we are bound to have failures in rhos-d, and the cycle of debugging and analyzing will continue without actually yielding what we wanted in the first place.

tests/rados/pool_tests.py
```diff
@@ -1025,7 +1038,9 @@ def run(ceph_cluster, **kw):
                     )
                     raise Exception("Reads failed on the cluster Error")
 
-                if pool_obj.do_rados_put(client=client_node, pool=pool_name, nobj=10):
+                if pool_obj.do_rados_put(
+                    client=client_node, pool=pool_name, nobj=10, obj_name="test-obj-1"
```
harshkumarRH, line 1042: By providing obj name, we are just overwriting the same object `nobj` times, are we not fixing the `do_rados_put` method in this PR?

tests/rados/test_pg_autoscale_flag.py
```diff
@@ -47,16 +49,18 @@ def run(ceph_cluster, **kw):
...
-        pool_conf = pool_configs["erasure"]["sample-pool-1"]
-        create_given_pool(rados_obj, pool_conf)
-        rados_obj.set_pool_property(
-            pool=pool_conf["pool_name"], props="pg_autoscale_mode", value="off"
-        )
+        if create_re_pool:
+            pool_conf = pool_configs["replicated"]["sample-pool-1"]
+            create_given_pool(rados_obj, pool_conf)
+            rados_obj.set_pool_property(
+                pool=pool_conf["pool_name"], props="pg_autoscale_mode", value="warn"
```
harshkumarRH, line 56: Any reason why we are setting the autoscaler to warn here and off in case of EC pool?

pdhiran: This is the test for autoscale flag. So there can be 3 modes for autoscaler, ON, OFF, WARN. I wanted to have pools in all the states. By default, all pools are set to ON.

That's the reason, when i'm creating new pools, I am creating one with OFF, and another with WARN, so that we have pools on the cluster with all allowed values of autoscaler.

tests/rados/pool_tests.py
pdhiran: Fixing the method was not the intention of the PR, however, if needed, we can take that up.

neha-gangadhar (commented):
lgtm

suites/reef/rados/test_rados_all_generic_features.yaml
pdhiran: removed the duplicate tests

SrinivasaBharath (approved):

neha-gangadhar (approved):
Approved

PR #3921, opened 2024-07-25
[Rados] Cluster fill method for Rados Performance workflow

harshkumarRH:
Building blocks for Performance evaluation workflow at RADOS layer using radosbench tool

This PR adds the module to warm up the cluster to an initial desired percentage.

### DIFF
--- a/ceph/rados/core_workflows.py
+++ b/ceph/rados/core_workflows.py
@@ -3849,3 +3849,16 @@ class RadosOrchestrator:
         details = out["nodes"][0]
         log.debug(f" OSD : {osd_id} details are : {details}")
         return details
+
+    def get_osd_size(self, osd_id):
+        """
+        Method returns the size of the OSD in KB
+        Args:
+            osd_id: osd id
+        Returns: The total size of the osd in KB
+
+        """
+        total_size_osd = self.get_osd_df_stats(
+            tree=False, filter_by="name", filter=f"osd.{osd_id}"
+        )["summary"]["total_kb"]
+        return total_size_osd
--- a/ceph/rados/perf_workflow.py
+++ b/ceph/rados/perf_workflow.py
@@ -1,0 +1,160 @@
+"""
+Module to perform specific functionalities needed to run
+Performance Benchmarking with RADOS
+"""
+
+import time
+from math import floor
+
+from ceph.ceph_admin import CephAdmin
+from ceph.rados.core_workflows import RadosOrchestrator
+from utility.log import Log
+from utility.utils import method_should_succeed
+
+log = Log(__name__)
+
+
+class PerfWorkflows:
+    """
+    Contains various functions to run performance benchmarking at rados layer
+    """
+
+    def __init__(self, node: CephAdmin):
+        """
+        initializes the env to run Perf evaluation for rados
+        Args:
+            node: CephAdmin object
+        """
+        self.rados_obj = RadosOrchestrator(node=node)
+        self.cluster = node.cluster
+        self.clients = node.cluster.get_nodes(role="client")
+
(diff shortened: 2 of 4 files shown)


ceph/rados/core_workflows.py
```diff
@@ -3849,3 +3849,16 @@ def get_osd_details(self, osd_id: int):
         details = out["nodes"][0]
         log.debug(f" OSD : {osd_id} details are : {details}")
         return details
+
+    def get_osd_size(self, osd_id):
+        """
+        Method returns the size of the OSD
```
pdhiran, line 3855: maybe just elaborate in the docstring that the size is returned in KB

ceph/rados/perf_workflow.py
```diff
@@ -0,0 +1,160 @@
...
+        Returns:
+            obj count to be written by the client
+        """
+        # total available space for the pool
+        pool_stat = self.rados_obj.get_cephdf_stats(pool_name=pool_name)
+        max_avail_kb = pool_stat["stats"]["max_avail"]
+        # max_avail_mb = floor(max_avail_kb / 1 << 10)
+
+        # amount of data to be written to the pool
+        total_write_data = floor(max_avail_kb * fill_percent / 100)
```
pdhiran, line 78: Are we taking into consideration the replica size? for calculating total amount of data to be written?

pdhiran: If for now, we are only targeting RE pools, we can add that in a doc string. It's good enough to start from RE pools

ceph/rados/perf_workflow.py
```diff
@@ -0,0 +1,160 @@
...
+        obj_size_kb_int = int(obj_size[:-1])
+        if "M" in obj_size:
+            obj_size_kb_int *= 1 << 10
+        obj_count = floor(total_write_client / obj_size_kb_int)
+
+        pool_multi = self.get_pool_multiplier(pool_name=pool_name)
+        obj_count = int(obj_count / pool_multi)
+        return obj_count
+
+    def get_pool_multiplier(self, pool_name) -> 3:
```
pdhiran, line 93: the default return type should be int?? `get_pool_multiplier(self, pool_name) -> int:`

ceph/rados/perf_workflow.py
```diff
@@ -0,0 +1,160 @@
...
+        Method to calculate the object count multiplier depending
+        on the pool type, size and min size
+        Args:
+            pool_name:
+        Returns:
+            multiplier value (int)
+        """
+        pool_detail = self.rados_obj.get_pool_details(pool=pool_name)
+        if pool_detail["erasure_code_profile"]:
+            return pool_detail["size"] / pool_detail["min_size"]
```
pdhiran, line 104: Maybe we should round the number up to max 2 decimal points.

ceph/rados/perf_workflow.py
harshkumarRH: The pool multiplier or maybe I should rename it to pool divider has been written to take care of replica size and facilitate object count calculation accordingly

ceph/rados/perf_workflow.py
harshkumarRH: Any reason that would be necessary? The more the decimal points, more the precision

ceph/rados/core_workflows.py
harshkumarRH: Added

ceph/rados/perf_workflow.py
harshkumarRH: Changed the return type to float

neha-gangadhar (commented):
lgtm

pdhiran (approved):
Looks good.

PR #3922, opened 2024-07-25
[Rados] Changing wait_for_device method across the board

harshkumarRH:
A new method to wait_for_device was introduced with an earlier PR but it was intended that the stability of new code be tested before implementing it across the board.

After monitoring the working of the new method wait_for_device_rados, it was decided that the old method can be replaced with the new one, this PR addresses the same

### DIFF
--- a/tests/rados/test_cephdf.py
+++ b/tests/rados/test_cephdf.py
@@ -12,7 +12,7 @@ from ceph.ceph_admin import CephAdmin
 from ceph.rados import utils
 from ceph.rados.core_workflows import RadosOrchestrator
 from ceph.rados.pool_workflows import PoolFunctions
-from tests.rados.rados_test_util import get_device_path, wait_for_device
+from tests.rados.rados_test_util import get_device_path, wait_for_device_rados
 from tests.rados.stretch_cluster import wait_for_clean_pg_sets
 from utility.log import Log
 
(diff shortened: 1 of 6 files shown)


tests/rados/test_osd_rebalance.py
```diff
@@ -154,7 +154,7 @@ def run(ceph_cluster, **kw):
     return 0
 
 
-def wait_for_device(host, osd_id, action: str) -> bool:
+def wait_for_device_rados(host, osd_id, action: str) -> bool:
```
pdhiran, line 157: Why is the old method name being changed here? 

The new method is already present here : https://github.com/harshkumarRH/cephci/blob/092c1afbf9607b7622c3761d606a23348c4a2ab1/tests/rados/rados_test_util.py#L161

pdhiran:
Looks good

tests/rados/test_osd_rebalance.py
harshkumarRH: Thanks for pointing out, this was a mistake, fixed it now

pdhiran (approved):
Looks good.

pdhiran:
Only 1 rados test remains with custom implementation of  `wait_for_device` in test `tests/rados/test_osd_rebalance.py`.

If there are failures in the test, the method implementation would be changed. For now, looks to be working as expected.

PR #3944, opened 2024-07-31
[tfa][dmfg] Fix AttributeError for string split method applied on list

vinpapnoi:
# Description

Problem:
The suite `suites/reef/cephadm/tier4-service-operations.yaml` was failing for the test "Verify the logrotate config created by cephadm take into account tcmu-runner"

Reason for Failure:
The test failed for the below error :
`    for lists in rotate_conf.split(" || "):`
`AttributeError: 'list' object has no attribute 'split'`

Here the rotate_conf is a list and we cannot perform split operation on it.

Solution:
Perform split operation on the list item instead of the entire list.

Logs:
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-GU2PCV


pranavprakash20 (approved):
Lgtm

PR #3947, opened 2024-08-01
[Rados] TFA fix for osd df stats & removal of superblock redundancy test

harshkumarRH:
This PR covers the following 3 changes:

- `test_osd_df.py`: A fix provided in test-osd-df test where an acting set was being fetched for PG 1 instead of the desired PG

- `test_bluestore_superblock.py`: Test has been commented until the feature has been merged in Squid

- `test_mon_mgr_weight.py`: Fix has been merged in Quincy (6.1z7), so this test should now run for Quincy

Logs:
 `test_mon_mgr_weight.py`: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-17ZD4U

### DIFF
--- a/tests/rados/test_osd_df.py
+++ b/tests/rados/test_osd_df.py
@@ -109,10 +109,10 @@ def run_test(ceph_cluster, **kw):
 
         osd_map = rados_obj.get_osd_map(pool=pool_name, obj=object_name)
         acting_pg_set = osd_map["acting"]
-        log.info(f"Acting set for {object_name} in {pool_name}: {acting_pg_set}")
         if not acting_pg_set:
             log.error("Failed to retrieve acting pg set")
             return 1
+        log.info(f"Acting set for {object_name} in {pool_name}: {acting_pg_set}")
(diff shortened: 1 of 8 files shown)


tests/rados/test_osd_df.py
```diff
@@ -109,10 +109,10 @@ def run_test(ceph_cluster, **kw):
```
pdhiran, line 109: We also discussed to increase the number of iterations for the object append, to increase the object size. Are we planning to continue without that?

pdhiran: I see we have increased the iterations.

pdhiran (approved):
looks good

tests/rados/test_osd_df.py
harshkumarRH: @pdhiran 
Write iterations has been increased from 4 to 10 in every test suite, this would warrant an increase of object size from 16 MB to 40 MB, I guess for this increase should be enough
Please let me know if you want to me to increase it further to 80 MB or 100 MB

PR #3993, opened 2024-08-19
CEPH-83594645: Ensure client connection over v1 port does not crash

harshkumarRH:
[CEPH-83594645](https://polarion.engineering.redhat.com/polarion/#/project/CEPH/workitem?id=CEPH-83594645): Tier-2 test to verify client connections over V1 interface and/or port do not result in crash

Jira tracker: [RHCEPHQE-14409](https://issues.redhat.com/browse/RHCEPHQE-14409)

Bugzilla tracker - [2237038](https://bugzilla.redhat.com/show_bug.cgi?id=2237038)

Steps:
    1. Deploy a ceph cluster
    2. Choose a node at random which is not already configured as a client
    3. Copy client keyring and ceph conf file having only v1 ports
    4. Install ceph-common package
    5. Run ceph command using new client and monitor any crash

Test modules added:
- `tests/rados/test_v1client.py`

Test suites modified:
- `suites/squid/rados/tier-2_rados_test-bugfixes.yaml`
- `suites/reef/rados/tier-2_rados_test-bugfixes.yaml`

Logs:
Pacific: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-PZNVTV/
Quincy: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-3MAUK3/
Reef: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-PJBUS2/
Squid: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-QBAK65/

Signed-off-by: Harsh Kumar [hakumar@redhat.com](mailto:hakumar@redhat.com)

### DIFF
--- a/tests/rados/test_v1client.py
+++ b/tests/rados/test_v1client.py
@@ -1,0 +1,125 @@
+"""
+Module to verify client connection over v1 port does not result in crashes
+"""
+
+import datetime
+import random
+import re
+import time
+
+from ceph.ceph_admin import CephAdmin
+from ceph.rados.core_workflows import RadosOrchestrator
+from tests.rados.monitor_configurations import MonConfigMethods
+from utility.log import Log
+
+log = Log(__name__)
+
+
+def run(ceph_cluster, **kw):
+    """
+    # CEPH-83594645
+    Covers:
+        - BZ-2237038
+    Test to verify client connection over v1 port does not result in crashes
+    Steps
+    1. Deploy a ceph cluster
+    2. Choose a node at random which is not already configured as a client
+    3. Copy client keyring and ceph conf file having only v1 ports
+    4. Install ceph-common package
+    5. Run ceph command using new client and monitor any crash
+    """
+    log.info(run.__doc__)
+    config = kw["config"]
+    cephadm = CephAdmin(cluster=ceph_cluster, **config)
+    rados_obj = RadosOrchestrator(node=cephadm)
+    mon_obj = MonConfigMethods(rados_obj=rados_obj)
+    log.info(
+        "Running test case to verify client connection over v1 port does not result in crashes"
+    )
+
+    try:
+        """
+        If you use a client that does not support reading options from the configuration database,
+        or if you still need to use ceph.conf to change your cluster configuration for other reasons,
+        run the following command:
+               ceph config set mgr mgr/cephadm/manage_etc_ceph_ceph_conf false
+        You must maintain and distribute the ceph.conf file across the storage cluster.
+        Ref: https://red.ht/46YnXif
+        """
+
+        # set ceph_conf mgr setting to false
+        assert mon_obj.set_config(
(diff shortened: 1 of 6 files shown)


tests/rados/test_v1client.py
```diff
@@ -0,0 +1,103 @@
...
+    try:
+        # if rhbuild.startswith("5"):
+        #     log.info("Test is not valid for Pacific, BZ yet to be fixed/raised")
+        #     return 0
+
+        # choose an osd host at random
+        osd_hosts = rados_obj.ceph_cluster.get_nodes(role="osd")
+        osd_host = random.choice(osd_hosts)
+
+        rados_obj.configure_host_as_client(host_node=osd_host)
```
pdhiran, line 47: I do not feel comfortable converting a OSD host into a client with only v1 addresses present.

OSDs and most other daemons refer ceph.conf file upon restart, for connecting to mon config db.

Though this would improve test coverage, having a separate client node would be best.

We can add a new client, and use that to perform tests here. Can be taken up as part of new task.

tests/rados/test_v1client.py
```diff
@@ -0,0 +1,103 @@
...
+        for cmd in cmds:
+            log.info(f"Running cmd: {cmd}")
+            log.info(osd_host.exec_command(cmd=cmd, sudo=True)[0])
+
+        # check for any crashes on the cluster for 120 secs
+        timeout_time = datetime.datetime.now() + datetime.timedelta(seconds=120)
+        while datetime.datetime.now() < timeout_time:
+            # list any reported crashes on the cluster, should not have any
+            crash_list = rados_obj.do_crash_ls()
+            if len(crash_list) != 0:
```
pdhiran, line 82: There might be existing crashes on the cluster, which are not related to tests performed here.

We need to take into account that existing crashes present on the cluster do not account for this test failure here.

pdhiran: Also, we can run the commands in the while loop. 

This way, before and after each command we can check for crashes.

pdhiran (changes requested):
Few minor changes needed.

tests/rados/test_v1client.py
harshkumarRH: An OSD node out of the box does not have ceph.conf or ceph.client.keyrings present in /etc/ceph/ so why would it matter if a ceph.conf is added with only v1 addresses, and if addition of a ceph.conf with v1 addresses causes problems with restart, we will have a bug.

tests/rados/test_v1client.py
harshkumarRH: We run `crash ls-new`, so ideally we should be getting redundant crashes already present in the cluster. And the prior test crash coverage will be covered in the upcoming enhancement to cluster health check module.

Included the client command execution into the while loop

pdhiran (approved):
looks good

PR #4037, opened 2024-08-28
[RADOS] Check for crashes after each test execution

harshkumarRH:
Jira tracker: [RHCEPHQE-14573](https://issues.redhat.com/browse/RHCEPHQE-14573)

This PR addresses a necessary enchacment in RADOS test execution workflows, it had been recently observed that if during or after a test execution a crash is observed, the crash was just being logged as a health warning but the test was still being marked as Pass.

Now by introducing an additional check, we ensure that any test where a crash is observed in the cluster after test execution is marked as Fail.

Test modules added:
- `check_crash_status` in `ceph/rados/core_workflows.py`


pdhiran (approved):
looks good. 

good work in adding coverage to all tests.

PR #4039, opened 2024-08-29
Disable cephadm before initiating cluster purge.

psathyan:
# Description

Disabling cephadm from deploying new daemons helps to prevent any race conditions happening during cluster purge.

This commit includes cephadm disablement along with deploying of cephadm package on all nodes before initiating the purge.

Generally, cephadm package is only installed on the installer node.

## Testing
- [x] Unit testing.


AmarnatReddy (approved):
LGTM

PR #4054, opened 2024-09-03
[Tier-1] Cluster deployment to stop the RGW io sending multisite sync data

anrao19:
# Description
 This suites perform io on a cluster where rgw deplyed using a adavance rgw spec file updation to stop the RGW io sending multisite sync data
 New feature of 8.0
test with conf file: conf/squid/rgw/rgw_multisite_spec.yaml

</details>

### DIFF
--- a/ceph/ceph_admin/helper.py
+++ b/ceph/ceph_admin/helper.py
@@ -544,20 +544,23 @@ class GenerateServiceSpec:
         if node_names:
             spec["placement"]["hosts"] = self.get_hostnames(node_names)
 
-        if spec["spec"].get("rgw_frontend_ssl_certificate") == "create-cert":
-            subject = {
-                "common_name": spec["placement"]["hosts"][0],
-                "ip_address": self.cluster.get_node_by_hostname(
-                    spec["placement"]["hosts"][0]
-                ).ip_address,
-            }
-            key, cert, ca = generate_self_signed_certificate(subject=subject)
-            pem = key + cert + ca
-            cert_value = "|\n" + pem
-            spec["spec"]["rgw_frontend_ssl_certificate"] = "\n    ".join(
-                cert_value.split("\n")
-            )
-            LOG.debug(pem)
+        if spec.get("spec", False):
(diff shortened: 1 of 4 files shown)


ceph/ceph_admin/helper.py
```diff
@@ -545,20 +545,23 @@ def generate_rgw_spec(self, spec):
...
-                ).ip_address,
-            }
-            key, cert, ca = generate_self_signed_certificate(subject=subject)
-            pem = key + cert + ca
-            cert_value = "|\n" + pem
-            spec["spec"]["rgw_frontend_ssl_certificate"] = "\n    ".join(
-                cert_value.split("\n")
-            )
-            LOG.debug(pem)
+        if spec.get("spec", False):
```
hmaheswa, line 548: @anrao19 
whats the problem with the existing code here
could you add it in the description as well

ceph/ceph_admin/helper.py
```diff
@@ -545,20 +545,23 @@ def generate_rgw_spec(self, spec):
...
+            if spec["spec"].get("rgw_frontend_ssl_certificate", False):
```
hmaheswa, line 549: @anrao19 
also the line previously here was
if spec["spec"].get("rgw_frontend_ssl_certificate") == "create-cert":

but you are modifying it to 
if spec["spec"].get("rgw_frontend_ssl_certificate", False):

is this intentionally done or by mistake?

ceph/ceph_admin/helper.py
anrao19: @hmaheswa , if we use "command: apply_spec" without key "spec" it was failling with key error spec

ceph/ceph_admin/helper.py
anrao19: as of now in script we have rgw_frontend_ssl_certificate assigned only with value create-cert, so this was intentional

hmaheswa (approved):

ckulal (commented):
Changes looks good

PR #4080, opened 2024-09-10
Enhance crash module to archive existing crashes & max avail added for OSD out

harshkumarRH:
With recent changes in PR #4037 we introduced a crash module intended to run at the execution of each test workflow logging and checking crashes present in the cluster.
The current PR is an enhancement to the same where we are archiving any existing crashes on the cluster after they are reported. This ensures only one test case fails with a unique crash on the cluster and a series of tests do not fail due to the same common crash which are not even related to or caused the running workflow.

ceph-ci pipeline runs for recent Squid build saw [numerous false failures](http://magna002.ceph.redhat.com/cephci-jenkins/results/openstack/RH/8.0/rhel-9/Weekly/19.1.1-19/rados/6/tier-4_rados_tests) which prompted this necessary change.


pdhiran (approved):
Looks good

PR #4090, opened 2024-09-12
Custom image override shouldnt be used for ocs deploy

psathyan:
# Description

We are dependant on the override as provided in the recipe file unless, it is required by the integrating application. Also, we use the images that published.

The integration application can override via the UMB message.

*Checklist*

- [x] Unit test completed.


tintumathew10 (approved):
lgtm

PR #4091, opened 2024-09-12
Remove default tier-0 build type

psathyan:
# Description

We no longer have tier-0 as a build type hence using the default value as latest. This would be an untested version and would be required only when we need to have testing against a new relase.

Also, correctly the details of the system on which the execution was carried out.

*Checklist*

- [x] Unit testing completed.


tintumathew10 (approved):
LGTM

PR #4116, opened 2024-09-20
Moving radosbench shell executions to long running

harshkumarRH:
Currently if a radosbench command is stuck due to not being able to write data to the cluster, it remains in that state indefinitely.

Expectation is that the default timeout of 600 secs should come into picture and automation should not get stuck with radosbench.

Recent failure from ceph-ci:
Testrun - https://ocs3xstage-jenkins-csb-rhgsocs3x.apps.ocp-c1.prod.psi.redhat.com/job/qe-squid-rados-regression-openstack/20
Testsuite - https://ocs3xstage-jenkins-csb-rhgsocs3x.apps.ocp-c1.prod.psi.redhat.com/job/qe-squid-rados-regression-openstack/20/execution/node/74/log

From Pragadeeswaran Sathyanarayanan:

> A new channel is opened everytime exec_command is executed and a default timeout of 600 seconds is set to make it a blocking call. This is done to ensure the Channel (paramiko) to raise a socket.timeout exception when it exceeds the allocated time.
> However, an exception is only raised when there is pending read/write operation.
> A workaround for now would be set the parameter long_running=true when calling exec_command. Here, we are not dependant on the Channel to timeout instead is handled internally (i.e. control is not yielded to paramiko).

Following the suggested workaround, all radosbench shell calls have been changed to use `long_running=True`

Logs - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-HKZRI2

### DIFF
--- a/ceph/rados/core_workflows.py
+++ b/ceph/rados/core_workflows.py
@@ -164,7 +164,7 @@ class RadosOrchestrator:
             if client_exec:
                 out, err = self.client.exec_command(cmd=cmd, sudo=True, timeout=timeout)
             else:
-                out, err = self.node.shell([cmd], timeout=timeout, print_output=False)
+                out, err = self.node.shell([cmd], timeout=timeout)
         except Exception as er:
             log.error(f"Exception hit while command execution. {er}")
             return None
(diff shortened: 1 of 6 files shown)


pdhiran (approved):
Looks good.

PR #4117, opened 2024-09-20
RHCEPHQE-15796: TFA fix to make osd_memory_target config update relative to cluster

harshkumarRH:
Currently the test tests/rados/test_osd_memory_target.py contains hard-coded values which are being as osd_memory_target at OSD, and OSD host level.

Now although these values are compatible with RHOS-D pipeline executions, it was found that they result in false failures when the cluster is a baremetal setup with higher memory or a VM in IBM cloud.

To tackle this problem, the `osd_memory_target` value set at different instances is now relative to the optimal `osd_memory_target` fetched from the cluster 

Test modules modified:
- tests/rados/test_osd_memory_target.py

Logs:
Pacific:
Quincy:
Reef:
Squid: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-IVP58V

### DIFF
--- a/tests/rados/test_bluestoretool_workflows.py
+++ b/tests/rados/test_bluestoretool_workflows.py
@@ -463,7 +463,7 @@ def run(ceph_cluster, **kw):
             log.info(out)
             assert "success" in out
 
-            if rhbuild.split(".")[0] >= "6":
+            if rhbuild.split(".")[0] != "8":
                 # Execute ceph-bluestore-tool qfsck --path <osd_path>
                 osd_id = random.choice(osd_list)
                 log.info(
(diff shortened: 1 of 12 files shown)


neha-gangadhar (commented):
lgtm

tests/rados/test_bluestoretool_workflows.py
```diff
@@ -672,21 +672,17 @@ def run(ceph_cluster, **kw):
...
-                    "DB",
-                    "SLOW",
-                    "MAXIMUMS",
-                    "TOTAL",
-                ]:
-                    assert pattern in out
-            else:
-                for pattern in ["device size", "wal_total", "db_total", "slow_total"]:
-                    assert pattern in out
+            for pattern in [
```
pdhiran, line 675: Make the failure graceful, with exception handling for lower releases without the fix.

harshkumarRH: More detailed failure message added

pdhiran (approved):
Looks good

PR #4152, opened 2024-10-07
Update setting public network at global level instead of mon

harshkumarRH:
Recently RADOS stretch mode deployments have been failing during RHCS 8.0 pipeline runs, later it was observed that public network was being set at `mon` level through automation while cluster was now expecting it to be at `global` level

This PR addresses the necessary change in order to set the public network at `global` level for Ceph 8.0 deployments.
Additionally, cluster network is supposed to be set at `global` level but was incorrectly being set at `mon` level, this has also been rectified.

RHSC 8.0 bug tracker where this change was introduced - [2314438](https://bugzilla.redhat.com/show_bug.cgi?id=2314438)

Backports:
- 6.x: [2314604](https://bugzilla.redhat.com/show_bug.cgi?id=2314604)
- 7.x: [2314606](https://bugzilla.redhat.com/show_bug.cgi?id=2314606)

### DIFF
--- a/ceph/ceph_admin/bootstrap.py
+++ b/ceph/ceph_admin/bootstrap.py
@@ -401,10 +401,26 @@ class BootstrapMixin:
             public_nws = ",".join(filter(lambda x: x, list(set(public_nws.split(",")))))
 
         if public_nws:
-            self.shell(args=["ceph", "config", "set", "mon public_network", public_nws])
+            # Refer:
+            # - https://bugzilla.redhat.com/show_bug.cgi?id=2314438
+            # - https://bugzilla.redhat.com/show_bug.cgi?id=2156919
+            # - https://access.redhat.com/solutions/7088483
+            # Todo: update public network config level to global for 6.x and 7.x when changes are in
+            # 6.x tracker: https://bugzilla.redhat.com/show_bug.cgi?id=2314604
+            # 7.x tracker: https://bugzilla.redhat.com/show_bug.cgi?id=2314606
+            config_level = "global" if rhbuild.split(".")[0] >= "8" else "mon"
+            self.shell(
+                args=[
+                    "ceph",
+                    "config",
+                    "set",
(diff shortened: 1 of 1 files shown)


ceph/ceph_admin/bootstrap.py
```diff
@@ -401,10 +401,26 @@ def bootstrap(self: CephAdmProtocol, config: Dict):
...
+                    "config",
+                    "set",
+                    f"{config_level} public_network",
+                    public_nws,
+                ]
+            )
         if cluster_nws:
             self.shell(
-                args=["ceph", "config", "set", "mon cluster_network", cluster_nws]
+                args=["ceph", "config", "set", "global cluster_network", cluster_nws]
```
psathyan, line 423: I believe we would need `config_level` variable here.

IMHO, we could

```
config_level = "global" if rhbuild.split(".")[0] >= "8" else "mon"
if public_nws:
  ...
```

pdhiran (commented):
looks good. 

This PR is blocker in 3AZ, stretch mode deployment scenarios.

ceph/ceph_admin/bootstrap.py
harshkumarRH: `cluster network` is always at `global level`, it should not be set at only `mon` level.
@pdhiran Please confirm the above.

pdhiran: Yes, cluster_network is always set at global level, and mon level would not work. ( configs from mon level aren't propagated at OSD level. Whereas Configs at global level will be propagated at all daemon levels )


IBM doc link : https://www.ibm.com/docs/en/storage-ceph/7?topic=configuration-configuring-private-network

psathyan: @pdhiran @harshkumarRH thanks for the confirmation.

psathyan (approved):
/lgtm

PR #4160, opened 2024-10-11
Switch rados bench long_running to timeouts | Fix CBT bluefs-stats validation for Pacific

harshkumarRH:
Part - 1
===========================================================================================
During TFA analysis of RADOS test test_slow_op_requests.py
Basically the timeout of 3600 secs for a long_running execution is not getting considered because the code is but in an else block and the if block checks for existence of variable timeout which is always defined by default in the parent module called shell in ceph/ceph_admin/shell.py which calls exec_command

Failure log - http://magna002.ceph.redhat.com/cephci-jenkins/results/openstack/RH/7.1/rhel-9/Weekly/18.2.1-251/rados/52/tier-2_rados_test-slow-op-requests/Limit_slow_request_details_to_cluster_log_0.log

As shell.py is setting default timeout to 600, it was advised that rados bench executions be moved to client node instead of installer node or switch from `long_running` to setting longer `timeout` explicitly.

This PR sets `timeout` explicitly in every rados bench call, making it relative to the `rados_bench_duration`

Part - 2
===============================================================================================
The output of ceph-bluestore-tool bluefs-stats for Pacific is still different than others, this PR re-introduces conditional check for Pacific executions.

Logs:
Squid: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-E55H2B

### DIFF
--- a/ceph/rados/core_workflows.py
+++ b/ceph/rados/core_workflows.py
@@ -420,6 +420,7 @@ class RadosOrchestrator:
                 - check_ec (bool) -> boolean to control exit code check
                 - background (bool) -> run rados bench as background process and continue with test execution
                 - nocleanup (bool) -> if false, the nocleanup flag will not be added and the objects would be deleted
+                - timeout (int) -> user defined timeout for rados bench process
         Returns: True -> pass, False -> fail
         """
         duration = kwargs.get("rados_write_duration", 200)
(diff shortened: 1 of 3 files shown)


ceph/rados/core_workflows.py
```diff
@@ -428,6 +428,7 @@ def bench_write(self, pool_name: str, **kwargs) -> bool:
         verify_stats = kwargs.get("verify_stats", True)
         check_ec = kwargs.get("check_ec", True)
         nocleanup = kwargs.get("nocleanup", True)
+        _timeout = kwargs.get("timeout", duration + 100)
```
pdhiran, line 431: add the arg in docstring

pdhiran (changes requested):
Code Changes look Good. Please attach a pass log, were we exercise both rados read and write methods.

pdhiran (approved):
Looks Good.

PR #4210, opened 2024-11-07
Fixing Stretch mode netsplit test & replaced arbiter -> tiebreaker

pdhiran:
1. Added host reboot after flushing the IP tables in netsplit tests.
2. Replaced "arbiter" with "tiebreaker" in all the Stretch mode tests and suites


neha-gangadhar (approved):
Approved

PR #4229, opened 2024-11-17
RHCEPHQE-16740 - Move to concurrent.futures

psathyan:
# Description

In this PR, parallel is refactored to use concurrent.futures instead of gvent. This allows us to use threads or processes.

Along with the introduction of timeouts at the parallel phase. It also allows to avoid the dependency on gvent.

- [x] Unit testing
- [ ] Pipeline testing

*Unit Testing*
RGW Tier - 1 execution results are [here](http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-MZ7EY5/)

### DIFF
--- a/ceph/parallel.py
+++ b/ceph/parallel.py
@@ -1,121 +1,122 @@
-import sys
+# -*- code: utf-8 -*-
+"""
+This module provides a context manager for running methods concurrently.
 
-import gevent.pool
-import gevent.queue
+For backward compatability, spawn method is leveraged however one can also
+choose to move submit. Likewise, thread pool executor is the default executor.
 
-from utility.log import Log
+Timeout is an inherited feature provided by concurrent futures. Additionally,
+one wait for all the threads/process to complete even when on thread or process
+encounters an exception. This is useful when multiple test modules are
+executing different test scenarios.
 
-log = Log(__name__)
+When a test module controls the threads then it can forcefully terminate all
+threads when an exception is encountered.
+
+Changelog:
+    Version 1.0 used gevent module for parallel method execution.
+    Version 2.0 uses concurrent.futures module instead of gevent.
+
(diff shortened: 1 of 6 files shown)


sunilkumarn417 (commented):
Hi @psathyan 
I think it would be good if we move this to new separate module as `concurrent.py` to avoid any issues during releases and slowly we can bring this to surface.

ceph/parallel.py
```diff
@@ -1,121 +1,95 @@
...
-        if greenlet.successful():
-            self.results.put(greenlet.value)
-        else:
-            self.results.put(greenlet.exception)
-
-        self.count -= 1
-        if self.count <= 0:
-            self.results.put(StopIteration())
+        for r in self._results:
+            yield r
```
sunilkumarn417, line 95: I suggest one more method for closing all threads when exit status runs bad with custom option to set that flag.

psathyan: Thanks @sunilkumarn417 

It is a good suggestion. Let me add it.

psathyan:
> Hi @psathyan I think it would be good if we move this to new separate module as `concurrent.py` to avoid any issues during releases and slowly we can bring this to surface.

@sunilkumarn417 moving to a different module forces us to touch all files using `parallel`. If the concern is about release impact... then that is addressed via the merge to `main` instead of `master`.

Once all the planned framework changes are done, we will switch off `master` and use only `main`. This way we wouldn't be impacting the on-going release activities.

ceph/parallel.py
psathyan: Done. There are two options that are natively supported. The latest commit exposes those options to the end user.

This way when an exception is encountered, we can control the behavior.

AmarnatReddy (approved):
LGTM

AmarnatReddy (approved):

PR #4241, opened 2024-11-20
Enhance method to fetch OSD list & Enable faster recovery during PG merge

harshkumarRH:
PR addresses the following two changes:

- Enhances existing method to fetch active OSDs to be able to provide list of OSD depending on the status argument provided by the user.
- TFA fix for no-recover flag test case: Recently it was observed that test would timeout just as soon as desired number of PGs were reached, to enable faster recovery and PG split and merge, recovery flags are being modified and timeout has been slightly increased.

### DIFF
--- a/tests/rados/test_norecovery_flag_functionality.py
+++ b/tests/rados/test_norecovery_flag_functionality.py
@@ -1,7 +1,7 @@
 """
 This file contains the  methods to verify the  norecover flag functionality.
 Bug- https://bugzilla.redhat.com/show_bug.cgi?id=2134786
-1.Checking that norecover falg  should not perform any autoscale on PG
+1.Checking that norecover flag  should not perform any autoscale on PG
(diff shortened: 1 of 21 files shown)


tests/rados/test_norecovery_flag_functionality.py
```diff
@@ -54,17 +56,17 @@ def check_pg_increase(pool_name):
                     log.debug(
                         f"PG count on the pool not reached desired levels,"
                         f"current count : {current_pg_num}, target count : {pg_num_final}"
-                        f" sleeping for 10 seconds and checking again"
+                        f" sleeping for 60 seconds and checking again"
                     )
-                    time.sleep(10)
+                    time.sleep(60)
```
pdhiran, line 61: Thanks for increasing the timeout. Will help with reducing log lines.

pdhiran (approved):
Approved

PR #4251, opened 2024-11-25
Cleanup fix

aarsharm:
Data pools were not getting cleaned up, hence added a block to list test pools in a cluster and cleanup them.
Pass logs: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-LD1NWW


tests/rbd/rbd_snap_clone_imported_image.py
```diff
@@ -58,9 +60,20 @@ def test_snap_clone(rbd, pool_type, **kw):
     finally:
         if not kw.get("config").get("do_not_cleanup_pool"):
             rbd.clean_up(pools=[pool])
+            config = kw["config"]
+            cephadm = CephAdmin(cluster=ceph_cluster, **config)
+            rados_obj = RadosOrchestrator(node=cephadm)
+            list_of_pools = rados_obj.list_pools()
```
sunilangadi2, line 66: Instead of adding this logic in each test can we add this logic in `rbd.clean_up` script itself,
so that it will be applicable for all other tests which leaves the pool at the end.

sunilangadi2 (changes requested):
thanks @aarsharm for the quick solution, but let's focus on all other tests as well,
also please test and attach the success logs

tests/rbd/rbd_snap_clone_imported_image.py
aarsharm: Added logic in rbd.clean_up script. Have put it under a flag to make sure other scripts do not regress which are intentionally just cleaning up specific pools.

aarsharm:
Pass logs : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-XRT6YP/snap_and_clone_on_imported_image_0.log

tests/rbd/rbd_utils.py
```diff
@@ -699,6 +699,24 @@ def clean_up(self, **kw):
...
+            if flag == "all":
+                out = self.exec_cmd(
+                    cmd="ceph df detail --format json-pretty", output=True
+                )
+                status = json.loads(out.rstrip())
+                list_of_pools = [entry["name"] for entry in status["pools"]]
+                list_of_pools = [
+                    x
+                    for x in list_of_pools
+                    if not (x.startswith(".") or x.startswith("default"))
```
sunilangadi2, line 712: we should not delete `device_health_metrics` pool also
```[root@ceph-aarti-kbz-srezbz-node4 ~]# ceph df
--- RAW STORAGE ---
CLASS     SIZE    AVAIL    USED  RAW USED  %RAW USED
hdd    135 GiB  100 GiB  35 GiB    35 GiB      25.96
TOTAL  135 GiB  100 GiB  35 GiB    35 GiB      25.96

--- POOLS ---
POOL                   ID  PGS   STORED  OBJECTS     USED  %USED  MAX AVAIL
device_health_metrics   1    1    876 B        9  2.6 KiB      0     27 GiB
.rgw.root               2   32  1.4 KiB        4   48 KiB      0     27 GiB
default.rgw.log         3   32  3.6 KiB      209  408 KiB      0     27 GiB
default.rgw.control     4   32      0 B        8      0 B      0     27 GiB
default.rgw.meta        5   32    382 B        2   24 KiB      0     27 GiB
pool1                   7   32  1.6 GiB   10.24k   33 GiB  28.88     27 GiB```

File: tests/rbd/rbd_snap_clone_imported_image.py [hunk 3]
```diff
@@ -57,7 +57,7 @@ def test_snap_clone(rbd, pool_type, **kw):
 
     finally:
         if not kw.get("config").get("do_not_cleanup_pool"):
-            rbd.clean_up(pools=[pool])
+            rbd.clean_up(pools=[pool], flag="all")
```
sunilangadi2, line 60: then we should add this way for all other existing tests too,
can you think some other solution?

sunilangadi2: Does it ok without `flag` as well, 
anyway we have to delete the pools that were created during that test 
@sunilkumarn417 Any thoughts?

sunilangadi2 (changes requested):

tests/rbd/rbd_utils.py
aarsharm: Added the check

tests/rbd/rbd_snap_clone_imported_image.py
aarsharm: Pass logs:  http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-AZMF2S

sunilangadi2: thanks @aarsharm this solution looks good to me,
but can you add this way for all other existing tests where ever EC pools exist.
bcz this is essential for bare-metal runs

aarsharm: Sure will check and add

sunilangadi2: ok

sunilangadi2 (approved):
LGTM.

PR #4323, opened 2024-12-23
Switch mergify to support main branch

psathyan:
# Description

The default branch is main hence modifying mergify to work with main instead of master branch.


tintumathew10 (approved):
/lgtm

PR #4324, opened 2024-12-23
Test Automation : Live migration of RAW format images with encryption. CEPH-83596444

chaitanyadommeti:
# Description

Module for test "Live migration of RAW format images with encryption. CEPH-83596444" :  tests/rbd/test_rbd_migrate_encryption.py
Modified suite file : suites/squid/rbd/tier-3_rbd_migration.yaml

Success log : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-7L878I/

A new single PR (#4338) has been raised  for tests CEPH-83596444 and CEPH-83596588. 
Please ignore this if you are merging PR #4338. Will close this PR once PR #4338 is merged.

PR #4338, opened 2024-12-27
Test Automation for CEPH-83596444 and CEPH-83596588

chaitanyadommeti:
# Description
This is a single module for the following tests:
 CEPH-83596444: Live migration of RAW format images with encryption
 CEPH-83596588: Live migration of qcow2 format images with encryption

Success logs for : 
CEPH-83596444: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-PJL2Y1/
CEPH-83596588: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-521A51/

Also, added the test config  for both tests in suites/squid/rbd/tier-3_rbd_migration.yaml

Note: Initially PR(https://github.com/red-hat-storage/cephci/pull/4324) has been raised for the test :  CEPH-83596444. Since  CEPH-83596588 is similar to the test CEPH-83596444, have created a new file which handles both the tests. 
if we are merging this PR, the previous PR(4324) can be ignored(closed).

### DIFF
--- a/suites/squid/rbd/tier-3_rbd_migration.yaml
+++ b/suites/squid/rbd/tier-3_rbd_migration.yaml
@@ -1,6 +1,6 @@
 tests:
 
-  # Setup the cluster
+  #Setup the cluster
   - test:
       abort-on-fail: true
       module: install_prereq.py
@@ -63,7 +63,7 @@ tests:
       polarion-id: CEPH-83573758
 
   - test:
-      desc: Install rbd-nbd and remove any epel packages
+      desc: Install rbd-nbd, qemu-img and remove any epel packages
       module: exec.py
       name: Install rbd-nbd
       config:
@@ -71,6 +71,7 @@ tests:
         commands:
           - "rm -rf /etc/yum.repos.d/epel*"
           - "dnf install rbd-nbd -y"
+          - "dnf install qemu-img -y"
 
   - test:
       name: Test image migration with external qcow data format
@@ -90,3 +91,49 @@ tests:
         fio:
           size: 100M
       polarion-id: CEPH-83584070
+
+  - test:
+      desc: Live migration of RAW format images with encryption
+      config:
+        encryption_type:
+          - luks1
+          - luks2
+        rep_pool_config:
+          num_pools: 1
+          do_not_create_image: true
+        ec_pool_config:
+          num_pools: 1
+          do_not_create_image: true
+        fio:
+          size: 10M
+          fs: ext4
+          io: true
+        testconfig:
+          id: CEPH-83596444
+          name: Live migration of raw format images with encryption
+      module: test_rbd_live_migrate_encryption.py
+      name: Live migration of RAW format images with encryption
+      polarion-id: CEPH-83596444
+
+  - test:
+      desc: Live migration of qcow2 format images with encryption
+      config:
+        encryption_type:
+          - luks1
+          - luks2
+        rep_pool_config:
+          num_pools: 1
+          do_not_create_image: true
+        ec_pool_config:
+          num_pools: 1
+          do_not_create_image: true
+        fio:
+          size: 10M
+          fs: ext4
+          io: true
+        testconfig:
+          id: CEPH-83596588
+          name: Live migration of qcow2 format images with encryption
+      module: test_rbd_live_migrate_encryption.py
+      name: Live migration of qcow2 format images with encryption
+      polarion-id: CEPH-83596588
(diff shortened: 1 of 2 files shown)


suites/squid/rbd/tier-3_rbd_migration.yaml
```diff
@@ -90,3 +91,49 @@ tests:
...
+          num_pools: 1
+          do_not_create_image: true
+        ec_pool_config:
+          num_pools: 1
+          do_not_create_image: true
+        fio:
+          size: 10M
+          fs: ext4
+          io: true
+        testconfig:
```
HaruChebrolu, line 134: We may not need `testconfig` since we can use `desc` and `polarionId` to represent the same attributes when dealing with them in test modules. May I know if there is any reason for having this here?

tests/rbd/test_rbd_live_migrate_encryption.py
```diff
@@ -0,0 +1,539 @@
...
+                raw_spec = {
+                    "type": "raw",
+                    "stream": {"type": "file", "file_path": f"{raw_file}"},
+                }
+
+                kw["cleanup_files"].append(raw_file)
+
+                kw[pool].update({"spec": raw_spec})
+                err = migrate_check_consistency(rbd, pool, image, **kw)
+                if err:
```
HaruChebrolu, line 174: I feel, its better to log an error message whenever an error occurs. This way, issues can be easily identified and diagnosed through the automation logs.

tests/rbd/test_rbd_live_migrate_encryption.py
```diff
@@ -0,0 +1,539 @@
...
+                image = (
+                    "image_" + kw[pool]["encryption_type"] + "_" + random_string(len=3)
+                )
+                kw[pool].update({"image": image})
+                err = run_io_on_encryption_formatted_image(rbd, pool, image, **kw)
+                if err:
+                    return 1
+
+                # Create spec file raw data
+                raw_file = "/tmp/" + image
```
HaruChebrolu, line 158: Can we use `tempfile` here instead of hardcoding

`raw_file = tempfile.mktemp(prefix=f"{image}_", suffix=".raw")`

tests/rbd/test_rbd_live_migrate_encryption.py
```diff
@@ -0,0 +1,539 @@
...
+                return 1
+            log.info(f"Test {kw['config']['testconfig']['name']} is successful")
+
+    except Exception as e:
+        log.error(f"Test {kw['config']['testconfig']['name']} failed: {str(e)}")
+        return 1
+
+    finally:
+        for file in kw["cleanup_files"]:
+            out, err = client.exec_command(sudo=True, cmd=f"rm -f {file}")
```
HaruChebrolu, line 532: If client.exec_command fails to delete a file or cleanup fails to delete a pool or image, no error is logged, and the function continues. This might leave:

- Temporary files in /tmp, consuming disk space.
- Pools or images in the Ceph cluster, leading to resource wastage.
- Inconsistent states in the system that complicate subsequent tests.

so we can have one try block for deleting temporary files and another try block for resource cleanup. WDYT?

suites/squid/rbd/tier-3_rbd_migration.yaml
chaitanyadommeti: I dont find the keys/values for 'desc' and 'polarionId' in the arguments passed to the run() in test module. 
Test details like description, polarion id, name are not passed to test_mod.run() of cephci/run.py (line# 826-848).

Hence, added those details under config, which is getting passed. I did not want to modify cephci/run.py tp pass these, as everyone uses this.

Let me know if i am missing something or can be implemented in any other better way.

tests/rbd/test_rbd_live_migrate_encryption.py
chaitanyadommeti: the function definition for migrate_check_consistency() is in the same module, where the error messages are logged for each function call inside it, before it returns to the calling function, in case of error. Hence, i thought the same error message would indicate the actual error occured and would be redundant if logged again. Let me know if you think otherwise. we can discuss.

tests/rbd/test_rbd_live_migrate_encryption.py
chaitanyadommeti: Thanks for letting me know this. Wasnt aware of it.  Fixed it

tests/rbd/test_rbd_live_migrate_encryption.py
chaitanyadommeti: Good Catch. Thanks for pointing. Its good to have a separate try block, to avoid missing cleanup of other entities. Fixed.

chaitanyadommeti:
Success log after implementing review comments: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-40ZWUX/

sunilkumarn417 (approved):

PR #4358, opened 2025-01-07
Fix mergify configuration

psathyan:
# Description

With the fix of code owners, we removed the top level reviewers group. This enabled us to align with the purpose of codeowners file.

After the fix merged, the mergify configuration was left unchanged resulting in a broken condition.

This PR fixes the mergify conditions and moves the merge mechanism to queue.


vamahaja (approved):

tintumathew10 (commented):
LGTM

PR #4377, opened 2025-01-15
[Conf] Fix BVT & Sanity cluster conf with colocation

vamahaja:
BVT & Sanity common cluster is not folling colocation guide. Update deployment node roles based on the recommendations.


obannak (approved):
/approve

PR #4390, opened 2025-01-20
Fixing Rados CIDR blocklisting test suite post Cluster config consolidation

pdhiran:
1. The client nodes used in CIDR blocklisting test had to be updated post consolidating the test conf files in rados. Updated the client nodes in the suite file
2. Updated the stretch suite conf for new conf, and also added the same conf for RHOS-D env


neha-gangadhar (approved):
Approved

PR #4421, opened 2025-02-04
Adhering to change in behavior for mirror token creation

AmarnatReddy:
# Description
Adhering to change in behavior for mirror token creation 
**Problem :** 
According to [Bugzilla #2304288](https://bugzilla.redhat.com/show_bug.cgi?id=2304288), there is a change in behavior from the previous build to Squid.

**Solution:**
To address this, we have added an additional -n option when creating the token for users with read (r), read-write (rw), or read-write-plus (rwp) permissions to MDS. The token creation should now fail for these users as expected.

Logs :  http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-UP346E/

<details>

</details>

### DIFF
--- a/tests/cephfs/cephfs_mirroring/test_cephfs_mirror_auth_caps.py
+++ b/tests/cephfs/cephfs_mirroring/test_cephfs_mirror_auth_caps.py
@@ -1,12 +1,14 @@
 import random
 import string
 import traceback
+from distutils.version import LooseVersion
 from itertools import product
 
 from ceph.ceph import CommandFailed
 from tests.cephfs.cephfs_mirroring.cephfs_mirroring_utils import CephfsMirroringUtils
 from tests.cephfs.cephfs_utilsV1 import FsUtils
 from utility.log import Log
+from utility.utils import get_ceph_version_from_cluster
 
 log = Log(__name__)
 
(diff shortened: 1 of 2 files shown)


tests/cephfs/cephfs_mirroring/test_cephfs_mirror_auth_caps.py
Manimaran-MM, line 1: Should the same logic need to be implemented on line number 528 as well?

Manimaran-MM (commented):
This lib `tests/cephfs/cephfs_mirroring/cephfs_mirroring_utils.py` also has the command. 

Can we make changes to only this lib and call this function?

AmarnatReddy:
> This lib `tests/cephfs/cephfs_mirroring/cephfs_mirroring_utils.py` also has the command.
> 
> Can we make changes to only this lib and call this function?

In Utils we want to create with admin only so we don't need to change the command

tests/cephfs/cephfs_mirroring/test_cephfs_mirror_auth_caps.py
AmarnatReddy: Updated the code. there was escape of the error in below code. after raising CommandFailed exception. it was get escaped because of nested exception blocks in **run** method and **perform_fs_mirroring_operations**. 
Now in run method also I am raising the CommandFailed exception

Thanks mani

Manimaran-MM (approved):
LGTM

PR #4422, opened 2025-02-05
[RADOS] TFA fix for OSD superblock redundancy false failure

harshkumarRH:
# Description

Sleep time provided during OSD restart was not enough to let the OSD boot up, so before OSD could start, we were trying to execute the next desired command using ceph-bluestore-tool through CBT automated module. The CBT run command stops the OSDs and then proceeds to the run the command, but when it tries to stop the OSD, it finds that OSD is already in `stopped` state because it is yet to boot up after `systemctl restart` has been triggered, now by the time the actual command gets executed, OSD is UP and therefore the execution fails as OSD is supposed to be in `stopped` state.

Pass logs -
Reef: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-GBYAHU
Squid:

### DIFF
--- a/ceph/rados/core_workflows.py
+++ b/ceph/rados/core_workflows.py
@@ -1440,50 +1440,44 @@ class RadosOrchestrator:
         )
         host.exec_command(sudo=True, cmd=cmd)
         # verifying the osd state
-        if action in ["start", "stop"]:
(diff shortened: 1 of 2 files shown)


pdhiran (approved):
Looks good

PR #4486, opened 2025-02-27
Updated rados upgrade paths to include n, n-1 & n-2 paths in all releases

pdhiran:
Updated upgrade paths in rados suites. Paths covered : 

Pacific :
5.3 z5 -> 5.x latest
5.3 CDN -> 5.x latest

Quincy:
5.3 z6 -> 6.x latest
5.3 CDN -> 6.x latest
6.1 CDN -> 6.x latest

Reef:
6.1 CDN -> 7.x latest
5.3 CDN -> 7.x latest
7.1 CDN -> 7.x latest

Squid:
7.1 z1 -> 8.x latest
6.1 CDN -> 8.x latest
8.0 CDN -> 8.x latest

### DIFF
--- a/suites/quincy/rados/tier-2_rados_ec-pool_recovery.yaml
+++ b/suites/quincy/rados/tier-2_rados_ec-pool_recovery.yaml
@@ -27,6 +27,8 @@ tests:
                 registry-url: registry.redhat.io
                 mon-ip: node1
                 allow-fqdn-hostname: true
+                rhcs-version: 6.1
+                release: rc
           - config:
               command: add_hosts
               service: host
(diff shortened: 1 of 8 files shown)


neha-gangadhar:
@pdhiran The Z stream build from which we are upgrading should also should be the latest z-stream before upgrading right? Unless we have a specific customer issue we are verifying from that particular Z-version. What do you think?

neha-gangadhar (approved):
Looks good

neha-gangadhar (approved):
Approved

suites/quincy/rados/tier-2_rados_ec-pool_recovery.yaml
```diff
@@ -27,6 +27,8 @@ tests:
                 registry-url: registry.redhat.io
                 mon-ip: node1
                 allow-fqdn-hostname: true
+                rhcs-version: 6.1
+                release: rc
```
harshkumarRH, line 31: What happens if the test suite is executed for 6.1 RC build itself? Shall we consider using an older z-stream version here, something which is not dynamic?
6.1 z-stream -> upgrade to 6.1 RC -> upgrade to 6.1 latest nightly

PR #4495, opened 2025-02-28
Capturing the Cli commands

SaiSuriya30:
# Description

I have added subcommands.py to the utility files, similar to how the sos report is handled. This file is responsible for capturing the output of CLI commands (radosgw-admin) and storing them in JSON format. It helps extract, process, and save command outputs efficiently for further analysis.
i have executed and it is saving in the log directory
the log link :http://magna002.ceph.redhat.com/cephci-jenkins/results/openstack/RH/8.0/rhel-9.5/Test/19.2.0-98/267/tier-1_rgw/

the output is assed by the following
output stored:
http://magna002.ceph.redhat.com/cephci-jenkins/results/openstack/RH/8.0/rhel-9.5/Test/19.2.0-98/267/tier-1_rgw/cephci-commands-results/

### DIFF
--- a/utility/rgw_command_extractor.py
+++ b/utility/rgw_command_extractor.py
@@ -178,7 +178,7 @@ def save_to_remote(command, output, subcomponent_filter, output_directory):
         output_directory (str): The directory where JSON files will be saved.
     """
     output_hash = compute_output_hash(output)
-    base_dir = os.path.join(output_directory, "cephci-commands-results")
+    base_dir = os.path.join(output_directory, "rgw_command_extractor")
     os.makedirs(base_dir, exist_ok=True)
 
     subcommand_match = re.search(r"radosgw-admin (\w+)", command)
--- a/utility/rgw_command_extractor.py
+++ b/utility/rgw_command_extractor.py
@@ -78,8 +78,8 @@ DOC = """
 Standard script to fetch and process log files from a given directory.
 
 Usage:
-    subcommands.py --logdir <log_directory> --filter <subcomponent_filter> --outdir <output_directory>
-    subcommands.py (-h | --help)
+    rgw_command_extractor.py --logdir <log_directory> --filter <subcomponent_filter> --outdir <output_directory>
+    rgw_command_extractor.py (-h | --help)
 
 Options:
     -h --help                      Show this help message
--- a/utility/rgw_command_extractor.py
+++ b/utility/rgw_command_extractor.py
@@ -72,6 +72,7 @@ import hashlib
 import json
 import os
 import re
+
 from docopt import docopt
 
 DOC = """
--- a/utility/rgw_command_extractor.py
+++ b/utility/rgw_command_extractor.py
@@ -22,9 +22,6 @@ It ensures efficient log parsing, prevents duplicate storage, and organizes outp
 5. **Duplicate Detection:**
    - Computes SHA-256 hashes for command outputs to prevent redundant storage.
 
-6. **Error Handling:**
-   - Manages file I/O operations and JSON parsing errors gracefully.
-
 ### **How the Script Works:**
 1. **Fetching Log Files:**
    - The get_log_files_from_directory function scans the given directory and collects all .log files.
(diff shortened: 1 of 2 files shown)


yuva-gayam (approved):

utility/subcommands.py
```diff
@@ -0,0 +1,229 @@
...
+                    existing_hashes.add(output_hash)
+            current_command, current_output_lines = cmd_match.group(0).strip(), []
+        else:
+            current_output_lines.append(cleaned_line)
+
+    return {"outputs": results}
+
+
+def save_to_remote(command, output, subcomponent_filter, output_directory):
+    """Save extracted command outputs to the specified output directory."""
```
anrao19, line 147: can we have documentation for each method..

run.py
```diff
@@ -486,6 +488,7 @@ def run(args):
     enable_eus = args.get("--enable-eus")
     skip_enabling_rhel_rpms = args.get("--skip-enabling-rhel-rpms")
     skip_sos_report = args.get("--skip-sos-report")
+    skip_subcommands = args.get("--skip-subcommands")
```
vamahaja, line 491: we already have lot of parameters to run.py, kindly avoid adding unnecessary parameters.

vamahaja (changes requested):

run.py
vamahaja: @psathyan ^^^^

anrao19: Noted @vamahaja we will run this only during rc and we will remove this parameter

run.py
```diff
@@ -1030,6 +1030,8 @@ def fetch_test_details(var) -> dict:
             get_ceph_var_logs(ceph_cluster_dict[cluster], run_dir)
         log.info(f"Generated sosreports location : {url_base}/sosreports\n")
 
+    if build == rc:
```
vamahaja, line 1033: 1. Why this needs to be part of `run.py` ? This will unnecessary add steps to process these files.
2. I don't understand how magna link will be formed even before copying data to the magna server.

vamahaja (changes requested):
I would suggest to add new script `cephci/generate_log_report.py` which will be initiated manually post test completion with log-dir path.

run.py
mkasturi18: @vamahaja , can we have a sync up to bring this to closure.

SaiSuriya30: the logs that are present in the run_dir is passed as the input .So that with those files we can analyse and store them along with the suites like the sos report itself

utility/subcommands.py
```diff
@@ -0,0 +1,264 @@
+"""
```
psathyan, line 1: Any reason why we named this file as `subcommands`? IMO, it does not match with the purpose or objective of the tasks performed.

utility/subcommands.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+    Processes a single log file for command extraction and saves the outputs.
+    Args:
+        file_path (str): The path to the log file.
+        subcomponent_filter (str): The subcomponent filter.
+        output_directory (str): The directory where extracted data will be saved.
+    """
+    try:
+        with open(file_path, "r") as file:
+            log_lines = file.readlines()
```
psathyan, line 235: In case of large size log files, the memory is going to ballon. Please read a chunk so that we don't overload the memory.

utility/subcommands.py
```diff
@@ -0,0 +1,264 @@
...
+            extracted_data = extract_radosgw_admin_commands(log_lines)
```
psathyan, line 236: This becomes `component` specific. IMO, we should check if it is `rgw` component and then only call this method.

utility/subcommands.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+    results, existing_hashes = [], set()
+    current_command, current_output_lines = None, []
+
+    for line in log_lines:
+        cleaned_line = clean_log_line(line)
+        if not cleaned_line:
+            continue
+
+        cmd_match = re.search(r"(?:sudo )?radosgw-admin[^;\n]+", cleaned_line)
```
psathyan, line 150: Can we redesign... this will cause issues as the Jenkins agents don't have large RAM memory.

utility/subcommands.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+    line = re.sub(
+        r"^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2},\d{3} - cephci - ceph:\d+ - (INFO|DEBUG|ERROR) -\s*",
+        "",
+        line,
+    ).strip()
+    line = re.sub(
+        r"^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2},\d{3} (INFO|DEBUG|ERROR):?\s*", "", line
+    ).strip()
+    return line if line else None
```
psathyan, line 116: I think this would be best done in `ceph-qe-scripts`. Over there, use a log filter to provide a separate artifact during runtime.

run.py
psathyan: IMHO, we shouldn't being doing this here. There are multiple other viable options 
- Custom log filters
- Separate script file that can be triggered when there is a signed built that can be execute asynchrously

This increases the job execution time which I believe can be avoided.

psathyan (changes requested):

mkasturi18:
Adding DNM & review-in-progress labels, as there are lot of changes requested.

utility/rgw_command_extractor.py
```diff
@@ -0,0 +1,277 @@
...
+        file_path (str): The path to the log file.
+        subcomponent_filter (str): The subcomponent filter.
+        output_directory (str): The directory where extracted data will be saved.
+    """
+    try:
+        with open(file_path, "r") as file:
+            buffer = []
+            for line in file:
+                buffer.append(line)
+                if len(buffer) >= chunk_size:
```
vamahaja, line 238: Instead of reading file in chunks, we can follow below pattern with readlines() -

Step 1 - Executing command -
  INFO: executing cmd: sudo radosgw-admin <parameters>
Step 2 - Execution Status
  Successfull - INFO: cmd excuted
  Failed - ERROR: cmd execution failed
Step 3 - Output
  Successfull - INFO: <multiline-output>
  Failed - ERROR: stderr: <multiline-output>

vamahaja (changes requested):

SaiSuriya30:
closing thus pr due to rebase issue

PR #4504, opened 2025-03-04
Fix issue with masking of sensitive data

psathyan:
# Description

Tuple is an immutable data structure. Filtering of log messages attempted to modify an immutable data structure resulting in an error.

This commit modifies the approach on encountering this datatype.

- [ ] Unit testing.


tintumathew10 (approved):
lgtm

PR #338, opened 2021-02-05
code changes service deployment using orchestration apply option

sunilkumarn417:
- Added service deployment code changes using "ceph orch apply <svc>" option.
- Updated sanity test suite.
- Updated service validation check.

tested sanity and deployment test suite in local.
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1612511527005/ (deployment)

Sanity test result: http://magna002.ceph.redhat.com/ceph-qe-logs/Sunil/cephadm-sanity-apply-1.log
```
TEST NAME                        TEST DESCRIPTION                                               DURATION                                  STATUS
Install ceph pre-requisites      installation of ceph pre-requisites                            0:05:52.391169                              Pass
Cephadm Bootstrap                cephadm cluster bootstrap                                      0:05:20.477147                              Pass
Add Host                         Add host to Bootstrapped Cluster                               0:00:13.285115                              Pass
Add host                         Add new host node with IP address                              0:00:10.672232                              Pass
Add host                         Add new host node with IP address and labels                   0:00:14.128434                              Pass
Remove Host label                Remove host label(s) from existing node                        0:00:20.229243                              Pass
Add Host label                   Add host label(s) to existing node                             0:00:30.251443                              Pass
Attach Host label to installer   Add host label(s) to existing installer node                   0:00:49.558146                              Pass
Set Host address                 Set IP address to a existing node                              0:00:10.118256                              Pass
Remove host                      Remove node from cluster                                       0:00:10.453463                              Pass
Remove Hosts                     Remove node from cluster                                       0:00:37.254187                              Pass
Add host                         Add all host node with IP address and labels                   0:01:59.115221                              Pass
Apply Monitor Service            Apply monitor service on all mon nodes                         0:00:55.602548                              Pass
Apply Manager Service            Apply manager service on all mgr nodes                         0:00:07.368161                              Pass
Apply OSD Service                Apply OSD service on all osd nodes                             0:00:49.770144                              Pass
Apply MDS Service                Apply MDS service on all mds nodes                             0:00:20.139781                              Pass
Apply ISCSI Service              Apply ISCSI target service on iscsi role nodes                 0:00:22.966621                              Pass
Apply Prometheus Service         Apply Prometheus service on role nodes                         0:00:39.231906                              Pass
Apply Node-exporter Service      Apply Node-exporter service on role nodes                      0:02:07.063138                              Pass
[...]

Issue #135, opened 2020-06-05
Interop suite 3 intermittent failures (config roll over task)


XDsubhash:
http://report-service-prod-interop.cloud.paas.psi.redhat.com/emails/7bf1edff2352469297ce0ddc9c88580c

veera-raghava-reddy:
Inter-op tests are passing - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-1614227552915/result.html

PR #4563, opened 2025-03-17
Module changes for group mirroring

chaitanyadommeti:
# Description

<details>

</details>

### DIFF
--- a/ceph/rbd/initial_config.py
+++ b/ceph/rbd/initial_config.py
@@ -469,7 +469,8 @@ def initial_rbd_config(ceph_cluster, **kw):
         ):
             log.error(f"RBD configuration failed for {pool_type}")
             return None
-        if kw.get("config").get(pool_type).get("rbd_mirror_group") is True:
+        # Create a group in case of group level mirroring.
+        if kw.get("config").get(pool_type).get("mirror_level") == "group":
             create_mirror_group(rbd, client, pool_type, **kw)
 
     return {"rbd": rbd, "client": client, "pool_types": pool_types}
--- a/ceph/rbd/initial_config.py
+++ b/ceph/rbd/initial_config.py
@@ -1,4 +1,5 @@
 from ceph.rbd.utils import getdict, random_string
+from ceph.rbd.workflows.group import create_mirror_group
 from ceph.rbd.workflows.rbd import create_pools_and_images
 from ceph.rbd.workflows.rbd_mirror import config_mirror_multi_pool
 from ceph.utils import get_node_by_id
@@ -97,6 +98,11 @@ def update_config(**kw):
             is_pool_config_updated = [
                 k for k in getdict(rep_pool_config).keys() if k not in ["test_config"]
             ]
+            if config.get("grouptype").startswith("single"):
+                rep_pool_config.update({"num_pools": 1})
+            elif config.get("grouptype").startswith("multi"):
+                if not config.get("num_pools") or rep_pool_config.get("num_pools") == 1:
+                    rep_pool_config.update({"num_pools": 2})
 
             if not is_pool_config_updated and rep_pool_config.get("num_pools"):
                 for i in range(rep_pool_config.get("num_pools")):
@@ -212,6 +218,12 @@ def update_config(**kw):
             is_pool_config_updated = [
                 k for k in getdict(ec_pool_config).keys() if k not in ["test_config"]
             ]
+            if config.get("grouptype").startswith("single"):
+                ec_pool_config.update({"num_pools": 1})
+            elif config.get("grouptype").startswith("multi"):
+                if not config.get("num_pools") or ec_pool_config.get("num_pools") == 1:
+                    ec_pool_config.update({"num_pools": 2})
+
             if not is_pool_config_updated and ec_pool_config.get("num_pools"):
                 for i in range(ec_pool_config.get("num_pools")):
                     pool = ec_pool_config.get(
@@ -457,6 +469,8 @@ def initial_rbd_config(ceph_cluster, **kw):
         ):
             log.error(f"RBD configuration failed for {pool_type}")
             return None
+        if kw.get("config").get(pool_type).get("rbd_mirror_group") is True:
+            create_mirror_group(rbd, client, pool_type, **kw)
 
     return {"rbd": rbd, "client": client, "pool_types": pool_types}
 
--- a/ceph/rbd/workflows/group.py
+++ b/ceph/rbd/workflows/group.py
@@ -1,10 +1,12 @@
 from copy import deepcopy
 
-from ceph.rbd.utils import random_string
+from ceph.rbd.utils import getdict, random_string
 from cli.rbd.rbd import Rbd
 from utility.log import Log
 
 log = Log(__name__)
+
+from ceph.rbd.workflows.namespace import create_namespace_and_verify
 
 
 def create_group_and_verify(**kw):
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
     # Group info
     (group_snap_out, group_snap_err) = rbd.group.snap.info(**group_snap_info_kw)
     return (group_snap_out, group_snap_err)
+
+
+def create_mirror_group(rbd, client, pool_type, **kw):
+    """
+    Creates a single group for mirroring.
+
+    Args:
+        rbd (module): The rbd object
+        client (object): client object.
+        pool_type (str): The type of pool to create the mirror group for.
+        kw (dict): A dictionary of keyword arguments.
+
+    Returns:
+        int: 0 if the mirror group was created successfully, otherwise a non-zero value.
+    """
+    size = kw["config"][pool_type]["size"]
+    rbd_config = kw.get("config", {}).get(pool_type, {})
+    multi_pool_config = getdict(rbd_config)
+    group_created = None
+    namespace = None
+    for pool, pool_config in multi_pool_config.items():
+        group = "gp_" + pool.split("_")[-1]
+        group_create_kw = {"client": client, "pool": pool, "group": group}
+        grouptype = kw.get("config").get("grouptype", "")
+        if grouptype in {"single_pool_with_namespace", "multi_pool_with_namespace"}:
+            if not kw.get("config", {}).get(pool_type, {}).get("group-namespace"):
+                if kw.get("is_secondary", False) is False:
+                    namespace = "ns_" + random_string(len=3)
+                    group_create_kw.update({"namespace": namespace})
+                    group_spec = f"{pool}/{namespace}/{group}"
+                    kw["config"][pool_type].get(pool, {}).update(
+                        {"namespace": namespace}
+                    )
+                else:
+                    namespace = (
+                        kw["config"][pool_type].get(pool, {}).get("namespace", None)
+                    )
+            else:
+                # Namespace images from the same pool or mutiple pools can be added to a group only if namespaces match
+                # The first namespace created is used as the namespace for the groups as well as pools
+                # on both primary and secondary side.
+                namespace = (
+                    kw.get("config", {}).get(pool_type, {}).get("group-namespace")
+                )
+            kw["config"][pool_type].get(pool, {}).update(
+                {"remote_namespace": namespace}
+            )
+            rc = create_namespace_and_verify(
+                **{"pool-name": pool, "namespace": namespace, "client": client}
+            )
+            if rc != 0:
+                return rc
+        else:
+            group_spec = f"{pool}/{group}"
+
+        if kw.get("is_secondary", False) is False:
+            if group_created is not True:
+                # A single group is created that contains images from one pool or from multiple pools
+                rc = create_group_and_verify(**group_create_kw)
+                if rc != 0:
+                    return rc
+                group_created = True
+                kw.get("config", {}).get(pool_type, {}).get(pool, {}).update(
+                    {"group-spec": group_spec}
+                )
+                kw.get("config", {}).get(pool_type, {}).update(
+                    {"mirror-group": group_spec}
+                )
+                if namespace:
+                    kw.get("config", {}).get(pool_type, {}).update(
+                        {"group-namespace": namespace}
+                    )
+                imagelevel = (
+                    kw.get("config", {}).get(pool_type, {}).get("mirror_level", {})
+                )
+                kw.get("config", {}).get(pool_type, {}).get(pool, {}).update(
+                    {"mirror_level": imagelevel}
+                )
+                kw["config"][pool_type].get(pool, {}).update({"group": group})
+
+            multi_image_config = getdict(pool_config)
+            for imagename in multi_image_config.keys():
+                if grouptype in {
+                    "single_pool_with_namespace",
+                    "multi_pool_with_namespace",
+                }:
+                    imagespec = f"{pool}/{kw.get('config', {}).get(pool_type, {}).get('group-namespace')}/{imagename}"
+                else:
+                    imagespec = f"{pool}/{imagename}"
+
+                rbd.create(**{"image-spec": imagespec, "size": size})
+                mirror_group_spec = (
+                    kw.get("config", {}).get(pool_type, {}).get("mirror-group", "")
+                )
+                # In case of multiple pools, images from all pools are added to the single group that was
+                # created on the first pool
+                rc = add_image_to_group_and_verify(
+                    **{
+                        "group-spec": mirror_group_spec,
+                        "image-spec": imagespec,
+                        "client": client,
+                    }
+                )
+                if rc != 0:
+                    return rc
+            kw["config"][pool_type].update({"group": group})
+    return 0
--- a/ceph/rbd/workflows/group.py
+++ b/ceph/rbd/workflows/group.py
@@ -359,22 +359,22 @@ def create_mirror_group(rbd, client, pool_type, **kw):
     group_created = None
     namespace = None
     for pool, pool_config in multi_pool_config.items():
-        group = "gp_" + pool.split("_")[-1]
-        group_create_kw = {"client": client, "pool": pool, "group": group}
-        grouptype = kw.get("config").get("grouptype", "")
+        group = "group_" + pool.split("_")[-1]
+        group_config = {"client": client, "pool": pool, "group": group}
+        grouptype = kw.get("config").get("grouptype", "single_pool_without_namespace")
         if grouptype in {"single_pool_with_namespace", "multi_pool_with_namespace"}:
             if not kw.get("config", {}).get(pool_type, {}).get("group-namespace"):
+                # Create namespace only on the primary site.
+                # For seconday site, same namespace as in primary is taken
                 if kw.get("is_secondary", False) is False:
-                    namespace = "ns_" + random_string(len=3)
-                    group_create_kw.update({"namespace": namespace})
+                    namespace = "namespace_" + random_string(len=3)
+                    group_config.update({"namespace": namespace})
                     group_spec = f"{pool}/{namespace}/{group}"
                     kw["config"][pool_type].get(pool, {}).update(
                         {"namespace": namespace}
                     )
                 else:
-                    namespace = (
-                        kw["config"][pool_type].get(pool, {}).get("namespace", None)
-                    )
+                    namespace = kw["config"][pool_type][pool]["namespace"]
             else:
                 # Namespace images from the same pool or mutiple pools can be added to a group only if namespaces match
                 # The first namespace created is used as the namespace for the groups as well as pools
@@ -396,7 +396,7 @@ def create_mirror_group(rbd, client, pool_type, **kw):
         if kw.get("is_secondary", False) is False:
             if group_created is not True:
                 # A single group is created that contains images from one pool or from multiple pools
-                rc = create_group_and_verify(**group_create_kw)
+                rc = create_group_and_verify(**group_config)
                 if rc != 0:
                     return rc
                 group_created = True
@@ -410,11 +410,11 @@ def create_mirror_group(rbd, client, pool_type, **kw):
                     kw.get("config", {}).get(pool_type, {}).update(
                         {"group-namespace": namespace}
                     )
-                imagelevel = (
+                mirror_level = (
                     kw.get("config", {}).get(pool_type, {}).get("mirror_level", {})
                 )
                 kw.get("config", {}).get(pool_type, {}).get(pool, {}).update(
-                    {"mirror_level": imagelevel}
+                    {"mirror_level": mirror_level}
                 )
                 kw["config"][pool_type].get(pool, {}).update({"group": group})
 
--- a/ceph/rbd/workflows/rbd_mirror.py
+++ b/ceph/rbd/workflows/rbd_mirror.py
@@ -406,72 +406,6 @@ def config_mirror(rbd_primary, rbd_secondary, **kw):
         )
 
 
-def enable_group_mirroring(primary_config, secondary_config, pool, **pool_config):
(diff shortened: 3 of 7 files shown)


ceph/rbd/workflows/rbd_mirror.py
```diff
@@ -105,6 +105,23 @@ def wait_for_status(rbd, cluster_name, **kw):
                 )
                 if kw.get("images_pattern") == num_image:
                     return 0
+        elif kw.get("groupspec", False):
+            if kw.get("state_pattern"):
+                out = value(
+                    "state",
+                    json.loads(
+                        rbd.mirror.group.status(
+                            **{"groupspec": kw.get("groupspec"), "format": "json"}
```
aarsharm, line 114: to maintain uniformity, please keep format outside of the spec as below.
**{"groupspec": kw.get("groupspec")}, format="json"

chaitanyadommeti: uniformity is maintained here already. We are keeping the entire arguments inside {}. 
same as in line # 84 or as in line #132. Please note that {} are not used in the line #84 and #132. Hence the confusion.

aarsharm: i could see format="json" as a separate param in below lines:
line #17 (status_config = {"image-spec": imagespec, "format": "json"})
line #70 (rbd.mirror.pool.status(pool=kw.get("poolname"), format="json")[)
line #80 (rbd.mirror.pool.status(pool=kw.get("poolname"), format="json")[)
line #135 (status_config = {
                        "image-spec": kw.get("imagespec"),
                        "format": "json",
                    })
line #163 status_config = {
                        "image-spec": kw.get("imagespec"),
                        "format": "json",
                    }
line #338 (rbd_primary.mirror.pool.info(
                pool=poolname, format="json", cluster=ceph_cluster_primary
            )[0])
 
 line #345             rbd_secondary.mirror.pool.info(
                pool=poolname, format="json", cluster=ceph_cluster_secondary
            )[0]
            
 line #669 (pool_config = {"pool": pool_name, "verbose": True, "format": "json"})
 
 line #736 (out, err = rbd.mirror.image.status(
                    pool=pool, image=image, format="json"
                ))

chaitanyadommeti: its still same as in examples listed out above. Lets discuss

aarsharm: as discussed offline, uniformity is maintained. Hence resolving the comment

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+            kw["config"][pool_type].get(pool, {}).update(
+                {"remote_namespace": namespace}
+            )
+            rc = create_namespace_and_verify(
+                **{"pool-name": pool, "namespace": namespace, "client": client}
+            )
+            if rc != 0:
+                return rc
+        else:
+            group_spec = f"{pool}/{group}"
```
aarsharm, line 394: Do we need to remove f strings everywhere ? as per paragadeeshwaran's ask? if yes, we have to modify all f strings used in this PR

cli/rbd/group.py
```diff
@@ -58,12 +58,14 @@ def info(self, **kw):
             Supported keys:
                 pool(str) : name of the pool into which namespace should be stored
                 namespace(str): namespace in the pool
-                pool-spec(str): <pool-name>[/<namespace>]
+                group-spec(str): <pool-name>[/<namespace>]group
```
aarsharm, line 61: change this to : <pool-name>[/<namespace>][/<group>]

cli/rbd/mirror/group.py
```diff
@@ -0,0 +1,108 @@
...
+        return self.execute_as_sudo(cmd=cmd)
+
+    def status(self, **kw):
+        """Wrapper for rbd mirror group status.
+        Args
+            kw: Key value pair of method arguments
+        Example::
+        Supported keys:
+            group-spec: poolname/[namespace]/groupname
+            See rbd help mirror group resync for more supported keys
```
aarsharm, line 102: Change this to "See rbd help mirror group status for more supported keys"

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+    multi_pool_config = getdict(rbd_config)
+    group_created = None
+    namespace = None
+    for pool, pool_config in multi_pool_config.items():
+        group = "gp_" + pool.split("_")[-1]
+        group_create_kw = {"client": client, "pool": pool, "group": group}
+        grouptype = kw.get("config").get("grouptype", "")
+        if grouptype in {"single_pool_with_namespace", "multi_pool_with_namespace"}:
+            if not kw.get("config", {}).get(pool_type, {}).get("group-namespace"):
+                if kw.get("is_secondary", False) is False:
```
sunilangadi2, line 367: please add comment why this check is needed

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+                    namespace = "ns_" + random_string(len=3)
```
sunilangadi2, line 368: `test_namespace_<random-string>` would be better

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+                    group_create_kw.update({"namespace": namespace})
+                    group_spec = f"{pool}/{namespace}/{group}"
```
sunilangadi2, line 370: Also does it considers 

> --pool --namespace --group-name

 options as well?

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+                    kw["config"][pool_type].get(pool, {}).update(
+                        {"namespace": namespace}
+                    )
+                else:
+                    namespace = (
```
sunilangadi2, line 375: the default `None` are we thinking empty default namespace here? if not the case we can't consider namespace as `None`

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+                        kw["config"][pool_type].get(pool, {}).get("namespace", None)
+                    )
+            else:
+                # Namespace images from the same pool or mutiple pools can be added to a group only if namespaces match
+                # The first namespace created is used as the namespace for the groups as well as pools
```
sunilangadi2, line 380: Does it constraints here only on first namespace?

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+                namespace = (
+                    kw.get("config", {}).get(pool_type, {}).get("group-namespace")
+                )
+            kw["config"][pool_type].get(pool, {}).update(
+                {"remote_namespace": namespace}
+            )
+            rc = create_namespace_and_verify(
+                **{"pool-name": pool, "namespace": namespace, "client": client}
+            )
+            if rc != 0:
```
sunilangadi2, line 391: can we add proper `error log` here then return rc

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+                return rc
+        else:
+            group_spec = f"{pool}/{group}"
+
+        if kw.get("is_secondary", False) is False:
+            if group_created is not True:
+                # A single group is created that contains images from one pool or from multiple pools
+                rc = create_group_and_verify(**group_create_kw)
+                if rc != 0:
```
sunilangadi2, line 400: same here

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+                    {"group-spec": group_spec}
+                )
+                kw.get("config", {}).get(pool_type, {}).update(
+                    {"mirror-group": group_spec}
+                )
+                if namespace:
+                    kw.get("config", {}).get(pool_type, {}).update(
+                        {"group-namespace": namespace}
+                    )
+                imagelevel = (
```
sunilangadi2, line 413: assigning `mirror_level` as `imagelevel` might confuse, it's good to use as `mirror-level` itself

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+                )
+                kw["config"][pool_type].get(pool, {}).update({"group": group})
+
+            multi_image_config = getdict(pool_config)
+            for imagename in multi_image_config.keys():
+                if grouptype in {
+                    "single_pool_with_namespace",
+                    "multi_pool_with_namespace",
+                }:
+                    imagespec = f"{pool}/{kw.get('config', {}).get(pool_type, {}).get('group-namespace')}/{imagename}"
```
sunilangadi2, line 427: can we take `imagename` separately from the config, then use in `imagespec` so that it can be used at later.

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+                else:
+                    imagespec = f"{pool}/{imagename}"
```
sunilangadi2, line 429: which `imagename` it is?

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+                # created on the first pool
+                rc = add_image_to_group_and_verify(
+                    **{
+                        "group-spec": mirror_group_spec,
+                        "image-spec": imagespec,
+                        "client": client,
+                    }
+                )
+                if rc != 0:
+                    return rc
```
sunilangadi2, line 445: same here

ceph/rbd/workflows/rbd_mirror.py
```diff
@@ -105,6 +105,23 @@ def wait_for_status(rbd, cluster_name, **kw):
                 )
                 if kw.get("images_pattern") == num_image:
                     return 0
+        elif kw.get("groupspec", False):
+            if kw.get("state_pattern"):
+                out = value(
+                    "state",
+                    json.loads(
+                        rbd.mirror.group.status(
```
sunilangadi2, line 113: can we make format as `config param` so that we can cover `json`, `text` as well?

ceph/rbd/workflows/rbd_mirror.py
```diff
@@ -105,6 +105,23 @@ def wait_for_status(rbd, cluster_name, **kw):
...
+                            **{"groupspec": kw.get("groupspec"), "format": "json"}
+                        )[0]
+                    ),
+                )
+                log.info(
+                    "State of {} group in {} cluster: {}".format(
+                        kw.get("groupspec"), cluster_name, out
+                    )
+                )
+                if kw.get("state_pattern") in out:
```
sunilangadi2, line 123: can we add proper state_pattern validation like `up+stooped` etc?

ceph/rbd/workflows/rbd_mirror.py
```diff
@@ -492,68 +509,70 @@ def config_mirror_multi_pool(
...
-                        "pool": pool,
-                        "image": image,
-                        "mirrormode": mirrormode,
-                        "io_total": image_config_val.get("io_total", None),
+            if pool_config.get("mirror_level", "") not in ["group", "namespace"]:
+                if (
+                    not kw.get("do_not_enable_mirror_on_image")
+                    and pool_config.get("mode") == "image"
+                ) or (
+                    pool_config.get("mode") == "pool"
```
sunilangadi2, line 517: snapshot based mirroring is applicable only for `image` mode not `pool` mode

ceph/rbd/workflows/rbd_mirror.py
```diff
@@ -492,68 +509,70 @@ def config_mirror_multi_pool(
...
-                                log.error(
-                                    f"Adding snapshot scheduling failed for image {pool}/{image}"
+                            if (
+                                out
+                                and "cannot enable mirroring: pool is not in image mirror mode"
+                                in out.strip()
+                            ):
+                                output = "Snapshot based mirroring cannot be enabled in pool mode"
+                            elif not is_secondary:
+                                output = (
```
sunilangadi2, line 551: are we covering this negative scenario here?

ceph/rbd/workflows/rbd_mirror.py
```diff
@@ -492,68 +509,70 @@ def config_mirror_multi_pool(
...
+                                image_config_val["snap_schedule_intervals"],
+                            ):
+                                snap_schedule_config = {
+                                    "pool": pool,
+                                    "image": image,
+                                    "level": level,
+                                    "interval": interval,
+                                }
+
+                                out, err = add_snapshot_scheduling(
```
sunilangadi2, line 569: are we covering all level of snapshot scheduling here?

cli/rbd/group.py
sunilangadi2, line 61: It should be `group-name`

cli/rbd/mirror/group.py
sunilangadi2, line 102: it should be `status`

sunilangadi2 (changes requested):

cli/rbd/group.py
chaitanyadommeti: Group is not an optional value here. [] is used for optional values only. Here Namespace is optional but not group.

ceph/rbd/workflows/rbd_mirror.py
chaitanyadommeti: No. Because we are not handling group mirroring enable here.
For namespace level, we are yet to think of the design when we start with that automation

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+    Args:
+        rbd (module): The rbd object
+        client (object): client object.
+        pool_type (str): The type of pool to create the mirror group for.
+        kw (dict): A dictionary of keyword arguments.
+
+    Returns:
+        int: 0 if the mirror group was created successfully, otherwise a non-zero value.
+    """
+    size = kw["config"][pool_type]["size"]
```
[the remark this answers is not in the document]
chaitanyadommeti: Changing will cause regression and confusion because the 'size' key is the existing one which I am using here. We might not want to change it

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+    rbd_config = kw.get("config", {}).get(pool_type, {})
+    multi_pool_config = getdict(rbd_config)
+    group_created = None
+    namespace = None
+    for pool, pool_config in multi_pool_config.items():
+        group = "gp_" + pool.split("_")[-1]
```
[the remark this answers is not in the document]
chaitanyadommeti: i started with the more readable format but it is printing long names which are not readable. Also we cant identify pool to which the group belongs.  Hence I shortened it as 'gp' but associated with the pools random number, so that we could identify the pool on which the group is created, so that we can keep track of the groups even of there are multiples

ceph/rbd/workflows/group.py
chaitanyadommeti: similar comment as above. Also when we use groupspecs, image specs, the length of the spec is even outgrowing the command length, thereby reducing the readability of the actions going on.

ceph/rbd/initial_config.py
```diff
@@ -97,6 +98,11 @@ def update_config(**kw):
             is_pool_config_updated = [
                 k for k in getdict(rep_pool_config).keys() if k not in ["test_config"]
             ]
+            if config.get("grouptype").startswith("single"):
+                rep_pool_config.update({"num_pools": 1})
+            elif config.get("grouptype").startswith("multi"):
+                if not config.get("num_pools") or rep_pool_config.get("num_pools") == 1:
```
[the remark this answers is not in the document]
chaitanyadommeti: we are overriding the value here to minimise the failures. if the user intends to use multipool option but mistakenly does not change number of pools, we are actually correcting it through module.

ceph/rbd/initial_config.py
```diff
@@ -212,6 +218,12 @@ def update_config(**kw):
             is_pool_config_updated = [
                 k for k in getdict(ec_pool_config).keys() if k not in ["test_config"]
             ]
+            if config.get("grouptype").startswith("single"):
+                ec_pool_config.update({"num_pools": 1})
+            elif config.get("grouptype").startswith("multi"):
+                if not config.get("num_pools") or ec_pool_config.get("num_pools") == 1:
```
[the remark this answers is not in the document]
chaitanyadommeti: we are overriding the value here to minimise the failures. if the user intends to use multipool option but mistakenly does not change number of pools, we are actually correcting it through module.

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
     # Group info
     (group_snap_out, group_snap_err) = rbd.group.snap.info(**group_snap_info_kw)
     return (group_snap_out, group_snap_err)
+
+
+def create_mirror_group(rbd, client, pool_type, **kw):
+    """
+    Creates a single group for mirroring.
```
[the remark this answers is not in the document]
chaitanyadommeti: Already used singular form of 'group' instead of plural form.

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+        int: 0 if the mirror group was created successfully, otherwise a non-zero value.
+    """
+    size = kw["config"][pool_type]["size"]
+    rbd_config = kw.get("config", {}).get(pool_type, {})
+    multi_pool_config = getdict(rbd_config)
+    group_created = None
+    namespace = None
+    for pool, pool_config in multi_pool_config.items():
+        group = "gp_" + pool.split("_")[-1]
+        group_create_kw = {"client": client, "pool": pool, "group": group}
```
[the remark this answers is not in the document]
chaitanyadommeti: Addressed

ceph/rbd/workflows/group.py
chaitanyadommeti: Addressed

ceph/rbd/workflows/group.py
```diff
@@ -336,3 +338,110 @@ def group_snap_info(**kw):
...
+        grouptype = kw.get("config").get("grouptype", "")
+        if grouptype in {"single_pool_with_namespace", "multi_pool_with_namespace"}:
```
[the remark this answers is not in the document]
chaitanyadommeti: Thanks for pointing this. Rather than failing its better 'single_pool_without_namespace' is put as default to help user

ceph/rbd/workflows/group.py
chaitanyadommeti: No we are not considering that.

ceph/rbd/initial_config.py
```diff
@@ -457,6 +469,8 @@ def initial_rbd_config(ceph_cluster, **kw):
         ):
             log.error(f"RBD configuration failed for {pool_type}")
             return None
+        if kw.get("config").get(pool_type).get("mirror_level") == "group":
```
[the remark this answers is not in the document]
chaitanyadommeti: Addressed

ceph/rbd/workflows/group.py
chaitanyadommeti: we are not considering default namespace here.  This line for secondary is executed only after  it gets executed for primary site. So we will never run into 'None' but neverthless we will make it fail, if it doesnt exist.
so will access the element directly instead of get()

ceph/rbd/workflows/group.py
chaitanyadommeti: Not a constraint but we are using that to avoid confusion and complexity both at the scripting level as well as debugging level.  After all, we are creating namespace here only for creating images and groups. so we are using that itself, instead of taking the next namespace and make it complex for the user to understand.

Consider the example of multiple pools. Even in this case we might want to create group thats associated with the first pool rather than some random pool and make the script logic and the debugging complex.

ceph/rbd/workflows/group.py
chaitanyadommeti: error message is taken care of by the function(create_namespace_and_verify) itself. it will be a duplicate logging if we add again.

ceph/rbd/workflows/group.py
chaitanyadommeti: Addressed

ceph/rbd/workflows/group.py
chaitanyadommeti: That is handled by the already existing logic in initital_config.py.  Image names are not created here. If the existing logic takes care of that, it should work here too.
Even if we dont provide from the config, we can still access the image name anywhere in the script and use it later.

ceph/rbd/workflows/group.py
chaitanyadommeti: the imagename that is created as in line #88 of initial_config.py

ceph/rbd/workflows/group.py
chaitanyadommeti: the error messages are taken care of by add_image_to_group_and_verify()

ceph/rbd/workflows/rbd_mirror.py
chaitanyadommeti: that would be good if we implement at all the levels in the module. 
Also for text, we need to implement pattern matching and hence needs an extra function. to deal with what is required.

we need to think of the design for covering the 'text' for all commands not this alone. lets discuss further.

ceph/rbd/workflows/rbd_mirror.py
chaitanyadommeti: we are doing it. we can always call 'wait for state' as in 
wait_for_status(
       rbd=rbd_primary,
       cluster_name=primary_cluster.name,
       groupspec=pool_config.get("group-spec", ""),
       state_pattern="up+stopped",
     )

ceph/rbd/workflows/rbd_mirror.py
chaitanyadommeti: this is all part of existing code. not done any modifications here

cli/rbd/group.py
chaitanyadommeti: Addressed

cli/rbd/mirror/group.py
chaitanyadommeti: Done

chaitanyadommeti: Done

ceph/rbd/workflows/group.py
chaitanyadommeti: Thanks for pointing this. Agreed this is confusing. Thought of changing  the variable name later but missed it.
Addressed.

ceph/rbd/workflows/group.py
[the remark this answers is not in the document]
chaitanyadommeti: Addressed as discussed

ceph/rbd/workflows/group.py
chaitanyadommeti: Addressed

aarsharm (approved):

sunilangadi2 (approved):
changes LGTM.

sunilkumarn417 (approved):

Manohar-Murthy (approved):
lgtm.

PR #4581, opened 2025-03-19
NVMeoF regression suites simplification

sunilkumarn417:
- Adding all NVMe test suites under simplification and compaction process.
- Removing install-prereq and deployment from current suites.
- Identify the test suite flow doesn't be bottleneck for next suite
- need to ensure proper cleanup.
- update test case configs w.r.t new cluster config


sunilkumarn417:
The PR was from wrong branch, so closing it now. opening new PR for the simplification.

PR #4584, opened 2025-03-20
RHCEPHQE-18630 | [RADOS] Ensure OSD removal method is time bound

harshkumarRH:
Jira tracker: [RHCEPHQE-18630](https://issues.redhat.com/browse/RHCEPHQE-18630)

Recently CephFS found and reported that current implementation of OSD removal triggers an infinite loop with no time boundary to drain OSD daemons and it only exits when OSD drain and removal has been successfull

This is a concerning test workflow as we already have bugs where cluster has been reported as stuck during backfill and recovery for more than 24 hours.
Refer: [BZ-2265371](https://bugzilla.redhat.com/show_bug.cgi?id=2265371)

With the code changes in the PR, the following are being addressed:
- Enhance the rm method implementation in `ceph_admin/osd.py` to not use While True loop but rather use a time bound while loop
- Allow users to control the retry interval instead of hard coding it to 2 secs

Jira tracker: [RHCEPHQE-18524](https://issues.redhat.com/browse/RHCEPHQE-18524)
The existing method to check daemon status was now being utilized to ascertain if the drain and removal of an OSD daemon was complete, it was expected that when an OSD daemon is no longer present in the cluster, the output of `ceph orch ps --daemon-type osd --daemon-id <osd-ID>` would return an empty list. Consequently, the method was still trying to fetch the contents from an empty list. Fix has been provided to address this false failure and ensure that method attempts to extract data only from a non-empty list.

### DIFF
--- a/ceph/ceph_admin/osd.py
+++ b/ceph/ceph_admin/osd.py
@@ -1,5 +1,6 @@
 """Manage OSD service via cephadm CLI."""
 
+import datetime
 import json
 from time import sleep
 from typing import Dict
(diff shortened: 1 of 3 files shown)


pdhiran (approved):
Looks good

neha-gangadhar:
@vamahaja Can you please review?

psathyan (approved):

PR #4614, opened 2025-03-26
[LibFix] Add support for multiple ssh key type for vm authentication

vamahaja:
Currently on RSA key support is available for VM authentication. Add new approach to support for authenticating using `RSA` & `ed25519Key`.

### DIFF
--- a/ceph/ceph.py
+++ b/ceph/ceph.py
@@ -10,6 +10,7 @@ import socket
 from distutils.version import LooseVersion
 from time import sleep, time
 
+import cryptography
 import paramiko
 import requests
 import yaml
@@ -1282,11 +1283,7 @@ class SSHConnectionManager(object):
         self.username = username
         self.password = password
         self.look_for_keys = look_for_keys
-        self.pkey = (
-            paramiko.RSAKey.from_private_key_file(private_key_file_path)
-            if look_for_keys
-            else None
-        )
+        self.pkey = self._get_ssh_key(private_key_file_path) if look_for_keys else None
         self.__client = paramiko.SSHClient()
         self.__client.set_missing_host_key_policy(paramiko.MissingHostKeyPolicy())
         self.__transport = None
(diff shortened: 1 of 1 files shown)


ceph/ceph.py
```diff
@@ -1297,6 +1293,19 @@ def __init__(
     def client(self):
         return self.get_client()
 
+    def _get_ssh_key(self, private_key_file_path):
+        """Get SSH key based on file type"""
+        with open(private_key_file_path, "r") as f:
+            _header = f.readline().strip()
+
+        if "BEGIN OPENSSH PRIVATE KEY" in _header:
```
psathyan, line 1301: An `ecdsa` key also can begin with this line.

ceph/ceph.py
```diff
@@ -1297,6 +1293,19 @@ def __init__(
...
+            return paramiko.Ed25519Key.from_private_key_file(private_key_file_path)
+        elif "BEGIN RSA PRIVATE KEY" in _header:
+            return paramiko.RSAKey.from_private_key_file(private_key_file_path)
+
+        logger.error(f"Unsupported ssh key {private_key_file_path}")
```
psathyan, line 1306: Please avoid using `fstrings` with log statement.

psathyan (changes requested):

ceph/ceph.py
vamahaja: @psathyan ecdsa private key starts with `BEGIN EC PRIVATE KEY`. Refer -
1. https://crypto.stackexchange.com/questions/48707/ecdsa-private-key-format
2. https://techdocs.akamai.com/iot-token-access-control/docs/generate-ecdsa-keys

ceph/ceph.py
vamahaja: done

ceph/ceph.py
psathyan: I believe that is when we use `openssl` however, we generate using `ssh-keygen`

compute/ibm_vpc.py
```diff
@@ -37,6 +37,7 @@ def get_ibm_service(access_key: str, service_url: str):
 
     service = VpcV1(authenticator=authenticator)
     service.set_service_url(service_url=service_url)
+    service.set_http_config({"timeout": 130})  # Increase to 130s
```
psathyan, line 40: Please create a separate commit or a new PR for this change. It helps backtrack review and keeps the history

vamahaja: Opened separate PR - https://github.com/red-hat-storage/cephci/pull/4670

ceph/ceph.py
vamahaja: Can u plz suggest alternative steps to differentiate keys ?

psathyan (approved):

PR #4662, opened 2025-04-09
Rados test suite optimizations - Update bench duration->max_obj

pdhiran:
Rados test suite optimizations - Update bench duration->max_obj & max read time to 10 sec.

This will help reduce the run time by :
1. Writing less objects via bench will complete faster.
2. the rebalancing time in the cluster will be fast due to less objects and will drastically reduce time for clean PGs on cluster. This should be fine as the intent is functional testing.


neha-gangadhar (approved):
Approved

psathyan (approved):

PR #4663, opened 2025-04-09
Add libraries for nvmeof namespace auto load balancing feature

rahullepakshi:
Library changes to add NS auto load balancing functinality to existing suites if rhcs_vrsion is 8.1

Design-
- Fetch the namespace count on each Gateway (GW) and compare them.
- Ensure that the number of namespaces for each GW is within the range [num_namespaces_per_gw + or - len(GWs)]
- Raises exception if a gateway's num-namespaces is out of the allowed range


HaruChebrolu (approved):
LGTM

PR #4685, opened 2025-04-14
Automate namespace auto load balancing tests w.r.t nvmeof GW scale up/down and replacement 

rahullepakshi:
Automate namespace auto load balancing tests w.r.t nvmeof GW scale up/down and replacement 

CEPH-83608837      Test NS load balancing by GW node replacement
CEPH-83609607	Test NS load balancing with incremental scale up and scale down of GWs in a GW group
CEPH-83609608	Test NS load balancing with one-time scale up and scale down in a GW group

### DIFF
--- a/suites/squid/nvmeof/tier-2_8-1_auto-load-balancing_tests.yaml
+++ b/suites/squid/nvmeof/tier-2_8-1_auto-load-balancing_tests.yaml
@@ -1,0 +1,282 @@
+# test suite for auto load balancing feature
+# cluster configuration file: conf/squid/nvmeof/ceph_nvmeof_4-nvmeof-gwgroup_2gw_cluster.yaml
+# inventory: conf/inventory/rhel-9.3-server-x86_64-xlarge.yaml
+
+tests:
+# Set up the cluster
+  - test:
+      abort-on-fail: true
+      module: install_prereq.py
+      name: install ceph pre-requisites
+  - test:
+      abort-on-fail: true
+      config:
+        verify_cluster_health: true
+        steps:
+          - config:
+              command: bootstrap
+              service: cephadm
+              args:
+                mon-ip: node1
+                registry-url: registry.redhat.io
+                allow-fqdn-hostname: true
+                log-to-file: true
+          - config:
+              command: add_hosts
+              service: host
+              args:
+                attach_ip_address: true
+                labels: apply-all-labels
+          - config:
+              command: apply
+              service: mgr
+              args:
+                placement:
+                  label: mgr
+          - config:
+              command: apply
+              service: mon
+              args:
+                placement:
+                  label: mon
+          - config:
+              command: apply
+              service: osd
+              args:
+                all-available-devices: true
+      desc: RHCS cluster deployment using cephadm
+      destroy-cluster: false
+      module: test_cephadm.py
+      name: deploy cluster
+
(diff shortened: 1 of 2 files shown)


suites/squid/nvmeof/tier-2_8-1_auto-load-balancing_tests.yaml
```diff
@@ -0,0 +1,282 @@
+# test suite for auto load balancing feature
+# cluster configuration file: conf/squid/nvmeof/ceph_nvmeof_4-nvmeof-gwgroup_2gw_cluster.yaml
+# inventory: conf/inventory/rhel-9.3-server-x86_64-xlarge.yaml
+
+tests:
```
HaruChebrolu, line 5: Do we need a new suite file for 8.1? Can't we use the existing suite for loadbalancing?

sunilkumarn417 (commented):
Hey @rahullepakshi 
I would suggest to move out the scale defintions to new module and call it in the test modules later.
And There is one scenario to explore here, when subsystems has unbalanced listeners and how auto load-balance takes place for the subsystem.

suites/squid/nvmeof/tier-2_8-1_auto-load-balancing_tests.yaml
rahullepakshi: @HaruChebrolu Initially I went with same approach but had to write new because of below reasons

-  this is a new feature where tests steps especially on validation vary from 8.0 and hence new polarion IDs and hence new suite because of new feature
-  Auto load balancing is best to verify with more number of namespaces
-  There are more scenarios like namespace deletion and namespaces addition to same LB group and deletion from same LB group that will be added to this suite  which are applicable to only 8.1 and irrelevant to 8.0

tests/nvmeof/workflows/ha.py
```diff
@@ -542,29 +542,30 @@ def scale_down(self, gateway_nodes_to_be_scaleddown):
...
-                    active = self.get_optimized_state(gateway.ana_group_id)
-
-                    # Find optimized path
-                    if active:
-                        end_counter, end_time = get_current_timestamp()
-                        LOG.info(
-                            f"{list(active[0])} is new and only Active GW for failed {hostname}"
-                        )
-                        break
+            if self.cluster.rhcs_version == "8":
```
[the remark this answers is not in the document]
rahullepakshi: Yes @sunilkumarn417 it worked and hence the logs. We skip this in 8.1 as DELETING state will be quickly disappear  and GW gets deleted so no point in checking for DELETING ana_grp ID

rahullepakshi:
> Hey @rahullepakshi I would suggest to move out the scale defintions to new module and call it in the test modules later. And There is one scenario to explore here, when subsystems has unbalanced listeners and how auto load-balance takes place for the subsystem.
@sunilkumarn417 Sure Sunil. I will take this by thursday/ Friday

"subsystems has unbalanced listeners" - when subsystems does not have listeners of all GWs in GW group. Thats a BZ actually, when not all listeners are added, we still load balance to GWs which does not have listeners. But documentation makes sure this does not happen but I will open up a bug to have restriction in code

HaruChebrolu (approved):
LGTM 
As we are going to cover other scenarios in another PR.

PR #4697, opened 2025-04-17
Skip Orch service updation if 'placement' key is missing

harshkumarRH:
Commit: [Skip Orch service updation if 'placement' key is missing](https://github.com/red-hat-storage/cephci/commit/b7fe4ab97e360b49138435783981443dc4bc183b)
Jira tracker: [RHCEPHQE-19052](https://issues.redhat.com/browse/RHCEPHQE-19052)
Recently it was found that the Orch service created for OSDs deployed manually using `ceph orch daemon add osd` method did not have any placement entry in their service spec file.

Due to the unavailability of `placement` key, modifying the 'unmanaged' status for such services is not feasible. Addition of custom `placement` key also does not help the cause as `placement` key can either take label or host/host pattern and what is required is individual daemon location. Consequently, it was decided to skip modification of such Orch services present on the cluster.

Test modules modified:
- `set_service_managed_type` in `ceph/rados/core_workflows.py`

Commit: 
Jira tracker: [RHCEPHQE-19103](https://issues.redhat.com/browse/RHCEPHQE-19103)
The changes ensure that cluster health detail and cluster status is logged immediately after a test fails or exception is encountered. It was found that the cluster health detail and cluster status logged in finally block was being recorded after considerable delay and sometimes after several other operations.

### DIFF
--- a/ceph/rados/core_workflows.py
+++ b/ceph/rados/core_workflows.py
@@ -4910,6 +4910,12 @@ EOF"""
         cmd_export = f"ceph orch ls {service_type} --export"
         out = self.run_ceph_command(cmd=cmd_export, client_exec=True)
         for _service in out:
+            if not _service.get("placement", False):
+                log.info(
+                    "Service %s does not have placement entry, skipping"
+                    % _service["service_name"]
+                )
+                continue
             if unmanaged:
                 log.debug(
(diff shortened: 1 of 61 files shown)


pdhiran (approved):
looks good

PR #4699, opened 2025-04-17
Fix: Handle non-JSON output from 'ceph orch upgrade status' when no upgrade is in progress

obannak:
# Description
RCA:

Previously, the "### ceph orch upgrade status" command would return a JSON output whether an upgrade was ongoing or had completed. However, this behaviour has changed. When no upgrade is running on the cluster, or if an ongoing upgrade has completed, the command now returns a string output: "There are no upgrades in progress currently."

FIX:

- Added error handling for invalid JSON response from 'ceph orch upgrade status'
- Used try-except block to gracefully handle JSON parsing failure
- Added fallback to check for specific text message in the output
- Raised ValueError when output doesn't match expected format or text message"

Log:
- http://10.70.46.85:8080/job/qe-squid-upgrade-openstack/2/

### DIFF
--- a/ceph/ceph_admin/upgrade.py
+++ b/ceph/ceph_admin/upgrade.py
@@ -1,10 +1,11 @@
 """Module that interfaces with ceph orch upgrade CLI."""
 
 from datetime import datetime, timedelta
-from json import loads
+from json import JSONDecodeError, loads
 from time import sleep
 from typing import Dict
 
+from ceph.ceph import CommandFailed
 from utility.log import Log
 
 from .common import config_dict_to_string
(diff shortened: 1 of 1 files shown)


ceph/ceph_admin/upgrade.py
```diff
@@ -65,7 +66,19 @@ def upgrade_status(self: OrchProtocol):
...
         """
         out, _ = self.shell(args=["ceph", "orch", "upgrade", "status"])
-        return loads(out)
+        try:
+            return loads(out)
+        except JSONDecodeError:
+            pass  # Not valid JSON, move to next check
+
+        if "There are no upgrades in progress currently." in out:
+            return out
```
vamahaja, line 75: Based on this result, [1] needs to be changed.
[1] https://github.com/red-hat-storage/cephci/pull/4699/files#diff-1b007be31b8ae465e57cdc902a3875d9d343cd2f7dba22fce8c222ccd361c534R112

harshkumarRH, line 75: Old output from 6.1:
```
[root@ceph-harsh-61-l7tedr-node7 ~]# ceph orch upgrade status 
{
    "target_image": null,
    "in_progress": false,
    "which": "<unknown>",
    "services_complete": [],
    "progress": null,
    "message": "",
    "is_paused": false
}
```

Updated output from 8.1:
```
There are no upgrades in progress currently.
```

IMO, to keep the changes to minimum, we should return a custom dictionary to maintain the behaviour of the method `upgrade_status`

The method `upgrade_status` is being called in `monitor_upgrade_status` that determines the status of upgrade by evaluating the value of key `"in_progress"` from the o/p of `upgrade_status`.

Instead of returning the raw output here, which could be a string, we need to return this -
```
return {"in_progress": False}
```

vamahaja (approved):

PR #4728, opened 2025-04-28
RHCEPHQE-19343: Apply workaround for Orch module failure during upgrades

harshkumarRH:
Jira tracker: [RHCEPHQE-19343](https://issues.redhat.com/browse/RHCEPHQE-19343)

Bug trackers:
https://bugzilla.redhat.com/show_bug.cgi?id=2314146
https://bugzilla.redhat.com/show_bug.cgi?id=2313471
https://bugzilla.redhat.com/show_bug.cgi?id=2361844

> Sometimes ceph-mgr modules are temporarily unavailable and their commands fail
> 
> Occasionally, the balancer module takes a long time to load after a ceph-mgr restart. As a result, other ceph-mgr modules can become temporarily unavailable and their commands fail.
> 
> For example,
> 
> [ceph: root@host01 /]# ceph crash ls
> Error ENOTSUP: Warning: due to ceph-mgr restart, some PG states may not be up to date
> Module 'crash' is not enabled/loaded (required by command 'crash ls'): use `ceph mgr module enable crash` to enable it
> 
> As a workaround, in cases after a ceph-mgr restart, commands from certain ceph-mgr modules fail, check the status of the balancer, using the ceph balancer status command. This might occur, for example, during an upgrade.
> 
> If the balancer was previously active `"active": true` but now it is marked as `“active": false`, check its status until it is active again, then rerun the other ceph-mgr module commands.
> In other cases, try to turn off the balancer ceph-mgr module.
> 
> ceph balancer off
> 
> After turning off the balancer, rerun the other ceph-mgr module commands.

The current code changes helps us achieve the following:
- If the test execution is going to upgrade the cluster to Squid (8.0/8.1), then ceph balancer module is turned off before triggering the upgrade.
- Post successful or unsuccessful upgrade, the ceph balancer module is restored to its original state (ON)

This should act as a workaround for the below error until the fix for the bug arrives in Squid, and should consequently unblock the upgrade CI pipeline -
```
Error ENOTSUP: Warning: due to ceph-mgr restart, some PG states may not be up to date
Module 'orchestrator' is not enabled/loaded (required by command 'orch upgrade status'): use ceph mgr module enable orchestrator to enable it
```

Two test cases were not modified as it was difficult to ascertain where they require any changes w.r.t the issue at hand:
- `test_rbd_snapshot_mirroring_with_upgrade.py`
- `test_cephadm_staggered_upgrade.py`

Requesting respective component owners to take care of the above two test cases.

### DIFF
--- a/tests/cephadm/test_cephadm_upgrade.py
+++ b/tests/cephadm/test_cephadm_upgrade.py
@@ -40,6 +40,7 @@ def run(ceph_cluster, **kwargs) -> int:
     """
     log.info("Upgrade Ceph cluster...")
     config = kwargs["config"]
+    rhbuild = config.get("rhbuild")
     config["overrides"] = kwargs.get("test_data", {}).get("custom-config")
     verify_image = bool(config.get("verify_cluster_health", False))
     orch = Orch(cluster=ceph_cluster, **config)
@@ -53,6 +54,24 @@ def run(ceph_cluster, **kwargs) -> int:
         executor = RadosBench(mon_node=client, clients=clients)
 
     try:
+        if rhbuild.startswith("8"):
+            log.info("Build passed for upgrade: %s" % rhbuild)
+            log_txt = """
+                Disabling the balancer module as a WA for bug : https://bugzilla.redhat.com/show_bug.cgi?id=2314146
+                Issue : If any mgr module based operation is performed right after mgr failover,
+                The command execution fails as the module isn't loaded by mgr daemon.
+                Issue was identified to be with Balancer module.
+                Disabling automatic balancing on the cluster as a WA until we get the fix for the same.
+                Disabling balancer should unblock Upgrade tests.
+                Error snippet :
+        Error ENOTSUP: Warning: due to ceph-mgr restart, some PG states may not be up to date
+        Module 'crash' is not enabled/loaded (required by command 'crash ls'): use `ceph mgr module enable crash`
+         to enable
+                """
+            log.info(log_txt)
+            out, err = client.exec_command(cmd="ceph balancer off", sudo=True)
+            log.debug(out + err)
+
         # Initiate thread pool to run rados bench
         if executor:
             executor.run(config=config["benchmark"])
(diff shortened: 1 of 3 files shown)


pdhiran (commented):
Do not enable balancer module post upgrade

pdhiran (changes requested):
Balancer module being enabled can/would still affect other tests where a mgr module based command is executed post a mgr failover, and not just the Upgrade test.

The Balancer module stays disabled post bootstrap as well, and I would suggest we keep same behaviour post upgrade as well, i.e to have balancer module disabled both in brownfield & Greenfield scenarios.

tests/cephadm/test_cephadm_upgrade.py
```diff
@@ -53,6 +54,24 @@ def run(ceph_cluster, **kwargs) -> int:
         executor = RadosBench(mon_node=client, clients=clients)
 
     try:
+        if rhbuild.startswith("8"):
```
sunilkumarn417, line 57: I would suggest to keep this outside the test modules, so it would be easy to use or remove anytime. and please make it re-usable, call it like a class with all necessary logging.
```
class BZ2314146:

      def balancer_off(self):
      ...
      def ...
      ...
```

harshkumarRH: Hi @sunilkumarn417 
The current code changes have been done keeping in mind that this workaround would only be around for next few weeks.
Had this been a permanent code change or a significant one, it would make sense to put in the effort to modularize the code.

Even we make it re-usable, we would still need to raise another PR to remove it from the test case modules when the fix arrives.

Please let me know if you'd still like a re-usable class to be created, I will remove the modified files other than rados upgrade test module from this PR and respective owners take a call on how they would like to include the workaround in their test case modules.

sunilangadi2 (approved):

sunilangadi2 (commented):
@harshkumarRH please add upgrade pass logs after this workaround applied.

harshkumarRH:
Closing the PR as disabling the `ceph balancer` module is not enough and no longer a viable workaround to get past this issue.
We have raised a needInfo on developer and awaiting further feedback.
https://bugzilla.redhat.com/show_bug.cgi?id=2314146#c82

PR #4740, opened 2025-04-30
Moving ceph log collection out of skip sos report collection check

tintumathew10:
# Description

Ceph cluster logs were not getting captured by default for failed test suites as skip-sos-report collection is defaulted to true. Moving the get_ceph_var_logs function out side of the skip-sos-report check.


tintumathew10:
This is handled in pipeline jobs. so closing this ticket.

PR #4741, opened 2025-04-30
[Workaround] Add steps to retry upgrade status command

vamahaja:
Sometimes ceph-mgr modules are temporarily unavailable and their commands fail. As a result, mgr commands fail including upgrade status.

To Fix issue add retry to check command status.

### DIFF
--- a/ceph/ceph_admin/upgrade.py
+++ b/ceph/ceph_admin/upgrade.py
@@ -6,6 +6,7 @@ from time import sleep
 from typing import Dict
 
 from ceph.ceph import CommandFailed
+from ceph.waiter import WaitUntil
 from utility.log import Log
 
 from .common import config_dict_to_string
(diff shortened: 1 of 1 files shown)


obannak (commented):
LGTM

ceph/ceph_admin/upgrade.py
```diff
@@ -57,15 +58,36 @@ def start_upgrade(self: OrchProtocol, config: Dict):
...
+    def upgrade_status(self: OrchProtocol, timeout: int = 600, interval: int = 10):
         """
         Execute the command ceph orch status.
 
         Returns:
             upgrade Status (Dict)
 
         """
-        out, _ = self.shell(args=["ceph", "orch", "upgrade", "status"])
+        # Retry upgrade command for interval time until it pass till timeout
```
sunilkumarn417, line 69: My suggestion is to keep this outside `upgrade.py`, maybe in any workflow and not to affect any libraries.

vamahaja: @sunilkumarn417 this is just retry and should not affect any upgrades. Kindly check latest upgrade pipeline results.

tintumathew10 (approved):
Latest upgrade pipeline results are good after picking these changes.

PR #4753, opened 2025-05-06
[LibFix] Add mon bootstrap configs to interop tests

vamahaja:
RCA:
  Steps to add mon to specific node is missing from bootstrap configs and it's causing to add mon on all available nodes. Ceph FS test `cephfs-basics` fails to validate IPs as it's expects node configs present in global-conf.


tintumathew10 (commented):
LGTM

sunilkumarn417 (approved):

PR #4781, opened 2025-05-12
UI selenium framework initial structure

pavankumarag:
This is a skeleton PR for UI automation, which will be subsequently used by rest of the FG members.

### DIFF
--- a/ui/common/click.py
+++ b/ui/common/click.py
@@ -1,0 +1,49 @@
+"""Helper for Clicking WebElement"""
+
+import time
+
+from selenium.common.exceptions import NoSuchElementException, TimeoutException
+from selenium.webdriver.common.action_chains import ActionChains
+from selenium.webdriver.support import expected_conditions as ec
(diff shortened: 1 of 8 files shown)


psathyan:
I strongly believe this should be in a separate repo and should not included here. Not all test suites require `selenium` dependencies.

ui/common/click.py
```diff
@@ -0,0 +1,49 @@
+"""Helper for Clicking WebElement"""
+
+import time
+
+from selenium.common.exceptions import NoSuchElementException, TimeoutException
+from selenium.webdriver.common.action_chains import ActionChains
+from selenium.webdriver.support import expected_conditions as ec
+from selenium.webdriver.support.ui import WebDriverWait
```
psathyan, line 8: `selenium` package would need to be part of the `requirements.txt`

psathyan (changes requested):

PR #4783, opened 2025-05-13
TFA [RADOS] RHCEPHQE-19671: Add smart wait to let Acting set return to original state

harshkumarRH:
Jira tracker: [RHCEPHQE-19671](https://issues.redhat.com/browse/RHCEPHQE-19671)
From CI pipeline runs it was observed that the test to setup osd failure due to allocator file corruption would fail every other run due to a change in Acting set after OSDs were removed and added back after modification of their device class.

The acting set change was given a very small blind sleep of 5 secs to return to its original state, this wait time was obviously not enough for network bandwidth constraints environment of rhos-01.

A smart wait of 180 secs has been added to let the Acting set recover, it the acting set remains in a changed state, the test is eventually failed.

Test modules modified:
- `tests/rados/test_osd_crashes.py`

Logs:
Squid 8.1 - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-0GC43A/

### DIFF
--- a/tests/rados/test_osd_crashes.py
+++ b/tests/rados/test_osd_crashes.py
@@ -4,6 +4,7 @@ Tests included:
 1. Bug 2335317 - multiple OSDs crashing while replaying bluefs -- ceph_assert(delta.offset == fnode.allocated)
 """
 
+import datetime
 import time
 
 from ceph.ceph_admin import CephAdmin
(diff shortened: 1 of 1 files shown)


pdhiran (approved):
looks good

PR #4789, opened 2025-05-13
Increased timeout value

chaitanyadommeti:
Test fails if running time is more than 1200 sec(20min). Sometimes due to network lags, the sript "cli_migration.sh" takes more than 20 min and the test is considered failed.
Increased the timeout value to 1800

http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-F8KL7S/


Manohar-Murthy (approved):
Approving the temporary fix.

PR #4801, opened 2025-05-15
[RADOS] TFA fixes for concurrent-parallel ops | OSD start false failure | Slow ops request test

harshkumarRH:
Jira trackers:
- [RHCEPHQE-18559](https://issues.redhat.com/browse/RHCEPHQE-18559)
- [RHCEPHQE-17018](https://issues.redhat.com/browse/RHCEPHQE-17018)
- [RHCEPHQE-19350](https://issues.redhat.com/browse/RHCEPHQE-19350)

TFA fixes provided:
- Concurrent and parallel ops
The test used to fail when run in conjuncture with test omaps. Although the test execution has stabilized over the previous two pipeline runs, it was decided the move this test to another test suite - location-stretch-mode
For the time being, the test is present in two test suites:
- `tier-3_rados_test-location-stretch-mode.yaml`
- `tier-2_rados_test_omap.yaml`

Depending upon the result of next few pipeline runs, a decision will be made to find the final single place for the test.

- Slow Ops request
The test was initially designed in such a way to overload the VM cluster and organically generate the "Slow Ops request" warning. However, this was done keeping in mind the then popular VM instance type - medium. Gradually we moved away from medium VM instance to a higher one - large or x-large. The workflow designed for medium instance was not enough to bring the cluster to a state where the warning would get generated due to OSD flapping. Even after increasing the load on the cluster, it was found that the environment in rhos-d and rhos-01 were too different to design a scenario were OSDs would flap organically and consistently in each run.
Finally, it was decided to no longer chase the organic generation this warning and reduce the value of `osd_op_complaint_time` parameter instead to capture "Slow ops request" warning easily.

- Smart wait to let OSD be reported as UP
At times it was seen that a recently added OSD is up and running but needs a small amount of time to reported as UP in the `ceph osd tree up` output. Changes have been made to wait for the state of OSD to get updated.

Logs:
Reef - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-C440FI
Squid - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-WZ1Z06

### DIFF
--- a/tests/rados/test_slow_op_requests.py
+++ b/tests/rados/test_slow_op_requests.py
@@ -1,10 +1,12 @@
+import datetime
+import re
+import time
 import traceback
 
 from ceph.ceph_admin import CephAdmin
 from ceph.rados.core_workflows import RadosOrchestrator
-from tests.rados.rados_test_util import create_pools, get_slow_requests_log
+from tests.rados.monitor_configurations import MonConfigMethods
 from utility.log import Log
-from utility.utils import method_should_succeed, should_not_be_empty
 
 log = Log(__name__)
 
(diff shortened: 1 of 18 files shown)


neha-gangadhar:
@harshkumarRH Please rebase as Vipin's PR is merged and a conflict needs to be resolved.

tests/rados/test_slow_op_requests.py
```diff
@@ -1,67 +1,131 @@
...
+
+        init_logs = rados_obj.get_journalctl_log(
+            start_time=start_time,
+            end_time=end_time,
+            daemon_type="osd",
+            daemon_id=max_osd.split(".")[-1],
+        )
+        slow_ops_log_list = []
+        for line in init_logs.splitlines():
+            if "slow ops" in line:
```
pdhiran, line 98: I think the test should fail even if we do not find any occurrence of slow_ops on OSD logs. This is because we have already validated that the selected OSD had recorded most of the slow_ops

pdhiran (changes requested):
1 small modification required.

pdhiran (approved):
Looks good

PR #4822, opened 2025-05-22
[TFA][upgrade] Add waiter for health check in staggered upgrade tests

vinpapnoi:
# Description

### Problem
The staggered upgrade CI runs are failing with health check post OSD upgrade.

### Reason
Suspicion is that the health check is being performed right after the upgrade and only once.

### Solution
Add waiter for the health check.


pranavprakash20 (approved):
LGTM

PR #4840, opened 2025-05-27
fix endtime for scale operations

HaruChebrolu:
# Description

Fix endtime for scaleup and scaledown operations during NS autoloadbalancing


Manohar-Murthy (approved):
lgtm

PR #4853, opened 2025-05-29
[RADOS] Create IBM specific test suites for Upgrade pipeline

harshkumarRH:
Recently it was observed that the below test suite would fail in the IBM upgrade pipeline as their initial deployment was pointing to RHCS versions and builds.

- `tier-2_rados_ec-pool_recovery.yaml`
- `tier-2_rados_test-brownfield.yaml`
- `tier-2_rados_test-stretch-mode-upgrade.yaml`

This PR adds equivalent test suites that map to similar IBM builds during bootstrap.

### DIFF
--- a/suites/reef/rados/tier-2_rados_ec-pool_recovery_ibm.yaml
+++ b/suites/reef/rados/tier-2_rados_ec-pool_recovery_ibm.yaml
@@ -1,0 +1,283 @@
+# Suite contains tier-2 rados test: EC Pool recovery
+# CLuster should have atleast 6 OSD hosts for testing.
+# Suited best for BM pipeline.
+# Can be run on RHOS-d env as well.
+
+tests:
+  - test:
+      name: setup install pre-requisistes
+      desc: Setup phase to deploy the required pre-requisites for running the tests.
+      module: install_prereq.py
+      abort-on-fail: true
+
+  - test:
+      name: cluster deployment
+      desc: Execute the cluster deployment workflow.
+      module: test_cephadm.py
+      polarion-id:
+      config:
+        verify_cluster_health: true
+        steps:
+          - config:
+              command: bootstrap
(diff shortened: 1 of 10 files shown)


suites/reef/rados/tier-2_rados_test-brownfield_ibm.yaml
```diff
@@ -0,0 +1,319 @@
...
+      desc: Upgrade cluster to 7.1z2 version
+      module: test_upgrade_warn.py
+      polarion-id: CEPH-83574934
+      config:
+        command: start
+        service: upgrade
+        base_cmd_args:
+          verbose: true
+        args:
+          rhcs-version: 7.1
```
neha-gangadhar, line 183: Here also we need changes to upgrade?

suites/reef/rados/tier-2_rados_ec-pool_recovery_ibm.yaml
```diff
@@ -0,0 +1,267 @@
+# Suite contains tier-2 rados test: EC Pool recovery
```
pdhiran, line 1: should we have a 5.x ->  7.x IBM upgrade path?

pdhiran: all paths currently added are only from n-1 & n GA releases

suites/reef/rados/tier-2_rados_test-brownfield_ibm.yaml
harshkumarRH: Added support for `custom_image` and `custom_repo` in `test_upgrade_warn.py`

pdhiran:
looks good. please add pass log for new upgrade path added.

neha-gangadhar (approved):
Approved

PR #4437, opened 2025-02-11
Automation of the Bug#2330755 - Scrub/deep scrub taking too long under mclock I/O scheduler

SrinivasaBharath:
# Description

Automation of the - Bug#2330755 - Scrub/deep scrub taking too long under mclock I/O scheduler
The verification steps are at - https://bugzilla.redhat.com/show_bug.cgi?id=2330755#c12
A new test case is added to the polarion- https://polarion.engineering.redhat.com/polarion/#/project/CEPH/workitem?id=CEPH-83605026

Please include Automation development guidelines. Source of Test case - New Feature/Regression Test/Close loop of customer BZs
<details>

<summary>click to expand checklist</summary>

- [ ] Create a test case in Polarion reviewed and approved.
- [ ] Create a design/automation approach doc. Optional for tests with similar tests already automated.
- [ ] Review the automation design
- [ ] Implement the test script and perform test runs
- [ ] Submit PR for code review and approve
- [ ] Update Polarion Test with Automation script details and update automation fields
- [ ] If automation is part of Close loop, update BZ flag qe-test_coverage “+” and link Polarion test
</details>

PR #4856, opened 2025-05-30
[RADOS] RHCEPHQE-19872: Include additional scenarios in 'Scrubbing with mClock' tests

harshkumarRH:
Follow up PR for #4437 

Adds the following:
- Check to determine time taken to complete deep-scrub with default/unaltered values of `osd_mclock_max_capacity_iops_hdd`
- Uses relative scrub time instead of blind wait to ensure scrubbing is not completed after reducing 
`osd_mclock_max_capacity_iops_hdd`
- Checks if deep-scrub time after restoring `osd_mclock_max_capacity_iops_hdd` is close to original value

Test cases modified:
- `tests/rados/test_scrub_deepscrub_timecheck.py`

Logs:
Squid -
Reef -

### DIFF
--- a/tests/rados/test_scrub_deepscrub_timecheck.py
+++ b/tests/rados/test_scrub_deepscrub_timecheck.py
@@ -18,6 +18,10 @@ log = Log(__name__)
 
 def run(ceph_cluster, **kw):
     """
+    # CEPH-83605026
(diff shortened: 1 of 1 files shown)


pdhiran (approved):
Looks good

PR #4860, opened 2025-06-01
Added rgw workflow

SaiSuriya30:
The link : http://magna002.ceph.redhat.com/ceph-qe-logs/Suriya/rest_api/testswithsleep/
This module is to automate the creation of  the bucket by checking the user id and list them out in the rgw component. 
for the fiur endpoints it is been automated .

The following steps are followed in the test:

1. User id is checked 
2. If User id exists then the bucket is created 
3. Else the user id is created then the bucket creation takes place 
4. The buckets are listed


SaiSuriya30:
commit issue

PR #4878, opened 2025-06-06
Addition of RADOS suite & conf files in tentacle folder

pdhiran:
Addition of conf and suite files for RADOS -> Tentacle release.

All test suites present in squid have been moved to tentacle, with updates in config as needed for Tentacle.

Upgrade suites :
tier-2_rados_ec-pool_recovery.yaml -> 7.1 RC -> 9.0
tier-2_rados_test-brownfield.yaml -> 7.1z1 -> 7.1z2 -> 8.0z5 -> 9.0
tier-2_rados_test-drain-customer-issue.yaml -> 7.1z0 -> 9.0
tier-2_rados_test-pg-dups-trimming.yaml -> 8.0 rc -> 9.0
tier-2_rados_test-stretch-mode-upgrade.yaml -> 7.1 RC -> 8.1 RC -> 9.0
tier-3_rados_test-3-AZ-Cluster.yaml -> 8.0 RC -> 9.0
tier-3_rados_test-4-node-8+6-msr-ecpools.yaml -> 8.0RC -> 9.0
tier-3_rados_test-4-node-ecpools.yaml -> 8.0 rc -> 9.0

### DIFF
--- a/suites/tentacle/rados/tier-2_rados_bm_add_db_wal_cbt.yaml
+++ b/suites/tentacle/rados/tier-2_rados_bm_add_db_wal_cbt.yaml
@@ -1,0 +1,167 @@
+# Suite contains  tier-2 rados tests which adds DB and wal to the OSD
+#===============================================================================================
+#------------------------------------------------------------------------------------------
+#-----      Tier-2- Using bluestore-tool add DB & Wal  to the existing OSD          ------
+#------------------------------------------------------------------------------------------
+# Conf: conf/tentacle/baremetal/mero_4_node_4_client_conf.yaml
+# Requirements:  Minimum one SSD and NVME is required to verify this testcase.In Mero nodes
+#                initially configuring the OSD's on HDD devices.On test OSD adding DB using SSD
+#                and adding WAL using NVME device.
+#===============================================================================================
+tests:
+  -
+    test:
+      abort-on-fail: true
+      desc: "Setup phase to deploy the required pre-requisites for running the tests."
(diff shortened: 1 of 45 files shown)


suites/tentacle/rados/tier-2_rados_bm_add_db_wal_cbt.yaml
harshkumarRH, line 1: The conf file for this suite does not exist now, has not been added with this PR and may not even be valid anymore. We can review this latest.

suites/tentacle/rados/tier-2_rados_test-brownfield.yaml
```diff
@@ -0,0 +1,334 @@
...
+      abort-on-fail: true
+
+  - test:
+      name: "verify fix-osd recovery with allocator file corruption"
+      module: test_osd_crashes.py
+      desc: failed osd recovery due to allocator file corruption during upgrade
+      polarion-id: CEPH-83609785
+      config:
+        verify_allocator_corruption:
+          verify_fix: true
```
harshkumarRH, line 234: This test would need to run after 8.x upgrade because 8.x would already contain the fix.
I do still think that we should skip the upgrade to 8.x altogether in this test suite, reason being - if the fix stops working in 9.0 suddenly, then we would never no because the failed OSD has already been recovered after upgrade to 8.x
By the time the cluster is about to upgrade to 9.0, it would already be in HEALTH_OK state.

PR #4886, opened 2025-06-09
[Fix] Update service verify timer

adityaramteke:
We're observing services taking longer expected to up and running in, CI enviromnet. Updating timeout to 300 to fix the issue.


vdas-redhat (approved):
LGTM

PR #4896, opened 2025-06-11
[RADOS] TFA fixes for CBT WAL/DB Addition | Mitigate OSD systemctl false failure | Fix bootstrap arguments for test suites

harshkumarRH:
PR addresses the following:

1. Changes the OSD deployment to method `raw` instead of `lvm` in order to be able to add dedicated WAL and DB using `ceph-bluestore-tool`
2. Mitigation for false failure of OSD systemctl operations with rhel 9.6
3. Fix incorrect positioning of arguments being provided to bootstrap in `rados/tier-2_rados_ec-pool_recovery.yaml` test suite
4. Removal of recently that IBM test suites that are now redundant.


pdhiran (approved):
looks good

PR #4901, opened 2025-06-11
Adding missed coverage in reef for 4 node pools.

pdhiran:
We have balanced tests between 2+2 & 8+6 MSR pools, as both are configured on 4 nodes. However, since there is no 8+6 MSR pool in reef, same test suite cannot be maintained as squid.

- Updated the missing coverage in reef


neha-gangadhar (approved):
Approved

PR #4907, opened 2025-06-12
Fix setting pg_num_min on pools for EC 4 node pools

pdhiran:
issue with pg_num_min being set before PG count increases. Fixing the issue, and also dynamically chooing value for pg_num_min for pools

### DIFF
--- a/tests/rados/pool_tests.py
+++ b/tests/rados/pool_tests.py
@@ -9,6 +9,7 @@ Tests included:
 """
 
 import datetime
+import math
 import random
 import re
 import time
@@ -514,6 +515,11 @@ def run(ceph_cluster, **kw):
         log.debug("Verifying pg_num_min on pools")
         target_configs = config["verify_pg_num_min"]["configurations"]
         try:
+            log.debug(
+                "Setting the default autoscaler mode to on, making sure that the Pool created will have "
+                "PG Autoscaler enabled on it."
+            )
+            rados_obj.configure_pg_autoscaler(default_mode="on")
             # Creating pools and starting the test
             for entry in target_configs.values():
                 log.debug(f"Creating {entry['pool_type']} pool on the cluster")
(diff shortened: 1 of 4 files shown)


neha-gangadhar (commented):
lgtm

tests/rados/pool_tests.py
```diff
@@ -514,6 +515,11 @@ def run(ceph_cluster, **kw):
         log.debug("Verifying pg_num_min on pools")
         target_configs = config["verify_pg_num_min"]["configurations"]
         try:
+            log.debug(
+                "Setting the default autoscaler mode to on, making sure that the Pool created will have "
+                "PG Autoscaler enabled on it."
+            )
+            rados_obj.configure_pg_autoscaler(default_mode="on")
```
harshkumarRH, line 522: Why is this something that we need to do? The default value is anyway ON, right?

tests/rados/pool_tests.py
```diff
@@ -526,23 +532,41 @@ def run(ceph_cluster, **kw):
...
+                    if pg_num_on_pool > 1:
+                        pg_num_min = math.ceil(pg_num_on_pool / 2)
+                        break
+                    time.sleep(5)
+
+                try:
+                    if not rados_obj.set_pool_property(
                         pool=entry["pool_name"],
                         props="pg_num_min",
                         value=entry["pg_num_min"],
```
pdhiran, line 555: Update the pg_num_min value here. use param created in line 536

tests/rados/pool_tests.py
pdhiran: Yes, the default is on. This is to avoid false errors/ clean-up misses from previous tests. 

No harm or foul done, as it's the default behaviour anyway.

tests/rados/pool_tests.py
pdhiran: done

PR #4916, opened 2025-06-16
fix endtime for scaledown and scaleup operations

HaruChebrolu:
# Description

Fix endtime for scaledown and scaleup operations in NS loadbalancing suites


tintumathew10 (approved):
lgtm

PR #4939, opened 2025-06-19
Automation for CEPH-83620649

aarsharm:
Pass logs: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-VAA1RP
Polarion Link: https://polarion.engineering.redhat.com/polarion/#/project/CEPH/workitem?id=CEPH-83620649

[CEPH-83620649 - Group info should verify snapshot completion before marking the group as primary


suites/squid/rbd/tier-2_rbd_group_mirror.yaml
```diff
@@ -456,3 +456,42 @@ tests:
...
+        secondary during force promote
+      module: test_group_info_while_force_promote.py
+      polarion-id: CEPH-83620649
+      clusters:
+        ceph-rbd1:
+          config:
+            rep_pool_config:
+              rbd_mirror_group: true
+              num_pools: 1
+              num_images: 1
```
sunilangadi2, line 476: For Group scenario a minimum number of images at least two should be good. 
also are we covering adding images to a group from different pool?

suites/squid/rbd/tier-2_rbd_group_mirror.yaml
```diff
@@ -456,3 +456,42 @@ tests:
...
+
+  - test:
+      abort-on-fail: true
+      desc: >
+        Verify group info and group snapshot status on
+        secondary during force promote
+      name: >
+        Test group info and group snapshot status on
+        secondary during force promote
+      module: test_group_info_while_force_promote.py
```
chaitanyadommeti, line 468: How about moving this and few other tests may be with negative scenarios into tier-3 category to run less frequent as in weekly? Lets discuss

tests/rbd_mirror/test_group_info_while_force_promote.py
```diff
@@ -0,0 +1,326 @@
...
+    if snap_state != verify_state:
+        raise Exception(
+            "Snap state should not turn "
+            + verify_state
+            + " before force promote completion"
+        )
+    log.info("Successfully verified snapshot state to be " + verify_state)
+
+
+def check_group_info(expected_primary_field, rbd_secondary, **group_config):
```
sunilangadi2, line 256: let's add this in `group workflow` folder so that it can be reused for other tests

tests/rbd_mirror/test_group_info_while_force_promote.py
```diff
@@ -0,0 +1,326 @@
...
+    (out, err) = rbd_secondary.mirror.group.promote(
+        **{"group-spec": group_spec, "force": True}
+    )
+    if err:
+        raise Exception("Failed to force promote group on site-B: " + str(err))
+    else:
+        log.info("Successfully completed force promote on secondary")
+
+
+def check_group_snap_list(snapshot_id, verify_state, rbd_secondary, **status_spec):
```
sunilangadi2, line 245: let's add this in group workflow folder so that it can be reused for other tests

tests/rbd_mirror/test_group_info_while_force_promote.py
```diff
@@ -0,0 +1,326 @@
...
+                snapshot_id, "complete", rbd_secondary=rbd_secondary, **status_spec
+            )
+
+            # Check group info, 'primary' field should be marked true
+            check_group_info(
+                expected_primary_field=True, rbd_secondary=rbd_secondary, **group_config
+            )
+
+
+def force_promote_secondary(group_spec, rbd_secondary):
```
sunilangadi2, line 235: let's add this in group workflow folder so that it can be reused for other tests
also make it generic as `group_force_promote` based on the site passed it should work

sunilangadi2 (changes requested):

tests/rbd_mirror/test_group_info_while_force_promote.py
```diff
@@ -0,0 +1,326 @@
...
+                "pool": pool,
+                "group": pool_config.get("group"),
+                "format": "json",
+            }
+            if "namespace" in pool_config:
+                status_spec.update({"namespace": pool_config.get("namespace")})
+            else:
+                group_spec = pool + "/" + pool_config.get("group")
+
+            wait_till_image_sync_percent(rbd_primary, 50, **group_config)
```
chaitanyadommeti, line 196: Can we put a logger to notify that we are waiting for 50% sync to complete?

tests/rbd_mirror/test_group_info_while_force_promote.py
chaitanyadommeti: same as above

tests/rbd_mirror/test_group_info_while_force_promote.py
sunilangadi2: i think we can call workflow functions also in parallel by importing

chaitanyadommeti: its not about calling functions, its about validating the result

suites/squid/rbd/tier-2_rbd_group_mirror.yaml
aarsharm: from different pool its not supported.
Also as the image size here is big almost 20-30G IO, hence i have kept one to contain runtime

suites/squid/rbd/tier-2_rbd_group_mirror.yaml
aarsharm: As per offline discussion, we will move this TC to separate yaml later as part of reshuffling or removing other TC also to weekly which is of relatively less priority

sunilangadi2 (dismissed):
LGTM.

chaitanyadommeti (dismissed):
LGTM

PR #4945, opened 2025-06-19
Adding fix for IO progression issue in NVMe load balancing test suites

manasagowri:
IO was getting initiated for only 16 out of the 20 namespaces across 2 subsystems in the load balancing test suite.
So, I added a new parameter max_workers to the Parallel class, which will then be used to instantiate ThreadPoolExecutor class, so that it doesn't take the default max_workers value and instead takes the value specified during initialisation.

I then set the max_workers value to max_workers=len(paths) + 4 where paths are the devices. So that we can instantiate enough number of threads to initiate IOs for all devices at once.

With this change IO progression is working as expected (refer logs below). However I uncovered another issue that after adding listeners to the scaled up node, few of the devices are missing in the initiator. Need to analyse this issue further as a separate new issue. So I will not be combining the analysis and fix of this issue as part of this PR.

Logs - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-98TQ97/

Logs for 4 subsystems - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-U4OIU6/

### DIFF
--- a/ceph/parallel.py
+++ b/ceph/parallel.py
@@ -36,6 +36,11 @@ In version 2, you can choose whether to use threads or processes by
     with parallel(thread_pool=False, timeout=10) as p:
         _r = [p.spawn(quux, x) for name in names]
 
+You can also specify the maximum number of worker threads/processes:
+    with parallel(max_workers=20) as p:
+        for foo in bar:
+            p.spawn(quux, foo, baz=True)
+
 If one of the spawned functions throws an exception, it will be thrown
 when iterating over the results, or when the with block ends.
 
@@ -59,6 +64,7 @@ class parallel:
         thread_pool=True,
         timeout=None,
         shutdown_cancel_pending=False,
+        max_workers=None,
     ):
         """Object initialization method.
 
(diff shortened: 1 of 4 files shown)


chebroluharika:
@manasagowri Changes looks good to me. Did you try running for 4 subsystems with 20ns each also?

chebroluharika:
> @HaruChebrolu I tried 4 subsystems with 10 namespaces each. Do you want me to try for 20 namespaces each? Any particular reason for it?

We're encountering this issue when dealing with a larger number of subsystems, each having more namespaces. I had previously tested with 10 namespaces per subsystem using a similar `max_workers` adjustment, which worked fine locally but failed when run through the pipeline. That's why I asked @rahullepakshi to reduce the number of namespaces.
Reference: [https://github.com/red-hat-storage/cephci/pull/4845/files](https://github.com/red-hat-storage/cephci/pull/4845/files). However, we want suites with 20namespaces each for subsystems.

manasagowri:
@HaruChebrolu As discussed logs for 2 subsystems with 20 namespaces each - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-8Y90Z7/Gw_node_replacement_0.log

HaruChebrolu:
> @HaruChebrolu As discussed logs for 2 subsystems with 20 namespaces each - http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-8Y90Z7/Gw_node_replacement_0.log

Looks good. Please change all test configs to have 2 subsystems with 20ns each.

rahullepakshi (dismissed):
LGTM post https://github.com/red-hat-storage/cephci/pull/4945#issuecomment-2995122298

HaruChebrolu (approved):
LGTM

PR #4951, opened 2025-06-20
[DNM][LibFix] Skip IBMC node cleanup based on prefix

vamahaja:
IBM Cloud does not provide an option to lock instances against accidental deletion via scripts. This caused issues when instance names shared a common prefix.

To prevent such scenarios, skip instance selection if the node name starts with the dnd prefix.

### DIFF
--- a/ceph/utils.py
+++ b/ceph/utils.py
@@ -50,7 +50,11 @@ def cleanup_ibmc_ceph_nodes(ibm_cred, pattern):
         log.warning("Failed to retrieve instances")
         return 1
 
-    instances = [i for i in resp.get_result()["instances"] if pattern in i["name"]]
+    instances = [
+        i
+        for i in resp.get_result()["instances"]
+        if pattern in i["name"] and not i["name"].startswith("dnd")
+    ]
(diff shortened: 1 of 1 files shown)


obannak (approved):

tintumathew10 (approved):

tintumathew10 (commented):
approve

ceph/utils.py
```diff
@@ -50,7 +50,11 @@ def cleanup_ibmc_ceph_nodes(ibm_cred, pattern):
         log.warning("Failed to retrieve instances")
         return 1
 
-    instances = [i for i in resp.get_result()["instances"] if pattern in i["name"]]
+    instances = [
+        i
+        for i in resp.get_result()["instances"]
+        if pattern in i["name"] and not i["name"].startswith("dnd")
```
psathyan, line 56: We shouldn't be doing this check. The infra or critical systems must be moved to a separate project which CI has no control. 

Adding such checks without documentation is a maintenance nightmare.

psathyan (changes requested):

PR #4963, opened 2025-06-24
[DNM][LibFix] Add support to use local rhel repos

vamahaja:
Add parameter `use-local-rhel-repos` to use local RHEL repositories. 

Usage - 
```
python run.py \
	--cloud openstack \
	--instances-name <prefix> \
	--rhbuild 8.1 \
	--platform rhel-9 \
	--global-conf conf/squid/smoke/1admin-4node-1client-bvt.yaml \
	--suite suites/squid/smoke/bvt.yaml \
	--osp-cred <osp-cred-path> \
	--inventory conf/inventory/rhel-9.6-server-x86_64.yaml \
	--rhs-ceph-repo <repo-path> \
	--docker-registry registry-proxy.engineering.redhat.com \
	--docker-image rh-osbs/rhceph \
	--docker-tag <tag> \
	--use-local-rhel-repos
```
Add below config in `~/.cephci.yaml` -
```
repo:
  local:
    rhel-9:
      - <appstream-repo-path>
      - <baseos-repo-path>

    rhel-8:
      - <appstream-repo-path>
      - <baseos-repo-path>
```


cli/utilities/repos.py
```diff
@@ -0,0 +1,32 @@
...
+from cli.utilities.packages import Package
+from cli.utilities.utils import os_major_version
+from utility.log import Log
+from utility.utils import get_cephci_config
+
+log = Log(__name__)
+
+
+def setup_local_repos(ceph):
+    # Get configuration details from `~/.cephci.yaml`
```
psathyan, line 11: Reading from `cephci.yaml` is going to put us in the same spot as `registry`. IMHO, we should avoid it.

This is required for the base image and it can come from the inventory file as we control what is getting installed.

psathyan: In case of `baremetal`, this would definitely be useful but it should be an input passed via `custom-config` to `install_prereq`

run.py
```diff
@@ -65,6 +65,7 @@
         [--osp-cred <file>]
         [--rhs-ceph-repo <repo>]
         [--ubuntu-repo <repo>]
+        [--use-local-rhel-repos]
```
psathyan, line 68: IMHO, this is specific to `install_prereq` module and should be passed to it. I don't see it's usage across the framework.

psathyan (changes requested):

run.py
vamahaja: `run.py` already have similar options (e.g `--skip-subscription`, `--add-repo` etc), adding this as custom-config option unnecessarily add complexity and no any config passed through test to install-prereq.

tintumathew10:
changes are looking good.

PR #4979, opened 2025-06-26
Changing number of ns per subsystem to suit available OSD space

manasagowri:
As discussed, added 20 namespaces of 5gb each for each test.


HaruChebrolu (approved):
LGTM

PR #4985, opened 2025-06-30
Added changes to support Inband at scale

HaruChebrolu:
# Description

Added changes to support Inband at scale.


rahullepakshi (approved):
This might not be a scale test as we will have 128 subsystems and 2K NS(not sure if enabling inband for 128 makes sense) but these are good tests for cross feature perspective

PR #5004, opened 2025-07-03
[TFA][mix-os] Remove unsupported IO

vinpapnoi:
# Description

With mix-OS deployment, the IO fails due to older python version (3.6) in RHEL 8.x which is not supported.
These IOs are running and passing in other suites and are not necessary here, hence removing them from these suites.


MohitBis (approved):
LGTM

pranavprakash20 (approved):
Lgtm

PR #5019, opened 2025-07-10
[DMFG][ceph-vol][smoke] Add suite and conf files for tentacle

vinpapnoi:
# Description

Add Tentacle release suite and conf files for ceph-volume and smoke components


vdas-redhat (approved):
LGTM

PR #5063, opened 2025-07-24
[Upstream] Added client package issue workaround

sunilangadi2:
# Description
The upstream currently has an open issue ([#71250](https://tracker.ceph.com/issues/71250)) related to missing OpenSSL dependencies. As a temporary workaround, this PR manually installs the required openssl package to unblock the upstream pipeline runs until a proper fix is merged upstream.

Test result:
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-LZ6BUK/

<details>

</details>

### DIFF
--- a/ceph/utils.py
+++ b/ceph/utils.py
@@ -1315,3 +1315,10 @@ def find_vm_node_by_hostname(ceph_cluster, target_hostname):
         ),
         None,
     )
+
+
+def is_client(ceph_node):
+    """
+    Return True if client node else False
+    """
+    return len(ceph_node.role.role_list) == 1 and "client" in ceph_node.role.role_list
--- a/run.py
+++ b/run.py
@@ -164,6 +164,10 @@ test_names = []
 run_summary = {}
 
 
+class CephCIArgumentError(Exception):
+    pass
+
+
 @retry(LibcloudError, tries=5, delay=15)
 def create_nodes(
     conf,
@@ -461,8 +465,11 @@ def run(args):
             base_url, docker_registry, docker_image, docker_tag = fetch_build_artifacts(
                 build, rhbuild, platform, upstream_build, ibm_build
             )
-    else:
-        build = None
+    elif build == "upstream":
+        if not upstream_build:
+            raise CephCIArgumentError(
+                "--upstream-build argument not provided Ex: tentacle"
+            )
 
     store = args.get("--store") or False
 
@@ -775,6 +782,7 @@ def run(args):
                     config["add-repo"] = repo
 
             config["build_type"] = build
+            config["upstream_build"] = upstream_build
             config["enable_eus"] = enable_eus
             config["skip_enabling_rhel_rpms"] = skip_enabling_rhel_rpms
             config["docker-insecure-registry"] = docker_insecure_registry
--- a/tests/misc_env/install_prereq.py
+++ b/tests/misc_env/install_prereq.py
@@ -4,7 +4,7 @@ import re
 import time
 
 from ceph.parallel import parallel
-from ceph.utils import config_ntp, update_ca_cert
+from ceph.utils import config_ntp, is_client, update_ca_cert
 from ceph.waiter import WaitUntil
 from cli.exceptions import ConfigError
 from cli.utilities.packages import Package
@@ -100,6 +100,21 @@ def run(**kw):
             )
             time.sleep(20)
 
+    # TODO: remove this method once the openssl issue resolved in
+    # upstream tentacle build, https://tracker.ceph.com/issues/71250
+    if (
+        config.get("build_type") == "upstream"
+        and config.get("upstream_build") == "tentacle"
+    ):
+        with parallel() as p:
+            for ceph in ceph_nodes:
+                if not is_client(ceph):
+                    continue
+
+                p.spawn(
+                    workaround_openssl_issue,
+                    ceph,
+                )
     return 0
 
 
@@ -489,3 +504,29 @@ def update_iptables(node):
                 node.exec_command(cmd=f"$(which iptables) -D {rule}", sudo=True)
     except Exception as err:
         log.error(f"iptables rpm do not exist... error : {err}")
+
+
+def workaround_openssl_issue(node) -> None:
+    """
+    Workaround for the openssl issue in RHEL 9.0
+    This is a temporary fix to install openssl-libs package
+    for tentacle upstream build.
+
+    NOTE:
+        remove this method once the openssl issue resolved in
+        upstream tentacle build, https://tracker.ceph.com/issues/71250
+
+    Args:
+        node: client node object
+    """
+
+    openssl_url = (
+        "https://mirror.stream.centos.org/9-stream/BaseOS/"
+        "x86_64/os/Packages/openssl-libs-3.5.1-1.el9.x86_64.rpm"
+    )
+    node.exec_command(cmd=f"wget {openssl_url}", sudo=True, long_running=True)
+    node.exec_command(
+        cmd="rpm --force -i openssl-libs-3.5.1-1.el9.x86_64.rpm",
+        sudo=True,
+        long_running=True,
+    )


Manohar-Murthy (dismissed):
lgtm

tests/cephadm/test_client.py
```diff
@@ -173,8 +173,20 @@ def setup(host):
...
+                    # Workaround for upstream tentacle branch package issue
+                    openssl_pkg = "openssl-libs-3.5.1-1.el9.x86_64.rpm"
+                    openssl_url = (
+                        "https://www.rpmfind.net/linux/centos-stream/9-stream/BaseOS/"
+                        "x86_64/os/Packages/openssl-libs-3.5.1-1.el9.x86_64.rpm"
+                    )
+                    _node.exec_command(
+                        cmd=f"wget -O {openssl_pkg} {openssl_url}", sudo=True
+                    )
+                    _node.exec_command(cmd=f"rpm --force -i {openssl_pkg}", sudo=True)
```
psathyan, line 189: can we limit this to only upstream?

sunilangadi2: @psathyan I tried your suggested approach using `--custom-config upstream=true` and `--build-type`, but those options aren't currently available in the config. Modifying it would also require changes to the pipeline metadata. Since this is a temporary workaround that we'll revert once the upstream issue is resolved, I’ve added a condition that applies only to upstream builds for now. Please review it.

sunilangadi2:
Test results:
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-AVDXQL/

tests/cephadm/test_client.py
```diff
@@ -175,6 +175,9 @@ def setup(host):
 
             # Install ceph-common
             if config.get("install_packages"):
+                # Added this condition for upstream build
+                if config.get("ceph_docker_registry") == "quay.ceph.io":
```
psathyan, line 179: This is not ideal as it creates a dependency on a non-related parameter. This will run irrespective of `tentacle` version.

tests/cephadm/test_client.py
```diff
@@ -252,6 +255,24 @@ def remove(cls, config: Dict) -> None:
...
+    for tentacle upstream build.
+    """
+    openssl_url = (
+        "https://www.rpmfind.net/linux/centos-stream/9-stream/BaseOS/x86_64/os/Packages/"
+        "openssl-libs-3.5.1-1.el9.x86_64.rpm"
+    )
+    node.exec_command(cmd=f"wget {openssl_url}", sudo=True)
+    node.exec_command(
+        cmd="rpm --force -i openssl-libs-3.5.1-1.el9.x86_64.rpm", sudo=True
+    )
```
psathyan, line 273: Please evaluate this with `dnf install -yq --skip-broken --nobest openssl_url` 

We would be able to avoid the need to download the file in a separate operation.

psathyan (changes requested):

tests/cephadm/test_client.py
psathyan: @sunilangadi2 it is available as part of `config[test_data]` that is past to the test module. As discussed earlier, it is available in `kw[test_data]`

Yes, I agree it requires metadata changes. It is singular point of change and also introduces a way to customize. I understand what you trying is short term but we cannot develop or implement solutions that short lived.

tests/cephadm/test_client.py
sunilangadi2: It's getting failed as below
```
[root@ceph-sangadi-up-7pqws6-node1-installer ~]# dnf install -yq --skip-broken --nobest https://www.rpmfind.net/linux/centos-stream/9-stream/BaseOS/x86_64/os/Packages/openssl-libs-3.5.1-1.el9.x86_64.rpm on 10.0.66.204
  File "/usr/lib64/python3.9/random.py", line 61, in <module>
    from _sha512 import sha512 as _sha512
ModuleNotFoundError: No module named '_sha512'

During handling of the above exception, another exception occurred:

  File "/usr/bin/dnf", line 61, in <module>
    from dnf.cli import main
  File "/usr/lib/python3.9/site-packages/dnf/__init__.py", line 23, in <module>
    import dnf.pycomp
  File "/usr/lib/python3.9/site-packages/dnf/pycomp.py", line 24, in <module>
    import email.mime.text
  File "/usr/lib64/python3.9/email/mime/text.py", line 10, in <module>
    from email.mime.nonmultipart import MIMENonMultipart
  File "/usr/lib64/python3.9/email/mime/nonmultipart.py", line 10, in <module>
    from email.mime.base import MIMEBase
  File "/usr/lib64/python3.9/email/mime/base.py", line 9, in <module>
    import email.policy
  File "/usr/lib64/python3.9/email/policy.py", line 7, in <module>
    from email._policybase import Policy, Compat32, compat32, _extend_docstrings
  File "/usr/lib64/python3.9/email/_policybase.py", line 9, in <module>
    from email.utils import _has_surrogates
  File "/usr/lib64/python3.9/email/utils.py", line 28, in <module>
    import random
  File "/usr/lib64/python3.9/random.py", line 64, in <module>
    from hashlib import sha512 as _sha512
  File "/usr/lib64/python3.9/hashlib.py", line 77, in <module>
    import _hashlib
ImportError: libcrypto.so.3: cannot open shared object file: No such file or directory
```

tests/cephadm/test_client.py
sunilangadi2: Checked it's getting like below

```
-> if config.get("install_packages"):
(Pdb) config[test_data]
*** NameError: name 'test_data' is not defined
(Pdb) kw
*** NameError: name 'kw' is not defined
```
also here https://github.com/red-hat-storage/cephci/blob/main/tests/cephadm/test_client.py#L279
`kw` is not getting passed

psathyan: Please see the `run` method ... it is not passed to `add` method.

tests/cephadm/test_client.py
psathyan: Let's go with `wget` and `rpm` for now.

harshkumarRH: @sunilangadi2 Please use the centos mirror link for pulling the rpm

tests/cephadm/test_client.py
harshkumarRH: ^ Unless of course we make changes to pipeline to propagate the build info as an argument to run.py execution.

Actually, in retrospect, we should ideally be passing `--build upstream` in every upstream run regardless of whether `--upstream-build` is identified or repo and image details are provided separately. Without setting `--build upstream`, the method `def setup_upstream_repository` will never be invoked.

Maybe outside the scope of this PR, but this should also be addressed.

tests/cephadm/test_client.py
```diff
@@ -252,6 +258,21 @@ def remove(cls, config: Dict) -> None:
...
+    """
+    Workaround for the openssl issue in RHEL 9.0
+    This is a temporary fix to install openssl-libs package
+    for tentacle upstream build.
+    """
+    openssl_url = (
+        "https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/Packages/"
+        "openssl-libs-3.5.1-1.el9.x86_64.rpm"
+    )
+    node.exec_command(cmd=f"dnf install -y {openssl_url}", sudo=True)
```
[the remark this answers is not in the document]
harshkumarRH: @sunilangadi2 
I was suggesting that the `dnf install` command does not help, the centos mirror url should still be fine and a better source than rpmfind.net

tests/cephadm/test_client.py
psathyan: Without knowing the arguments passed to `run.py`, it is difficult to confirm or deny if it is broken.

The `build` is set to `None` when you have set `--build upstream` and also have passed the options `-- rhs-ceph-repo ... --docker-{registry,image,image-tag}` When these parameters are passed a logical override occurs. I don't have the backgroud / context on why it was done.

@sunilkumarn417 these changes were done a couple of years ago. Could you provide @sunilangadi2 the context?

`--build upstream` reads the recipe file for the details of repo and image. That is the purpose it should serve.

I am not sure about your inhibitions to use `--custom-config` despite mentioning to go down that path. It becomes difficult if you keep trying something or then make a statement to guide or review.

tests/cephadm/test_client.py
psathyan: > So the only way to do a conditional check for Upstream Tentacle would be to ensure rhbuild starts with 9 and --docker-registry is quay.ceph.io

I disagree with this statement.

We can introduce `--custom-config test_client_override_patch_openssl=true` and evaluate the same inside `test_client.py`. At the same time introduce this option to metadata in the pipelines of the specified upstream projects.

This way it is a single change point when the fix is introduced. Looking at `9` and `quay.ceph.io` in the long run is a bit difficult and does not provide any scope of extension

The primary purpose of `--custom-config` was to allow test configurations to be overriden outside of the test data.

tests/cephadm/test_client.py
harshkumarRH: @psathyan 
Are you suggesting that `--build upstream`'s only purpose is to direct the framework to read the recipe files?

I see that we rely on the `build_type` being set to `upstream` in order to invoke `def setup_upstream_repository` as well

Although, this might need to be addressed with another PR after more context is shared by @sunilkumarn417

tests/cephadm/test_client.py
harshkumarRH: > We can introduce --custom-config test_client_override_patch_openssl=true and evaluate the same inside test_client.py. At the same time introduce this option to metadata in the pipelines of the specified upstream projects. 
> This way it is a single change point when the fix is introduced. Looking at 9 and quay.ceph.io in the long run is a bit difficult and does not provide any scope of extension

Understood, I assumed that we are not open to making changes in the metadata/pipeline for this workaround, that's why the inefficient suggestion.
Using `--custom-config test_client_override_patch_openssl=true` sounds reasonable, thanks

tests/cephadm/test_client.py
harshkumarRH: @sunilangadi2
I propose the following code changes - https://github.com/harshkumarRH/cephci/commit/c4a1fc708d8d339118836f9d6b5a19013040c3a7

A total of 3 changes:
1. Using `--custom-config` with run.py to control invocation of openssl workaround
2. `test_data` dictionary to be appended to the `config` dictionary 
   - `--custom-config` is passed to the test modules with `test_data` dict, the same would need to be read in order to parse user provided configs.
3. Conditional checks to decide installation of openssl packages based on user input resides inside the `workaround_openssl_issue` method. Keeps the implementation logically isolated and may or may not require removal in the future.

custom-config that should be passed during Jenkins job execution under the `CUSTOM_CONFIG` build parameter:
`<any-text>patch_openssl<any-text>=True`

Please discuss with others accordingly

sunilangadi2: Thanks @harshkumarRH for sharing your solution: https://github.com/harshkumarRH/cephci/commit/c4a1fc708d8d339118836f9d6b5a19013040c3a7#r162930129
I'm good with the approach, but I’ve reiterated a few of my earlier comments — please take a look when you get a chance.

Requesting @psathyan and @sunilkumarn417 to review and help to get this wrapped up quickly.

> custom-config that should be passed during Jenkins job execution under the CUSTOM_CONFIG build parameter:
> <any-text>patch_openssl<any-text>=True

@vamahaja hope you ok with this approach, 
please suggest if you have any better solution.

tests/cephadm/test_client.py
sunilangadi2: thanks @harshkumarRH tested your solution worked fine
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-2HVH6X/

AmarnatReddy (approved):

vamahaja (commented):
LGTM

tintumathew10 (approved):
/lgtm

PR #5095, opened 2025-07-31
FIX: make build type argument non-required

sunilkumarn417:
As of the test framework, we need build to be always pointed to `released` if none provided for build agrument. 

- We need build arg as mandatory, but thats needs changes in pipelines and environment.
- need to revert this PR, once it is ready.

```
  --build <latest>                  Type of build to be use for testing
                                    eg: latest|tier-0|tier-1|tier-2|released|upstream
                                    [default: released]
```

### DIFF
--- a/run.py
+++ b/run.py
@@ -496,6 +496,11 @@ def run(args):
             raise CephCIArgumentError(
                 "--upstream-build argument not provided Ex: tentacle"
             )
+    else:
+        # TODO: By default, build=released, but Pipelines and environment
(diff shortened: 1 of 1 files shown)


sunilangadi2 (approved):
Thanks @sunilkumarn417 also please create jira task for the revert work.

tintumathew10 (commented):
lgtm

sunilkumarn417:
> LGTM. Thanks for reverting it sunil.

Actually, we need our env to be ready for `build=released`. please check description.

PR #5115, opened 2025-08-06
Adding support for using SDP profile in multi vpc environment

psathyan:
# Description

As part of this commit, we are enhancing the IBMC compute driver to support

- boot volume profile selection
- switching the profile depending on the VPC

SDP is software defined profile enabled for us in eu-de region.

- [x] Perform test runs


krishnaramaswamy (approved):
lgtm!

krishnaramaswamy (approved):
lgtm

PR #5175, opened 2025-08-27
[SMB][Automation]Add new scenario

MohitBis:
Scenarios: 
1. Upgrade with case-sensitive disable
2. Case-sensitive bug automation https://bugzilla.redhat.com/show_bug.cgi?id=2386474
3. Deployment and Upgrade (SMB and NFS services in same node)


adityaramteke (approved):
LGTM

PR #5194, opened 2025-09-04
Added inventory for latest rhel-8

obannak:
The suite tier1-mix-os-deployment requires admin node to be RHEL 9.x and rest of the nodes to be RHEL 8.x. While running this test in IBMC, the test fails with TestSetupFailure because of missing RHEL 8.x inventory in IBMC.


vinpapnoi (approved):
LGTM

PR #5213, opened 2025-09-10
Add Tentacle support to DMFG regression test suite

obannak:
Added tentacle support in the DMFG regression test suite.


vinpapnoi (approved):
LGTM

PR #5217, opened 2025-09-11
Reorder the swift at root test which is causing the keystone tests to…

TejasC88:
… fail

Swift at root parameter causes all S3 tests to fail, so I am moving this test to the end of the suite in reef only . We have bifurcated the suites in squid and further so that this issue is not seen.

Pass log : http://magna002.ceph.redhat.com/cephci-jenkins/results/openstack/RH/7.1/rhel-9.5/Test/18.2.1-350/338/tier-2_rgw_regression_test/


yuva-gayam (approved):

ckulal (approved):

mkasturi18 (approved):
lgtm

PR #5221, opened 2025-09-15
Add library and workflow support for nvme namespace reservation

rahullepakshi:
Add library and workflow support for nvme namespace reservation 

Operations delt -
Reservation lifecycle: register, acquire, report, release and preempt


ceph/nvmeof/initiators/linux/nvme_cli.py
```diff
@@ -151,3 +169,42 @@ def connect_all(self, **kwargs):
...
         )
+
+    def register_reservation(self, device, nsid, nrkey=None, rrega=0, cptpl=0):
+        """
+        Register reservation for a namespace on an initiator
+        """
+        cmd = f"nvme resv-register {device} --namespace-id {nsid} --nrkey {nrkey} --cptpl {cptpl} --rrega {rrega} -v"
+        return self.execute(cmd=cmd, sudo=True)
+
+    def acquire_reservation(self, device, nsid, rtype, crkey=None, racqa=0):
```
chaitanyadommeti, line 180: [optional arguments] same as above

ceph/nvmeof/initiators/linux/nvme_cli.py
```diff
@@ -151,3 +169,42 @@ def connect_all(self, **kwargs):
...
+        """
+        cmd = f"nvme resv-release {device} --namespace-id {nsid} --crkey {crkey} --rtype {rtype} --rrela {rrela} -v"
+        return self.execute(cmd=cmd, sudo=True)
+
+    def preempt_reservation(self, device, nsid, rtype, crkey=None, prkey=None, racqa=0):
+        """
+        Acquire reservation for a namespace on an initiator
+        """
+        cmd = (
+            f"nvme resv-acquire {device} --namespace-id {nsid} "
```
chaitanyadommeti, line 206: Since this is the same CLI as of acquire, this should be handled with a preempt flag in acquire_reservation() itself.

Also, if the user does not provide prkey here, this defaults to 'none'  which means we are not actually preempting the reservation key here.

-p <prkey>, --prkey=<prkey>

Preempt Reservation Key: If the Reservation Acquire Action is set to 001b (i.e., Preempt) or 010b (i.e., Preempt and Abort), then this field specifies the reservation key to be unregistered from the namespace. For all other Reservation Acquire Action values, this field is reserved.

ceph/nvmeof/initiators/linux/nvme_cli.py
```diff
@@ -89,7 +89,7 @@ def list_subsys(self, **kwargs):
             cmd=f"nvme list-subsys {device} {config_dict_to_string(kwargs)}", sudo=True
         )
 
-    def list_spdk_drives(self):
+    def list_spdk_drives(self, nsid_device_pair=None):
         """List the NVMe Targets only SPDK drives.
```
manasagowri, line 94: When and why does a user need to pass nsid_device_pair and what happens when its passed or left at None.

rahullepakshi: This has to be passed when we need device name and NSID whcih are mandatory parameters for reservation cmds 
if it is None this would be executed the usual way we get devices today, when passed we get device and NSID pair

chaitanyadommeti (approved):
LGTM

Manohar-Murthy (approved):
Approving based on the team review.

PR #5246, opened 2025-09-25
fixing false results in CI pipeline

Harishacharya-redhat:
THis PR is to fix false results in CI - for nfs
example:
in this run : http://magna002.ceph.redhat.com/cephci-jenkins/results/openstack/RH/9.0/rhel-9/Regression/20.1.0-12/nfs/18/tier1-nfs-ganesha-v4-2/ , few tests are returning PASS, even though its failed.

this is due to the design:
try:
except:
finally:

return 0 
if return 1/failed scenarios are not defined, then the test will return PASS, even though test failed.

the fix as of my knowledge:
try:
 return 0
except:
 
finally:

Since by default CI framework will return  return 0/ fail, So its better to define return 0/pass signal:

logs with changes:
failed results :
http://magna002.ceph.redhat.com/ceph-qe-logs/hacharya/false_results/try1/logs/.  - 9.0

pass results:
http://magna002.ceph.redhat.com/ceph-qe-logs/hacharya/false_results/try1/logs-pass-false-1/ - 8.1
http://magna002.ceph.redhat.com/ceph-qe-logs/hacharya/false_results/try1/false-results-pass2/ - 8.1

### DIFF
--- a/tests/nfs/nfs_client_permission_export.py
+++ b/tests/nfs/nfs_client_permission_export.py
@@ -84,7 +84,7 @@ def run(ceph_cluster, **kw):
             pass
         else:
             log.error(f"Mount passed on unauthorized client: {clients[0].hostname}")
-
+        return 0
     except Exception as e:
         log.error(f"Failed to perform export client addr validation : {e}")
         return 1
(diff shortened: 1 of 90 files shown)


msainiRedhat (approved):
LGTM

PR #5250, opened 2025-09-26
[Tentacle][SMB]Add QoS support

MohitBis:
[Tentacle][SMB]Add QoS support

Test Cases Added:
Verify SMB QoS via CLI command (Imperative Style)
Verify SMB QoS via CLI command with multiple shares (Imperative Style)
Verify SMB QoS using CLI with two shares - one configured with QoS and one without. (Imperative Style)
Verify SMB QoS via spec file (Declarative Style)
Verify SMB QoS via spec file with multiple shares (Declarative Style)
Verify SMB QoS using spec file with two shares - one configured with QoS and one without. (Declarative Style)
Verify set x qos limit on a share and verify throughput remains within the limit
Verify set X qos limits on multiple shares and verify per-share enforcement
Verify that the SMB QoS settings are updated as expected in the smb conf
Verify that the user is able to list the SMB QoS values.

### DIFF
--- a/cli/ceph/smb/share.py
+++ b/cli/ceph/smb/share.py
@@ -51,3 +51,16 @@ class Share(Cli):
         if isinstance(out, tuple):
             return out[0].strip()
         return out
+
+    def update_cephfs_qos(self, cluster_id, share_id, **kw):
+        """Update qos per share
+
+        Args:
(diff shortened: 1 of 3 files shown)


adityaramteke (approved):
LGTM

PR #5273, opened 2025-10-13
RHCEPHQE-18628: Enable coredump collection during daemon crashes

harshkumarRH:
# Description

Jira tracker: [RHCEPHQE-18628](https://issues.redhat.com/browse/RHCEPHQE-18628)

At times we have noticed that ceph daemons crash during test execution in CI pipeline where we auto delete the cluster upon test completion
As part of test tear-down in RADOS, we currently have checks to ensure that if any crash is reported during or after the test case execution, the test is reported as failure.

In most of the cases, developers have repeatedly asked to attach the coredump whenever a bug related to daemon crash is reported. Due to the unavailability of concerned cluster and low reproducibility of complex crashes, it gets difficult to furnish these additional details.

Steps to enable coredump collection for containerized ceph deployments - https://bugzilla.redhat.com/show_bug.cgi?id=2170213#c1

- By design, coredump will be enabled on all the nodes as part of `setup install pre-requisistes` but will be collected only during pipeline failure. 
- ~~New flag `--ceph-coredump` introduced in run.py to preserve coredumps regardless of test case failure status.~~
- Custom config option `collect-coredump` added to preserve coredump even if no failure occurs
- Email template modified to include coredump location

cmd:
`python3 run.py --rhbuild 8.1 --global-conf conf/squid/rados/7-node-cluster.yaml --osp-cred /home/hakumar/osp-cred-ceph-systest.yaml --inventory conf/inventory/rhel-9-latest.yaml --suite /home/hakumar/deploy-cluster.yaml --log-level info --platform rhel-9 --skip-sos-report --build latest --store --custom-config collect-coredump=True`

Logs -
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-LMNFE2/
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-E10I79/
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-682GT4/

### DIFF
--- a/run.py
+++ b/run.py
@@ -27,7 +27,7 @@ from ceph.utils import (
     create_ceph_nodes,
     create_ibmc_ceph_nodes,
 )
-from cephci.cluster_info import get_ceph_var_logs
+from cephci.cluster_info import collect_ceph_coredumps, get_ceph_var_logs
 from cephci.utils.build_info import CephTestManifest
 from cli.performance.memory_and_cpu_utils import (
     start_logging_processes,
@@ -487,6 +487,8 @@ def run(args):
 
     # FixMe: We should be using product for differentiation.
     ibm_build = False
+    # disable coredump collection by default
+    collect_coredump = False
 
     # Custom or override configurations
     kernel_repo = args.get("--kernel-repo")
@@ -563,6 +565,8 @@ def run(args):
     enable_eus = args.get("--enable-eus")
     skip_enabling_rhel_rpms = args.get("--skip-enabling-rhel-rpms")
     skip_sos_report = args.get("--skip-sos-report")
+    if "collect-coredump" in custom_config_dict.keys():
+        collect_coredump = bool(custom_config_dict["collect-coredump"])
 
     # load config, suite and inventory yaml files
     conf = load_file(glb_file)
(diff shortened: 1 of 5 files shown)


run.py
```diff
@@ -154,6 +155,8 @@
                                     rhel images for Interop runs
   --skip-sos-report                 Enables to collect sos-report on test suite failures
                                     [default: false]
+  --collect-coredump                Enables presevation of coredump directory
+                                    [default: false]
```
psathyan: IMHO, I would like to avoid providing such nobs... unless it is required by multiple modules, then it makes sense.

If it is required only in the `install_prereq` see if you could leverage `--custom-config install_prereq_enable_coredump`

tests/misc_env/install_prereq.py
```diff
@@ -115,6 +115,12 @@ def run(**kw):
                     workaround_openssl_issue,
                     ceph,
                 )
+
+    # enable coredump on all nodes
+    with parallel() as p:
+        for ceph in ceph_nodes:
+            p.spawn(enable_coredump, ceph)
+            time.sleep(20)
```
psathyan: Could we add it to `def install_prereq(...)` rather running through the thread again?

psathyan (commented):
Thanks @harshkumarRH  for adding this functionality 

some minor suggestions.

psathyan (commented):
/approve

PR #5313, opened 2025-10-27
[DMFG][upgrade] Fix ceph versions pre-upgrade

vinpapnoi:
# Description

Update rhcs version for pre-upgrade installations


obannak (approved):
LGTM

PR #5316, opened 2025-10-28
Added Support for host, daemon & service level upgrades

pdhiran:
Enhanced the upgrade workflow to accept host, daemon & service level upgrades.

```
# Upgrade Options Example usage :
#   args:
#     rhcs-version: 8.0
#     release: rc
#     daemon_types: "mon,mgr"  # Upgrade only mon and mgr daemons
#     hosts: "node1,node2"     # Upgrade daemons only on node1 and node2
#     services: "mon"          # Upgrade only the mon service
```
Note that all the options in args are optional.

### DIFF
--- a/ceph/ceph_admin/upgrade.py
+++ b/ceph/ceph_admin/upgrade.py
@@ -42,6 +42,9 @@ class UpgradeMixin:
                 args:
                     image: "latest"         # pick latest image from config
                     version: 16.0.0         # Not supported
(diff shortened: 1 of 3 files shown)


pdhiran:
Pass log for deployment + when both ceph version details + daemon/host/service details passed via args: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-LPZK5A

PR #4836, opened 2025-05-27
TFA fixes for regression runs

AmarnatReddy:
# Description
TFA fixes for regression runs
This has fixes for 
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-ULVI2Y/multifs_reboot_with_fstab_0.log
http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-ULVI2Y/Running_basic_bash_commands_on_fuse,Kernel_and_NFS_mounts_0.log

Please include Automation development guidelines. Source of Test case - New Feature/Regression Test/Close loop of customer BZs
<details>

<summary>click to expand checklist</summary>

- [ ] Create a test case in Polarion reviewed and approved.
- [ ] Create a design/automation approach doc. Optional for tests with similar tests already automated.
- [ ] Review the automation design
- [ ] Implement the test script and perform test runs
- [ ] Submit PR for code review and approve
- [ ] Update Polarion Test with Automation script details and update automation fields
- [ ] If automation is part of Close loop, update BZ flag qe-test_coverage “+” and link Polarion test
</details>

PR #5327, opened 2025-10-30
[RADOS] [TFA] Enforce type cast during paramater check | Handle SocketTimeout with rados bench IO

harshkumarRH:
Jira tracker: [RHCEPHQE-22136](https://issues.redhat.com/browse/RHCEPHQE-22136)
With PR #4836 a change in was introduced in method `verify_set_config` under `tests/rados/monitor_configurations.py`, this change was trimming zeros at the end of the parameter value unconditionally, and resulted in incorrect comparison between the value set by the user and the value recognized/accepted/reported by the cluster.
Fix ensures trimming and type casting for better comparison of config parameters across the board.

Test modules modified:
- `tests/rados/monitor_configurations.py`

Logs: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-YU2COS

Jira tracker: [RHCEPHQE-22145](https://issues.redhat.com/browse/RHCEPHQE-22145)
Once an OSD is full, no pools can service IOs
If OSD is full before desired numbers of objects have been written, it results in a TimeOut Exception.
Although code is instructed to not check for exit_code, the timeout exception results in a failure and needs to be handled in a try except block

Test modules modified:
- `tests/rados/test_osd_full.py`

Logs:

### DIFF
--- a/tests/rados/test_osd_full.py
+++ b/tests/rados/test_osd_full.py
@@ -7,6 +7,7 @@ import datetime
 import json
 import time
 
+from ceph.ceph import SocketTimeoutException
 from ceph.ceph_admin import CephAdmin
 from ceph.rados import utils
 from ceph.rados.core_workflows import RadosOrchestrator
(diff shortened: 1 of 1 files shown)


pdhiran (approved):
Looks good

pdhiran (approved):
Looks good

PR #5330, opened 2025-10-31
Rados bench commands for omap iterations: read+writes

pdhiran:
Module tests OMAP operations on objects created by fill_omap_entries().

Objects are created using fill_omap_entries() method which:
- Creates objects named: omap_obj_{PID}_{index} (e.g., omap_obj_12345_0)
- Adds OMAP keys named: key_0, key_1, key_2, ... key_N
- Adds OMAP values named: value_0, value_1, value_2, ... value_N

Default Scale:
- 500 objects with 10,000 OMAP keys each = 5,000,000 total OMAP entries
- Uses efficient librados Python API for bulk creation

Tests include:
- Direct OMAP read operations (listomapkeys, listomapvals, getomapval)
- OMAP iteration and filtering
- Performance timing measurements

### DIFF
--- a/tests/rados/test_rados_bench_omap.py
+++ b/tests/rados/test_rados_bench_omap.py
@@ -1,0 +1,629 @@
+"""
+RADOS OMAP Performance and Functionality Testing
+
+This module tests OMAP operations on objects created by fill_omap_entries().
+
+Objects are created using fill_omap_entries() method which:
+- Creates objects named: omap_obj_{PID}_{index} (e.g., omap_obj_12345_0)
+- Adds OMAP keys named: key_0, key_1, key_2, ... key_N
+- Adds OMAP values named: value_0, value_1, value_2, ... value_N
+
+Default Scale:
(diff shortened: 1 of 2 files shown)


tests/rados/test_rados_bench_omap.py
```diff
@@ -0,0 +1,629 @@
...
+    """
+    cmd = "rados -p %s listomapkeys %s" % (pool, obj)
+    if max_return:
+        cmd += " --max-return %s" % max_return
+
+    start_time = time.time()
+    out, _ = node.exec_command(cmd=cmd, sudo=True, timeout=300)
+    elapsed = time.time() - start_time
+
+    key_count = len([line for line in out.split("\n") if line.strip()])
```
s-vipin: here key_count should equal max_return? for all max return I think the command is returning 10,000 keys
```
Result: 10000 keys in 1.01s (9925.33 keys/sec)
```

PR #5335, opened 2025-11-04
Fast EC optimization tests during staggered upgrades

pdhiran:
Test suite for Fast EC optimizations during partial cluster upgrades

        This test verifies the behavior of EC optimization parameters when the cluster
        is in a partially upgraded state with mixed daemon versions.

        Test covers 5 upgrade scenarios:
        1. Few or all Mgr's in 9.x only
        2. Few or Mons in 9.x + Few or all Mgr's in 9.x only
        3. Few OSDs in 9.x + Few or Mons in 9.x + Few or all Mgr's in 9.x only
        4. All core daemons (Mon, Mgr, OSD) in 9.x - Tests EC pool with optimizations
        5. All core daemons (Mon, Mgr, OSD) in 9.x + NO MDS upgraded - Tests backward compatibility:
           NEW OSDs (9.x) with EC optimizations work with older MDS clients

        Note: When all core daemons are upgraded:
        - Case 4 ALWAYS runs to test EC pool functionality
        - Case 5 ONLY runs if NO MDS are upgraded to 9.x (validates backward compatibility)

### DIFF
--- a/suites/reef/rados/tier-2_rados_test-mon-db-trimming.yaml
+++ b/suites/reef/rados/tier-2_rados_test-mon-db-trimming.yaml
@@ -290,7 +290,7 @@ tests:
   - test:
       name: Rados hotfix OSD beacons
       module: test_mon_osd_mgr_resiliency.py
-      polarion-id: TBD
+      polarion-id: CEPH-83632176
       config:
         test_duration: 10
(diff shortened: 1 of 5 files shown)


harshkumarRH (approved):
LGTM

PR #5358, opened 2025-11-10
Refactor DNS operations: add pagination and retry logic

sunilkumarn417:
- Extract DNS record operations into separate functions
- Add pagination (50 records/page) for list operations
- Add retry decorators for API and network exceptions
- Use DNS_RETRY_EXCEPTIONS tuple for consistent error handling

### DIFF
--- a/compute/ibm_vpc.py
+++ b/compute/ibm_vpc.py
@@ -5,7 +5,8 @@ import socket
 from copy import deepcopy
 from datetime import datetime, timedelta
 from time import sleep
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, Iterator, List, Optional
+from urllib.parse import parse_qs, urlparse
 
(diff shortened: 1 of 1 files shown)


compute/ibm_vpc.py
```diff
@@ -638,6 +598,225 @@ def _wait_until_vm_state(
...
+            instance_id=dns_svc_id,
+            dnszone_id=dns_zone_id,
+            limit=limit,
+            offset=offset,
+        )
+        return resp.get_result()
+
+    def _list_resource_records_with_pagination(
+        self, dns_svc_id: str, dns_zone_id: str, limit: int = 50
+    ) -> List[Dict]:
```
psathyan: Wdyt about using `iterator` instead this way we could avoid some of loops that we are introducing.

psathyan (commented):
/lgtm

PR #5381, opened 2025-11-14
fix: revert YAML configuration to match previous test

aarsharm:
Pass log: 
[cephci-run-ZMZD9A.zip](https://github.com/user-attachments/files/23543273/cephci-run-ZMZD9A.zip)


manasagowri (approved):
LGTM

PR #5412, opened 2025-11-24
Fix RHEL major version extraction logic

obannak:
# Description

Fixed the version parsing logic to correctly extract the major version number from RHEL version strings. The previous implementation was incorrectly parsing version strings, causing issues with multi-digit major versions (e.g., returning "1" instead of "10" for version "10.1").

This fix ensures that the major version is correctly extracted for all RHEL versions, including single-digit (8.x, 9.x) and multi-digit (10.x, 11.x, etc.) major versions.

The issue was discovered when testing with RHEL 10.1, which incorrectly returned "1" as the major version instead of "10".

### DIFF
--- a/tests/cephadm/test_client.py
+++ b/tests/cephadm/test_client.py
@@ -71,7 +71,7 @@ def add(cls, config: Dict) -> None:
             else:
                 rhcs_version = "default"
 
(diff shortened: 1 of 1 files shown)


vinpapnoi (approved):
LGTM

PR #5419, opened 2025-11-26
[RADOS] RHCEPHQE-22470: Add support for new qe-ceph-manifest file in RADOS test workflows

harshkumarRH:
Adds support for new qe-ceph-manifest file and methods. Removes deprecated methods.

Provides platform as rhel-9 as an explicit argument to bootstrap.py to fetch the appropriate repo to deploy older releases while testing Tentacle and above releases

### DIFF
--- a/suites/tentacle/rados/tier-2_rados_ec-pool_recovery.yaml
+++ b/suites/tentacle/rados/tier-2_rados_ec-pool_recovery.yaml
@@ -29,6 +29,7 @@ tests:
               args:
                 rhcs-version: 7.1
                 release: rc
+                platform: rhel-9
                 orphan-initial-daemons: true
                 skip-monitoring-stack: true
                 registry-url: registry.redhat.io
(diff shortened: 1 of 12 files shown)


suites/tentacle/rados/tier-2_rados_ec-pool_recovery.yaml
```diff
@@ -29,6 +29,7 @@ tests:
               args:
                 rhcs-version: 7.1
```
pdhiran: The changes look good, but along with this, can we also enhance the tests to use latest Z release for the given major release? ( basically use the latest CDN builds ). 

This changes can be made if the test does not explicitly require an older version of ceph specifically.

pdhiran: OK, this can be taken up in later PR's as well. this PR will hold us back from running upgrade pipeline for rados. 

This PR can be merged

pdhiran (commented):
looks good

pdhiran (approved):
looks good.

PR #5434, opened 2025-12-02
[TFA][DMFG] Remove NFS test from the DMFG suite

vinpapnoi:
# Description

Since NFS related tests have been added separately in all pipelines, removing it from DMFG suite.


obannak (approved):
LGTM

PR #5447, opened 2025-12-04
Zero append tests for Fast EC pools

pdhiran:
Test: Fast EC Space Gain Verification

        Verifies that EC optimizations prevent zero-padding for objects smaller than
        the stripe_unit, resulting in significant space savings at both pool and shard levels.

        Test Steps:
        1. Disable global EC optimizations flag
        2. Create EC pool WITHOUT optimizations (test_ec_pool_without_optimizations)
        3. Create EC pool WITH optimizations enabled (test_ec_pool_with_optimizations)
        4. Write 100 identical 4KB objects to both pools using 'rados put'
        5. Wait for stats to stabilize and trigger deep-scrub
        6. Compare pool-level space usage (bytes_used from 'ceph df detail')
        7. Verify shard-level sizes using:
           - rados stat: Confirms logical object size (4KB) is same for both pools
           - ceph-objectstore-tool: Confirms actual shard size differs (16KB vs 4KB)

        EC Configuration (k=2, m=2, stripe_unit=16KB):
        - stripe_unit = 16KB (chunk size per data shard)
        - stripe_width = k * stripe_unit = 32KB
        - full stripe = (k + m) * stripe_unit = 64KB

        Expected Results:
        - Without optimizations: 4KB object padded to stripe_unit → shard = 16KB
        - With optimizations: 4KB object stored as-is → shard = 4KB
        - Pool-level space savings: ~80% (4x reduction)
        - Shard-level size ratio: 4x (16KB vs 4KB)

        Failure Conditions:
        - Pool with optimizations doesn't use less bytes than pool without
        - Space savings < 50% threshold
        - rados stat object size != 4096 bytes
        - ceph-objectstore-tool shard size: pool1 <= pool2

### DIFF
--- a/tests/rados/test_fast_ec_config_params.py
+++ b/tests/rados/test_fast_ec_config_params.py
@@ -7,12 +7,17 @@ This module tests the behavior of erasure-coded pools with various configuration
 - Enabling/disabling allow_ec_optimizations at pool level
 - Compression with EC pools
 - Read/write operations with different EC optimization states
+- Space savings verification for objects smaller than stripe_unit (zero-padding prevention)
 """
 
+import json
+import random
+import re
 import time
 
 from ceph.ceph_admin import CephAdmin
 from ceph.rados.core_workflows import RadosOrchestrator
+from ceph.rados.objectstoretool_workflows import objectstoreToolWorkflows
 from tests.rados.monitor_configurations import MonConfigMethods
 from utility.log import Log
 
@@ -31,8 +36,11 @@ def run(ceph_cluster, **kw):
         kw: Test configuration parameters
 
     Supported test scenarios:
-        - test_fast_ec_optimization_params: Comprehensive verification of EC pool optimization parameters
-          with default configurations, runtime changes, and RBD data operations
+        - test_fast_ec_optimization_params: Comprehensive verification of EC pool optimization
+          parameters with default configurations, runtime changes, and RBD data operations
+        - test_fast_ec_space_gain: Verifies space savings when EC optimizations prevent
+          zero-padding for objects smaller than stripe_unit. Compares pool-level bytes_used
+          and shard-level sizes using ceph-objectstore-tool.
 
     Test Requirements:
         - Ceph version 9.0 or above (squid+)
@@ -46,6 +54,7 @@ def run(ceph_cluster, **kw):
     cephadm = CephAdmin(cluster=ceph_cluster, **config)
     rados_obj = RadosOrchestrator(node=cephadm)
     mon_obj = MonConfigMethods(rados_obj=rados_obj)
+    cot_obj = objectstoreToolWorkflows(node=cephadm)
     client_node = ceph_cluster.get_nodes(role="client")[0]
 
     if config.get("test_fast_ec_optimization_params"):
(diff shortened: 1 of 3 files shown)


tests/rados/test_fast_ec_config_params.py
```diff
@@ -935,6 +944,659 @@ def run(ceph_cluster, **kw):
...
+            # Step 4: Write objects smaller than stripe width (16KB) to both pools
+            # Writing 4KB objects to demonstrate zero-padding difference
+            # Note: We use 'rados put' instead of 'rados bench' because rados bench
+            # rounds up object sizes to stripe_width for EC pools.
+            # Without optimizations: 4KB padded to 16KB stripe_unit = 4x storage overhead
+            # With optimizations: 4KB stored as-is = normal 2x EC overhead
+            log.info(
+                "Step 4: Writing objects (4KB each, smaller than 16KB stripe width) to both pools"
+            )
+            num_objects = 100
```
SrinivasaBharath: In the test steps mentioned that writing 200  identical 4KB objects but num_objects are 100.

tests/rados/test_fast_ec_config_params.py
```diff
@@ -935,6 +944,659 @@ def run(ceph_cluster, **kw):
...
+                # Get list of objects in the pool using run_ceph_command (returns JSON)
+                obj_list = rados_obj.run_ceph_command(cmd=f"rados -p {pool_name} ls")
+                objects = [obj["name"] for obj in obj_list] if obj_list else []
+
+                if not objects:
+                    log.warning("No objects found in pool %s", pool_name)
+                    continue
+
+                # Select a random object for verification
+                test_obj = random.choice(objects)
```
SrinivasaBharath: By using "rados -p <pool_name> ls | shuf -n 1"  we can get an random object.

tests/rados/test_fast_ec_config_params.py
```diff
@@ -935,6 +944,659 @@ def run(ceph_cluster, **kw):
...
+                section="global",
+                name="osd_pool_default_flag_ec_optimizations",
+                value="false",
+            )
+            log.info(
+                "Set osd_pool_default_flag_ec_optimizations to false at global level"
+            )
+            time.sleep(2)
+
+            # Step 2: Create first EC pool (WITHOUT optimizations) with 16K stripe_unit
```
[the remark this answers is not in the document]
pdhiran: done, created a new method to create both the pools as required

tests/rados/test_fast_ec_config_params.py
```diff
@@ -935,6 +944,659 @@ def run(ceph_cluster, **kw):
...
+                )
+                log.info("  - USED bytes saved: %s bytes", space_saved)
+                log.info("  - Percentage saved: %.2f%%", space_saved_percent)
+                log.info("=" * 80)
+
+                # Calculate expected space savings dynamically based on stripe_unit and object_size
+                # Expected savings = (stripe_unit - object_size) / stripe_unit * 100
+                # Without optimizations: object padded to stripe_unit
+                # With optimizations: object stored as-is
+                stripe_unit_int = int(stripe_unit)
```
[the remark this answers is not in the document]
pdhiran: Updated docstrings.

note that this is not a officially provided formula, it is a guess based on the behavior of the code and object size we write and expect, And we just take into account the object size, not omap or metadata size associated with the object, so it's just approximate value.

tests/rados/test_fast_ec_config_params.py
pdhiran: I prefer the current implementation. let me know if the changes are mandatory.

PR #5456, opened 2025-12-05
Fix DMFG Upgrade IBM Suites Configuration

obannak:
`custom_image` and `custom_repo` added in the bootstrap configuration were not used by the bootstrap script, so the default Cephadm image/repo gets selected instead — which leads to failures like:

`Error: Unable to find a match: cephadm-2:20.1.0-112*.el9cp`

These fields replaced with the `rhcs-version` and `release` to pick correct image and repo are during upgrades.

Log:
- https://149.81.216.83/view/Tentacle/job/tentacle-upgrade-ci/33/


vinpapnoi (approved):
LGTM

PR #5470, opened 2025-12-09
Updating Fast EC test variable from chunk_size to stripe_unit

pdhiran:
1. Updated variable name
2. Added test to confirm the pool created vs config passed
3. Added docstring to add tests run for fast EC


harshkumarRH (approved):
Looks good!

PR #5495, opened 2025-12-15
Workflow to verify file contents are preserved after unmount-remount 

pdhiran:
…with kernel client

When a file is written to a CephFS subvolume and the container/mount is unmounted and remounted, the file size is correct but the contents are all NUL bytes instead of the written data. Addition of test workflow to verify the behaviour.

Polarion : https://polarion.engineering.redhat.com/polarion/#/project/CEPH/workitem?id=CEPH-83632223
bugzilla : https://bugzilla.redhat.com/show_bug.cgi?id=2421457


neha-gangadhar (approved):
Approved.

PR #5496, opened 2025-12-16
Fixed Delay before mirroring as per dev's suggestion

aarsharm:
Pass log: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-6HQSUZ/

As per dev's suggestion the delay flag needs to be set before mirroring starts. Hence removed the prior IO and sync


sunilangadi2 (changes requested):
Let's fix these suite changes in Quincy and reef as well.

chaitanyadommeti (approved):
lgtm

Manohar-Murthy (approved):

PR #5502, opened 2025-12-16
Workflow to verify file contents are preserved after unmount-remount with fuse mount

pdhiran:
When a file is written to a CephFS subvolume and the container/mount is unmounted and remounted, the file size is correct but the contents are all NUL bytes instead of the written data. Addition of test workflow to verify the behaviour.

With PR : https://github.com/red-hat-storage/cephci/pull/5495, added coverage for Kernel mount

With this PR, adding coverage for fuse mount

### DIFF
--- a/tests/rados/pool_tests.py
+++ b/tests/rados/pool_tests.py
@@ -700,6 +700,8 @@ def run(ceph_cluster, **kw):
         return 0
 
     # Blocked with bug : https://bugzilla.redhat.com/show_bug.cgi?id=2172795. Bug fixed.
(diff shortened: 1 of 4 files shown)


tests/rados/pool_tests.py
```diff
@@ -756,6 +758,10 @@ def run(ceph_cluster, **kw):
                 log.debug(
                     f"autoscale status on pools: \n{rados_obj.node.shell(['ceph osd pool autoscale-status'])}"
                 )
+                if float(rhbuild.split("-")[0]) > 8.1:
```
harshkumarRH: Shouldn't this be >=9.0 ?

harshkumarRH (approved):
LGTM

PR #5512, opened 2025-12-18
Addition of Thrash OSD test in rados

pdhiran:
OSD Thrashing Test with Concurrent I/O

This module performs aggressive OSD thrashing with concurrent I/O workload to verify
cluster stability under stress conditions. It tests both EC and replicated pools,
and checks for any crashes or issues during OSD state transitions.


pdhiran:
Pass log :  http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-UEGZOT

PR #5577, opened 2026-01-19
Added mon service for node5 .

obannak:
# Description

On IBM clud the issue of deploying monitoring secrvice node-exporter is failing occasionally which looks like system issue. All other nodes in cluster have running either mon/mgr services but node on which either of this service is not running is failing so added `mon` service to the node.

Log: https://149.81.216.83/job/tentacle-test-executor/610/


MohitBis (approved):
LGTM

mkasturi18 (commented):
lgtm

mkasturi18 (approved):
lgtm

PR #5651, opened 2026-02-16
Add CephFS mirroring system tests for Mirror daemon component

kdhaduk81:
This test verifies that:
1. Background IOs continue running during Mirror daemon failure operations
2. Snapshot sync is not affected by Mirror daemon failure operations
3. Cluster remains healthy after each operation
4. Clients reconnect properly after Mirror daemon failure operations

Test operations performed:
- SIGTERM: Graceful termination signal
- SIGHUP: Reload configuration signal
- SIGKILL: Forceful kill signal
- Systemctl Restart: Restart mirror daemon using systemctl restart ceph-mds@<id>
- Node Reboot: Reboot mirror daemon nodes
- mirror daemon container restart
- mirror daemon redeploy

Logs:  http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-ADJTKG


hkadam134 (dismissed):

hkadam134:
Tox failing - please fix them,
Also, the test files look too long - do you want to move any reuseable function to utils file ?

kdhaduk81:
> Tox failing - please fix them, Also, the test files look too long - do you want to move any reuseable function to utils file ?

Hi Hemanth, Thank you for the review. I will fix the tox failures. I agree with the big test file but since there are 7 tests in file is big. I can not move the functions to common util as well because they are wrappers already using some library from the common utils and they are specific to particular daemon. It should be possible to generalise them but it might need major code restructuring.

hkadam134 (dismissed):

tests/cephfs/cephfs_mirroring_system/test_cephfs_mirroring_system_mirror_daemon.py
```diff
@@ -0,0 +1,1200 @@
...
+            service_name = fs_util_v1.deamon_name(node, r"cephfs-mirror\.")
+            log.info("Restarting cephfs-mirror service: %s", service_name)
+            node.exec_command(
+                sudo=True,
+                cmd=f"systemctl restart {service_name}",
+            )
+
+        # Wait for cephfs-mirror to recover
+        log.info("Waiting for cephfs-mirror to recover after systemctl restart")
+        time.sleep(30)
```
Manimaran-MM: Same comment as above

Manimaran-MM (changes requested):

kdhaduk81:
> > > Tox failing - please fix them, Also, the test files look too long - do you want to move any reuseable function to utils file ?
> > 
> > 
> > Hi Hemanth, Thank you for the review. I will fix the tox failures. I agree with the big test file but since there are 7 tests in file is big. I can not move the functions to common util as well because they are wrappers already using some library from the common utils and they are specific to particular daemon. It should be possible to generalise them but it might need major code restructuring.
> 
> No Issue - We might need to rework once the new mirroring enhancements are in place. We can go ahead with this now..

@hkadam134 Fixed the issue. Thank you.

hkadam134 (dismissed):
/lgtm

hkadam134:
Looks like there is a commit from Vipin part of the PR - can you check once

hkadam134 (approved):
/lgtm

Manimaran-MM (approved):
LGTM

PR #5672, opened 2026-02-23
Add rhel 9 repo to cdn repo

kdhaduk81:
Added cdn repo for rhel-9 to reolve the configure client test.
Failed logs: http://magna002.ceph.redhat.com/ceph/ceph-qe-logs/kdhaduk/logs_ibm_fixed1/
Passed logs: http://magna002.ceph.redhat.com/ceph/ceph-qe-logs/kdhaduk/logs_ibm_fixed2/


psathyan (commented):
Client nodes were treated differently due to the need of having support of client workloads from different ceph version and operating system. We support custom images via the Node object here https://github.com/red-hat-storage/cephci/blob/main/ceph/utils.py#L456

This is one of the reason, why we don't have it linked to the common steps as part of `misc_env/install_prereq.py`

Another reason is that the cluster needs to have homogenous nodes while clients can be heterogenous. 

Honestly, I would like to keep this flexibility. However, there is not much data wherein this flexibility is used.

PR #5674, opened 2026-02-23
conf: add Rocky-10.1 inventory for upstream builds

vamahaja:
Changes consists of -
- Add Rocky 10.1 inventory
- Add epel repos for rocky-10
- Update bootstrap steps to use upstream cephadm package rpm

### DIFF
--- a/ceph/ceph_admin/__init__.py
+++ b/ceph/ceph_admin/__init__.py
@@ -175,12 +175,6 @@ class CephAdmin(BootstrapMixin, ShellMixin, RegistryLoginMixin):
         Args:
             repo_url: repo file URL link (default: None)
         """
-        EPEL_REPOS = {
-            "7": "https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm",
-            "8": "https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm",
-            "9": "https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm",
-        }
-
         if not repo_url:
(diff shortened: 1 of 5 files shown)


ceph/ceph_admin/__init__.py
```diff
@@ -191,10 +185,24 @@ def setup_upstream_repository(self, repo_url=None):
             )
             node.exec_command(sudo=True, cmd="yum update metadata", check_ec=False)
 
+            # Install EPEL based on RHEL version
+            epel_repo = ""
+            version_id = node.distro_info["VERSION_ID"][0]
+            if version_id.startswith("8"):
```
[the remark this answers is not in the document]
vamahaja: done

conf/inventory/rocky-10.1-server-x86_64.yaml
```diff
@@ -0,0 +1,41 @@
+---
+version_id: 10.1
+id: rhel
+instance:
+  create:
+    image-name: teuthology-rocky-10.1-x86_64
+    vm-size: ci.standard.medium
```
psathyan: I would like to avoid having inventory files based on sizes instead opt to use `--custom-config openstack_flavor=ci.standard.medium` like we have for `vpc_profile`

psathyan (commented):
minor nit: If support is not there then lets go ahead

conf/inventory/rocky-10.1-server-x86_64.yaml
vamahaja: @pragadee older pipeline doesn't support this one

PR #5683, opened 2026-03-03
TFA Fixs for weekly pipeline

kdhaduk81:
1. test_clone_delete_with_FS_fail.py
Issue:
The test was intermittently failing when attempting to cancel a clone operation that had already completed by the time the cancel request was issued.
Fix:
Added a validation step to check whether the clone is still in progress before attempting cancellation. The test now only triggers the cancel operation if the clone state is in-progress, preventing false failures due to timing conditions.

Added cluster health check at the start of the test. Add log information if health is not ok. Other minor fixes to initialise variables so that test does not fail with var not initialised in finally.
Logs: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-HSP5N4/

### DIFF
--- a/tests/cephfs/cephfs_bugs/test_clone_delete_with_FS_fail.py
+++ b/tests/cephfs/cephfs_bugs/test_clone_delete_with_FS_fail.py
@@ -1,3 +1,4 @@
+import json
 import random
 import string
 import traceback
@@ -5,6 +6,7 @@ import traceback
 from ceph.ceph import CommandFailed
 from ceph.parallel import parallel
 from tests.cephfs.cephfs_utilsV1 import FsUtils
+from tests.cephfs.lib.cephfs_common_lib import CephFSCommonUtils
 from utility.log import Log
 from utility.retry import retry
 
@@ -57,6 +59,7 @@ def run(ceph_cluster, **kw):
         build = config.get("build", config.get("rhbuild"))
         fs_util.prepare_clients(clients, build)
         fs_util.auth_list(clients)
+        cephfs_common_utils = CephFSCommonUtils(ceph_cluster)
         log.info("checking Pre-requisites")
         if len(clients) < 1:
             log.info(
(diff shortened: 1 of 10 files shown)


tests/cephfs/cephfs_bugs/test_clone_delete_with_FS_fail.py
```diff
@@ -57,17 +59,27 @@ def run(ceph_cluster, **kw):
...
             cmd="ceph config set mgr mgr/volumes/snapshot_clone_no_wait false",
         )
+        log.info("Simulate clone create delay")
+        client1.exec_command(
+            sudo=True,
+            cmd="ceph config set mgr mgr/volumes/snapshot_clone_delay 2",
+        )
+        log.info("Check ceph Health before starting the test")
+        if cephfs_common_utils.wait_for_healthy_ceph(client1):
+            raise Exception("Cluster health is not OK before starting the test")
```
Manimaran-MM: Please have log.error instead of Exception and have return 1 so that it becomes easy to debug on the failure

tests/cephfs/cephfs_bugs/test_clone_delete_with_FS_fail.py
```diff
@@ -144,18 +157,30 @@ def run(ceph_cluster, **kw):
...
+        client1.exec_command(
             sudo=True, cmd="ceph config set mgr mgr/volumes/snapshot_clone_no_wait true"
         )
+        client1.exec_command(
+            sudo=True,
+            cmd="ceph config set mgr mgr/volumes/snapshot_clone_delay 0",
+        )
         fs_util.remove_snapshot(client1, **snapshot)
         fs_util.remove_subvolume(client1, **subvolume)
+        if cephfs_common_utils.wait_for_healthy_ceph(client1):
```
Manimaran-MM: Same comment here as well

kdhaduk81: wait_for_healthy_ceph(self, client, wait_time=300): Default wait_time is 300s so I did not passed while calling. I hope that is fine.

hkadam134 (approved):
/lgtm

Manimaran-MM (approved):

PR #5687, opened 2026-03-04
fix: enable CodeReady Builder (CRB) for Ceph build dependencies

vamahaja:
Fix ceph dependency issue for rocky10 build by enabling CodeReady Builder (CRB)


psathyan (commented):
/lgtm

PR #5761, opened 2026-04-08
Added backoff=1 to avoid test hung for more than 2 days

kdhaduk81:
1. Replace manual retry loop in validate_snapshot_sync_status with the @retry decorator to eliminate nested outer/inner retry logic that caused jobs to hang for 2+ days when dnf install failed
2. Add backoff=1 to @retry decorators to enforce constant delay instead of exponential backoff
3. Add wait_for_sync_idle() helper to wait until all mirroring paths reach idle state before asserting sync counts
4. Remove snap schedules before capturing final snapshot counts to prevent race conditions
5. Fix bug where snaps_synced2 was incorrectly checking subvol1_path instead of subvol2_path
6. Make dnf install ceph-common non-fatal by wrapping it in try/except

Logs: http://10.64.24.74/logs/ceph-qe-logs/IBM/9.1/rhel-10/executor/20.2.1-135/cephfs/1163/logs/tier_1_cephfs_mirror/

### DIFF
--- a/tests/cephfs/cephfs_mirroring/cephfs_mirroring_utils.py
+++ b/tests/cephfs/cephfs_mirroring/cephfs_mirroring_utils.py
@@ -809,7 +809,7 @@ class CephfsMirroringUtils(object):
         log.error("last synced Snapshot not found or not synced")
         raise CommandFailed("last synced Snapshot not found or not synced")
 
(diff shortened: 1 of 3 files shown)


Manimaran-MM (approved):
LGTM

PR #5762, opened 2026-04-09
Addition of RDMA support for Thrash tests

pdhiran:
- Simplified NFS host selection to use ceph orch host ls as source of truth:
- if nfs_placement_label is set, use matching orch-labeled hosts;
- if no label match, fall back to all orch hosts.
- Hardened orch host parsing to safely handle both list and dict JSON shapes and avoid type-related failures.
- Updated docs/comments in core_workflows.py and test_osd_thrashing.py so they match the new orch-first placement behavior.
- Expanded workflow_results initialization in test_osd_thrashing.py to include NFS/MDS/subvolume workflows for complete iteration summary keys.

log : http://magna002.ceph.redhat.com/ceph-qe-logs/cephci-run-MJOA5Y/

### DIFF
--- a/ceph/rados/core_workflows.py
+++ b/ceph/rados/core_workflows.py
@@ -18,6 +18,7 @@ import random
 import re
 import time
 from collections import namedtuple
+from typing import Optional
 
 from ceph.ceph import CommandFailed, SocketTimeoutException, TimeoutException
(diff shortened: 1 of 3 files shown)


harshkumarRH (approved):
Looks good!

PR #5775, opened 2026-04-14
Fix yum_repos gpgcheck typo and Rocky SSH key YAML indent

rakeshcn:
- Replace mistyped gpcheck with gpgcheck for BaseOS/CRB in IBM EU RHEL 9.x inventories (cloud-init ignores unknown keys).
- Normalize ssh-authorized-keys list indent to 10 spaces in Rocky 10.1 inventories to match other server profiles.

Made-with: Cursor


psathyan (dismissed):
Thanks @rakeshcn good catch.

psathyan (approved):

PR #5779, opened 2026-04-14
Add BlueStore ESB extent map resharding verification

pdhiran:
Add enable_esb_verification option to test_osd_thrashing.py that configures BlueStore to exercise the elastic shared blobs extent map resharding path (Tracker IBMCEPH-10839). When enabled, applies bluestore_elastic_shared_blobs=true, bluestore_write_v2=false, osd_memory_target=939524096, bluestore_onode_segment_size=0, and bluestore_debug_extent_map_encode_check=true with debug_bluestore=5/5 to validate extent maps on every encode during OSD thrashing. Includes a new tier-3 suite entry (CEPH-83632230) targeting EC pools with overwrites and crush thrashing to reproduce the crash scenario.

### DIFF
--- a/suites/tentacle/rados/tier-3_rados_osd_thrashing_tests.yaml
+++ b/suites/tentacle/rados/tier-3_rados_osd_thrashing_tests.yaml
@@ -367,6 +367,45 @@ tests:
         enable_cephfs_subvolume_thrashing: false
 
   # ===========================================================================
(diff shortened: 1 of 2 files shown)


harshkumarRH (approved):
Looks good

neha-gangadhar (approved):
lgtm

PR #5780, opened 2026-04-15
Configure configure_subsystems() function is broken because of PR#5758

jprakash-redhat:
configure_subsystems() function is broken because of PR#5758

Commit for reference https://github.com/red-hat-storage/cephci/commit/a1a79fea136476547ac9f005df884f5d460cd45b#diff-1e84b39c7d4def5cd8e73cf78c640daf860bbd58249d14ff4ed19dc930fe4a1fR8-R61

before it was 
`def configure_subsystems(nvme_service):`

After above commit it is was changed to below
`def configure_subsystems(nvme_service, ceph_cluster=None):`

Inside configure_subsystem() below lines of code is introduced because of above commit
```
ceph_version = get_ceph_version_from_cluster(
            ceph_cluster.get_nodes(role="client")[0]
        )
```

When ever ceph_cluster is None then it will fail as below error in all scripts
`ERROR - 'NoneType' object has no attribute 'get_nodes'`

I didn't collected the logs because entire regression suite scripts are changed here and need to re-execute regression suite

### DIFF
--- a/tests/nvmeof/test_ceph_nvmeof_alerts_events.py
+++ b/tests/nvmeof/test_ceph_nvmeof_alerts_events.py
@@ -412,7 +412,7 @@ def test_ceph_83610948(ceph_cluster, config):
     ha = HighAvailability(ceph_cluster, config["gw_nodes"], **config)
     ha.gateways = nvme_service.gateways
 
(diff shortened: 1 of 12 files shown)


chaitanyadommeti (approved):
lgtm

manasagowri (approved):

PR #5790, opened 2026-04-17
Fixed check_nfs_daemons_removed_retry() Logic

aravindrrh:
# Description

The GRPC Failures were due to previous test case not cleaning up the NFS-Cluster.
I have fixed the logic in check_nfs_daemons_removed_retry()

Here are the IBMC logs: 

[http://magna002.ceph.redhat.com/cephci-jenkins/ibm-cos/RH/9.1/rhel-10/executor/20.2.1-179/nfs/1224/logs/tier0_nfs_ganesha_func_sanity/](http://magna002.ceph.redhat.com/cephci-jenkins/ibm-cos/RH/9.1/rhel-10/executor/20.2.1-179/nfs/1224/logs/tier0_nfs_ganesha_func_sanity/)

### DIFF
--- a/tests/nfs/nfs_operations.py
+++ b/tests/nfs/nfs_operations.py
@@ -830,7 +830,7 @@ def check_nfs_daemons_removed(client):
     check_nfs_daemons_removed_retry(client)
 
 
-@retry(OperationFailedError, tries=30, delay=10, backoff=1)
+@retry(OperationFailedError, tries=10, delay=10, backoff=1)
 def check_nfs_daemons_removed_retry(client):
     """
     Helper function to check if NFS daemons are removed.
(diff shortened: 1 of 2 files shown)


tests/nfs/nfs_operations.py
```diff
@@ -837,14 +837,14 @@ def check_nfs_daemons_removed_retry(client):
...
-        # hence we check if not err, and not if err
-        if out:
-            raise OperationFailedError("NFS daemons are still present")
-    except Exception as e:
-        log.warning(f"Caugt Exception: {e}")
+        count = int(out.strip())
+    except ValueError:
+        raise OperationFailedError(f"Unexpected output: {out}")
+
+    if count > 0:
```
Manimaran-MM: CAn we move this inside try block?
Also line number 840 to be moved inside try block

aravindrrh: Changed the logic as the call returns "No services reported" when there are no NFS services

tests/nfs/nfs_operations.py
```diff
@@ -830,24 +830,25 @@ def check_nfs_daemons_removed(client):
...
+        if "No services reported" in out:
+            log.info("All NFS daemons have been removed.")
+            return True
+        else:
+            raise OperationFailedError("NFS daemons still present")
 
-    log.info("All NFS daemons have been removed.")
-    return True
+    except Exception as err:
+        raise OperationFailedError(f"Unexpected Error: {err} / Output: {out}")
```
Manimaran-MM: All looks good.. Can we make this OperationFailedError to log.error with return 1?
So that any other failures will abruptly stop rather than retrying for 10 times.

aravindrrh: Done 👍

Manimaran-MM (approved):

PR #5795, opened 2026-04-21
Fix log rotate in OSD thrashing and add unique name for RBD images

harshkumarRH:
Log rotation is essential to contain the space utilization on OS disk.
Current code was attempting log rotation with OSD hostname, switched it to use OSD host object.

During continuous runs, it was observed that having the same RBD image name was causing issues as clean-up at times would not work. This resulted in failure of re-creation of the image during subsequent runs. To overcome this, unique RBD image name is used for each run.

### DIFF
--- a/tests/rados/test_osd_thrashing.py
+++ b/tests/rados/test_osd_thrashing.py
@@ -109,6 +109,7 @@ from typing import Any, Dict, List, Optional
 
 import yaml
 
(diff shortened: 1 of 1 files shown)


s-vipin (approved):
Looks good to me

pdhiran (approved):
looks good

PR #5798, opened 2026-04-21
Revert the fix for IBMCEPH-13427

vinpapnoi:
# Description

As a workaround for the bug [IBMCEPH-13427](https://ibm-ceph.atlassian.net/browse/IBMCEPH-13427) , the `--automatically-accept-license` flag was added for both IBM and RH ceph deployments.

Since, this flag is only applicable for IBM ceph and the bug has been fixed, reverting the workaround to work for only IBM clusters.


obannak (approved):
LGTM

psathyan (approved):

PR #5799, opened 2026-04-21
Tfa fixes for regression + extended regression suite

kdhaduk81:
Regression Suite:
1. move_data_bw_nfs_and_cephfs_mounts.py: Added extra_params for fs= on kernel mount and --client_fs on fuse mount so the correct filesystem is targeted. Initialised variables to avoid cleanup crash.
2. nfs_ro_rw_access.py: Fixed sed pattern from 's/RW/RO/g' to 's/rw/ro/gI' (case-insensitive) because the NFS export config now uses lowercase rw/ro.
Logs: http://10.64.24.74/logs/ceph-qe-logs/IBM/9.1/rhel-10/regression/20.2.1-173/cephfs/48/logs/

Extended Regression:
1. test_cephfs_pool_size_change.py: Remove the test filesystem and avoid leftover health warnings instead of setting the config back.
2. snap_rollback_with_node_reboots.py: Broke one large parallel() block into 4 separate parallel() blocks. Previously, checksum verification ran inside the parallel context (racing with the revert), causing false failures. Now each snapshot revert + reboot runs in its own parallel block, and checksum validation happens after each block completes. Also moved rm -rf * out of the parallel block.
3. mds_nfs_node_failure_ops.py: Wrapped IO loop body in try/except for CommandFailed and generic Exception, since IO failures are expected during node-disruption tests. Without this, a transient IO error would abort the entire test.
4. cephfs_system_utils.py: Added mkdir -p /etc/ceph before copying ceph.conf and keyring files, fixing "No such file or directory" errors on nodes where /etc/ceph doesn't exist yet.
5. cephfs_top_dump.py: Added fs_util.wait_for_mds_process() after filesystem creation. 
6. test_selinux_relabel.py: modified file creation with xargs -n 500 -P 16 for parallel file creation (1M files).  It was failing with timeout issue at file creation.

Logs: http://10.64.24.74/logs/ceph-qe-logs/IBM/8.1/rhel-9/extended_regression/19.2.1-355/cephfs/4/logs/

### DIFF
--- a/tests/cephfs/cephfs_system/mds_nfs_node_failure_ops.py
+++ b/tests/cephfs/cephfs_system/mds_nfs_node_failure_ops.py
@@ -24,6 +24,8 @@ def start_io_time(fs_util, client1, mounting_dir, timeout=300):
     global stop_flag
     stop_flag = False
     iter = 0
+    success = 0
+    failures = 0
     if timeout:
         stop = datetime.now() + timedelta(seconds=timeout)
     else:
@@ -33,12 +35,31 @@ def start_io_time(fs_util, client1, mounting_dir, timeout=300):
         if stop and datetime.now() > stop:
             log.info("Timed out *************************")
             break
-        client1.exec_command(sudo=True, cmd=f"mkdir -p {mounting_dir}/run_ios_{iter}")
-        fs_util.run_ios(client1, f"{mounting_dir}/run_ios_{iter}/", io_tools=["dd"])
-        client1.exec_command(
-            sudo=True, cmd=f"rm -rf {mounting_dir}/run_ios_{iter}", timeout=3600
-        )
+        try:
+            client1.exec_command(
+                sudo=True, cmd=f"mkdir -p {mounting_dir}/run_ios_{iter}"
+            )
+            fs_util.run_ios(client1, f"{mounting_dir}/run_ios_{iter}/", io_tools=["dd"])
+            success += 1
+        except CommandFailed as e:
+            failures += 1
+            log.warning(f"IO command failed during disruption (expected): {e}")
+        except Exception as e:
+            failures += 1
+            log.warning(f"IO operation hit an error during disruption: {e}")
+        finally:
+            client1.exec_command(
+                sudo=True,
+                cmd=f"rm -rf {mounting_dir}/run_ios_{iter}",
+                timeout=3600,
+                check_ec=False,
+            )
         iter = iter + 1
+
+    log.info(f"IO loop summary: success={success}, failures={failures}")
+    if success == 0:
+        log.error("No successful IO during disruption window")
+        raise Exception("IO loop completed with zero successful iterations")
 
 
 @retry(CommandFailed, tries=3, delay=60)
--- a/tests/cephfs/cephfs_system/snap_rollback_with_node_reboots.py
+++ b/tests/cephfs/cephfs_system/snap_rollback_with_node_reboots.py
@@ -155,12 +155,14 @@ def run(ceph_cluster, **kw):
         files_checksum_snap_4 = fs_util.get_files_and_checksum(
             client1, fuse_mounting_dir_1
         )
+        retry_revert = retry(CommandFailed, tries=3, delay=60)(client1.exec_command)
+
+        client1.exec_command(
+            sudo=True,
+            cmd=f'test -n "{kernel_mounting_dir_1}" && cd "{kernel_mounting_dir_1}" && rm -rf -- *',
+        )
+
         with parallel() as p:
-            client1.exec_command(
-                sudo=True,
-                cmd=f"cd {kernel_mounting_dir_1};rm -rf *",
-            )
-            retry_revert = retry(CommandFailed, tries=3, delay=60)(client1.exec_command)
             p.spawn(
                 retry_revert,
                 sudo=True,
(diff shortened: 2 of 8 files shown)


tests/cephfs/cephfs_system/snap_rollback_with_node_reboots.py
```diff
@@ -155,12 +155,14 @@ def run(ceph_cluster, **kw):
         files_checksum_snap_4 = fs_util.get_files_and_checksum(
             client1, fuse_mounting_dir_1
         )
+        retry_revert = retry(CommandFailed, tries=3, delay=60)(client1.exec_command)
+
+        client1.exec_command(
+            sudo=True,
+            cmd=f"cd {kernel_mounting_dir_1};rm -rf *",
```
Manimaran-MM: unsafe rm -rf * (can delete wrong directory if cd fails)

**What’s there now**
`cmd=f"cd {kernel_mounting_dir_1};rm -rf *"`

**Why this is a blocker**
If cd fails (bad path, not mounted yet, variable empty/unexpected), the rm -rf * may run in whatever the current directory is for that remote session. That’s catastrophic on CI hosts.

**How to fix (use fail-closed commands + path guard)**
Use && (not ;) and add a prefix guard to ensure the directory is what we expect.

**Example replacement:**
Safer single-line command:
`cmd=f'test -n "{kernel_mounting_dir_1}" && echo "{kernel_mounting_dir_1}" | grep -q "^/mnt/cephfs_kernel" && cd "{kernel_mounting_dir_1}" && rm -rf -- *'`

Even safer (avoid glob edge-cases, delete contents without relying on *):
`cmd=f'test -n "{kernel_mounting_dir_1}" && echo "{kernel_mounting_dir_1}" | grep -q "^/mnt/cephfs_kernel" && find "{kernel_mounting_dir_1}" -mindepth 1 -maxdepth 1 -exec rm -rf -- {{}} +'`

If your environment guarantees bash, you can also do:
`cmd=f'cd "{kernel_mounting_dir_1}" && rm -rf -- ./*' (still prefer the prefix guard)`

tests/cephfs/cephfs_system/mds_nfs_node_failure_ops.py
```diff
@@ -33,11 +33,20 @@ def start_io_time(fs_util, client1, mounting_dir, timeout=300):
...
-        fs_util.run_ios(client1, f"{mounting_dir}/run_ios_{iter}/", io_tools=["dd"])
-        client1.exec_command(
-            sudo=True, cmd=f"rm -rf {mounting_dir}/run_ios_{iter}", timeout=3600
-        )
+        try:
+            client1.exec_command(
+                sudo=True, cmd=f"mkdir -p {mounting_dir}/run_ios_{iter}"
+            )
+            fs_util.run_ios(client1, f"{mounting_dir}/run_ios_{iter}/", io_tools=["dd"])
+            client1.exec_command(
```
Manimaran-MM, line 51: **What’s there now**
The IO loop catches CommandFailed and Exception and only logs warnings, then continues.

**Why this is a blocker**
In disruption tests, some IO failures are expected, but “IO failed every time” should not silently pass as a healthy run. Otherwise we lose real coverage.

**How to fix (track success/failure + enforce minimal success)**
Add counters and require at least one successful IO iteration (or fail if failures exceed a threshold).

**Concrete example fix:**
Initialize: success = 0; failures = 0
On success: success += 1
On exception: failures += 1

After loop: `if success == 0: log.error("No successful IO during disruption window"); return 1`
Also log a summary:
`log.info(f"IO loop summary: success={success}, failures={failures}")`

tests/cephfs/cephfs_system/mds_nfs_node_failure_ops.py
```diff
@@ -33,11 +33,20 @@ def start_io_time(fs_util, client1, mounting_dir, timeout=300):
...
+                sudo=True,
+                cmd=f"rm -rf {mounting_dir}/run_ios_{iter}",
```
Manimaran-MM, line 53: Also make sure rm is handled in finally block

tests/cephfs/clients/test_selinux_relabel.py
```diff
@@ -139,7 +145,8 @@ def run(ceph_cluster, **kw):
             log.info(f"Creating {num_files} files in {test_dir}")
             client.exec_command(
                 sudo=True,
-                cmd=f"for i in $(seq 1 {num_files}); do touch {test_dir}/file_$i; done",
+                cmd=f"seq 1 {num_files} | xargs -n 500 -P 16 "
```
Manimaran-MM: **What’s there now**
seq ... | xargs -n 500 -P 16 sh -c '...'

**Why it’s minor**
May overload smaller CI nodes or be suboptimal on larger ones.

**How to fix (parameterize concurrency)**
Make P configurable:
parallel = config.get("file_create_parallelism", 16)
Then: ... xargs -n 500 -P {parallel} ...

tests/cephfs/cephfs_top/cephfs_top_dump.py
```diff
@@ -462,13 +469,18 @@ def run(ceph_cluster, **kw):
             fuse_mounting_dir_2,
             fuse_mounting_dir_3,
         ]:
+            if mount_dir:
+                fs_util.client_clean_up(
```
Manimaran-MM, line 473: **What’s there now**
Guards added; kernel/fuse cleanup separated.

**How to fix (reduce duplication)**
Build lists and iterate:
```
for d in [fuse_mounting_dir_1, fuse_mounting_dir_2, fuse_mounting_dir_3]: if d: client_clean_up(...)
if kernel_mounting_dir_1: client_clean_up(...) (already mostly done—just keep consistent patterns)
```

Manimaran-MM (changes requested):
Below code review by Co-pilot. Kindly review the same

Manimaran-MM:
Also can we merge multiple commit into single commit since the changes are related to only TFA

Manimaran-MM (approved):
LGTM

PR #5800, opened 2026-04-21
Fix to address --build released and upgrade scenario

obannak:
# Description

Issue: CDN image pulls used stage credentials when they should follow the build. 

Fix: Choose registry credentials by build type so released and staging builds use the correct registry.

> Sample test log:
> IBM build: https://149.81.216.83/job/qe-tentacle-released/25/
> RH build: https://149.81.216.83/job/qe-tentacle-released/26/

### DIFF
--- a/ceph/ceph_admin/bootstrap.py
+++ b/ceph/ceph_admin/bootstrap.py
@@ -32,7 +32,11 @@ __DEFAULT_SSH_PATH = "/etc/ceph/ceph.pub"
 
 
 def construct_registry(
-    cls, registry: str, json_file: bool = False, ibm_build: bool = False
+    cls,
+    registry: str,
+    json_file: bool = False,
+    ibm_build: bool = False,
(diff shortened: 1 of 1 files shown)


rakeshcn (changes requested):
Offline discussion on "Missing build_type Parameter in Function Calls"

obannak:
> Issue: CDN image pulls used stage credentials when they should follow the build.
> 
> Fix: Choose registry credentials by build type so released and staging builds use the correct registry.

> Thanks for the review—the attached doc was written against an earlier diff; the current PR already passes build_type into both construct_registry() call sites and documents it in the docstring, so the “missing build_type” / not-merge-ready finding no longer applies. The implementation uses the nested credentials.registry.<vendor>.<cdn|stage> keys (with legacy fallback), not the flat key names in that write-up.

rakeshcn (approved):
LGTM

PR #5801, opened 2026-04-21
remove the ssl spec in non-ssl test cases

sunilkumarn417:
Issue:
Error EINVAL: "ssl" field must be set to true
when "generate_cert" is set to true

Fix:
remove ssl spec which is not required as per test case


shivamdurgbuns (approved):
LGTM

psathyan (approved):

PR #5828, opened 2026-05-11
Support license acceptance during upgrade to IBM Ceph 9.1 builds

harshkumarRH:
To-be-added-shortly

### DIFF
--- a/ceph/ceph_admin/upgrade.py
+++ b/ceph/ceph_admin/upgrade.py
@@ -5,10 +5,14 @@ from json import JSONDecodeError, loads
 from time import sleep
 from typing import Dict
 
+from looseversion import LooseVersion
+from packaging.version import Version
+
 from ceph.ceph import CommandFailed
 from ceph.waiter import WaitUntil
 from utility.log import Log
 
+from ..utils import get_daemon_versions
 from .common import config_dict_to_string
 from .typing_ import OrchProtocol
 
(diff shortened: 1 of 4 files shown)


harshkumarRH:
Rebased to include https://github.com/red-hat-storage/cephci/pull/5830

ceph/utils.py
```diff
@@ -2020,3 +2025,30 @@ def enable_coredump(node):
...
+        ceph_versions = json.loads(out)
+    except json.JSONDecodeError as e:
+        log.exception(e)
+        raise
+
+    if ceph_versions.get(daemon_type):
+        daemon_ver_text = next(iter(ceph_versions[daemon_type]))
+        return extract_ceph_version(daemon_ver_text), extract_version(daemon_ver_text)
+
+    log.warning("No versions found for ", daemon_type)
```
rakeshcn:
```suggestion
    log.warning("No versions found for %s", daemon_type)
```

ceph/ceph_admin/upgrade.py
```diff
@@ -71,6 +74,27 @@ def start_upgrade(self: OrchProtocol, config: Dict):
...
+            and not skip_license
+        ):
+            LOG.debug("Ceph product: %s", config.get("product"))
+            LOG.debug("Target version: %s", config.get("release"))
+            mgr_version_common, mgr_version_downstream = get_daemon_versions(
+                node=self.installer, daemon_type="mgr"
+            )
+            LOG.debug("MGR common version: %s", mgr_version_common)
+            LOG.debug("MGR downstream version: %s", mgr_version_downstream)
+            if mgr_version_common >= "20.2.1" or mgr_version_downstream >= "9.9.1.0":
```
rakeshcn: Bob recommends
String comparison is unreliable for semantic versions
Should use `packaging.version.Version `instead

ceph/ceph_admin/upgrade.py
```diff
@@ -71,6 +74,27 @@ def start_upgrade(self: OrchProtocol, config: Dict):
         if args and "services" in args:
             cmd.append(f"--services {args['services']}")
 
+        skip_license = None
+        if args and "skip_license" in args or config.get("skip_license"):
```
rakeshcn: Bob suggests:
Operator precedence issue
Should use proper Boolean logic

ceph/utils.py
harshkumarRH: Thanks for catching this. Pushed the fix.

ceph/ceph_admin/upgrade.py
rakeshcn: if (args and "skip_license" in args) or config.get("skip_license"):
or
if ("skip_license" in args if args else False) or config.get("skip_license"):

to solve the precedence issue, either of the above can be used, which makes the truth table comparison clearer

harshkumarRH: Thanks, this has been addressed now

ceph/ceph_admin/upgrade.py
harshkumarRH: Applied comparison with `Version`

rakeshcn (approved):
LGTM

neha-gangadhar (approved):
lgtm

PR #5841, opened 2026-05-15
Enhance test_stripe_unit_pool_creation with varied stripe units and negative validation

pdhiran:
- Add precondition check for osd_pool_default_flag_ec_optimizations=true
- Add 5 fast-EC pools with explicit 4K-aligned stripe_unit values (4K, 8K, 32K, 64K, 128K) to verify non-default stripe units work
- Add 4 non-fast-EC pools with non-4K-aligned stripe_unit values (6112, 7168, 5120, 2048) created with --force to test that fast EC is correctly rejected for misaligned stripe units
- Add negative validation step: attempt to enable allow_ec_optimizations on ineligible pools, verify EINVAL error with correct message for non-4K-aligned pools, and confirm property stays False
- Update pool_configs tuple to include stripe_unit and force fields

### DIFF
--- a/suites/tentacle/rados/tier-3_rados_ec_partial_reads_writes.yaml
+++ b/suites/tentacle/rados/tier-3_rados_ec_partial_reads_writes.yaml
@@ -241,6 +241,16 @@ tests:
                 - mon_osd_down_out_subtree_limit
                 - host
 
+  # Test: Stripe unit and Fast EC default verification
+  - test:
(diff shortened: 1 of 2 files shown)


pdhiran:
Pass logs : http://magna002.ceph.redhat.com/ceph-qe-logs/cephci-run-GSTUBE/

harshkumarRH (approved):
Looks good

PR #5891, opened 2026-06-02
Fix ceph mon dump parsing failure on Ceph 20.x (tentacle)

kdhaduk81:
Ceph 20.x added new auth_* fields (auth_epoch, auth_service_cipher, auth_allowed_ciphers, auth_preferred_cipher) to 'ceph mon dump' output. These appear after the monitor entries, breaking the hardcoded [-3], [-2], [-1] indexing in get_clients() which assumed monitors were the last 3 lines.

Replace the fragile 'ceph mon dump | awk' parsing with get_mon_node_ips() that reads monitor IPs directly from the cluster config object, matching the approach already used in cephfs_utilsV1.py.

This fixes ~9 test failures in regression_suite_1 on RHEL 10 with Ceph 20.x and prevents breakage for any future 'ceph mon dump' format changes.

Pass Results: https://10.8.128.2/cephci-jenkins/ibm-cos//RH/9.1/rhel-10/regression/20.2.1-280/cephfs/69/logs/

### DIFF
--- a/tests/cephfs/cephfs_utils.py
+++ b/tests/cephfs/cephfs_utils.py
@@ -4,8 +4,6 @@ import re
 import string
 import time
 
-from looseversion import LooseVersion
-
 from ceph.ceph import CommandFailed
 from utility.log import Log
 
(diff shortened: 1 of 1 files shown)


Manimaran-MM (approved):
LGTM

PR #5892, opened 2026-06-02
Fix A signature was required, but no signature exists eroor in 10.2

aravindrrh:
# Description

Fixes the issue with Upgrades:

`Trying to pull registry.redhat.io/rhceph/rhceph-7-rhel9:latest...
Error: unable to copy from source docker://registry.redhat.io/rhceph/rhceph-7-rhel9:latest: Source image rejected: A signature was required, but no signature exists
 and code 125`


Manimaran-MM:
Closing after discussion with @aravindrrh

PR #5679, opened 2026-02-25
Fix cephadm bootstrap for signed & unsigned Podman images

obannak:
PR Description:

> Problem: cephadm bootstrap failed with Podman due to signature verification errors:

> Error: invalid policy in "/etc/containers/policy.json": Default policy is missing

Root Cause:

> The current /etc/containers/policy.json did not include a proper "default" policy, which is mandatory for Podman. Additionally, critical registries used in Ceph CI (registry.redhat.io, registry.stage.redhat.io, quay.io) were rejected due to missing or strict signature policies.

Solution:

> Added a "default": [{"type": "insecureAcceptAnything"}] policy to satisfy Podman’s requirement.

> Added "insecureAcceptAnything" for the following registries:

- > registry.redhat.io
- > registry.stage.redhat.io
- > quay.io

Impact:

> Fixes cephadm bootstrap and image pull failures for critical CI registries.
> No global insecure policies beyond the necessary default; all other registries remain unaffected.
> Verified successfully on test VMs.

Initially i added below content to the inventory and tried. This worked for images from registry.redhat.io (which are signed).

However, during upgrade testing, pulls from quay.io failed because Podman rejected the image based on the container signature policy. The reason was that /etc/containers/policy.json did not properly allow images from quay.io.

- registry.redhat.io → worked
- quay.io → failed (rejected by policy)
Log: https://149.81.216.83/view/Executor/job/tentacle-test-executor/851/

`- content: |
    {
        "default": [
            {
                "type": "reject"
            }
        ],
        "transports": {
            "docker": {
                "registry.redhat.io": [
                    {
                        "type": "insecureAcceptAnything"
                    }
                ]
            }
        }
    }
  path: /etc/containers/policy.json
  owner: root:root
  permissions: '0644'`

I have updated policy:
  - Adds a valid "default" policy (required by Podman).
  - Sets "default": [{"type": "insecureAcceptAnything"}] so other registries are not rejected.

Explicitly allows:

- registry.redhat.io
- registry.stage.redhat.io
- quay.io

Now both signed and unsigned images can be pulled successfully, and cephadm bootstrap and upgrade checks complete without policy errors.

Log: https://149.81.216.83/view/Executor/job/tentacle-test-executor/852/

below is the content added to inventory.
`
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports": {
        "docker": {
            "registry.redhat.io": [
                {
                    "type": "insecureAcceptAnything"
                }
            ],
[...]

PR #5895, opened 2026-06-03
fix(inventory): add podman container policy to fix cephadm bootstrap on eu-de RHEL 9.8 and 10.2

obannak:
`Summary`
Bootstrap on IBM eu-de lab VMs using RHEL 9.8 and RHEL 10.2 inventory can fail during cephadm bootstrap even when registry credentials are correct. Podman rejects pulls from registry.redhat.io because fresh nodes do not have container image signature trust configured:

Error: unable to copy from source docker://registry.redhat.io/rhceph/rhceph-7-rhel9:7.1-1:
Source image rejected: A signature was required, but no signature exists
This PR adds /etc/containers/policy.json via cloud-init in the 9.8 and 10.2 eu-de inventory files, matching the policy already used on 9.7 and other IBM eu-de images.

Note: please refer previously merged [PR](https://github.com/red-hat-storage/cephci/pull/5679) which was added to same issue with rhel-10.1 and 9.7

Test Log: https://149.81.216.83/job/tentacle-dmfg-upgrade-ci/61/

### DIFF
--- a/conf/inventory/ibm-eu-rhel-10.2.yaml
+++ b/conf/inventory/ibm-eu-rhel-10.2.yaml
@@ -85,6 +85,36 @@ instance:
         path: /etc/ssh/sshd_config.d/50-cephci.conf
         owner: root:root
         permissions: '0644'
+      - content: |
+          {
+              "default": [
+                  {
+                      "type": "insecureAcceptAnything"
+                  }
+              ],
+              "transports": {
+                  "docker": {
+                      "registry.redhat.io": [
+                          {
+                              "type": "insecureAcceptAnything"
+                          }
+                      ],
+                      "registry.stage.redhat.io": [
+                          {
+                              "type": "insecureAcceptAnything"
+                          }
+                      ],
+                      "quay.io": [
+                          {
+                              "type": "insecureAcceptAnything"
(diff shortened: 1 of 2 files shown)


rakeshcn (approved):
LGTM with the following constraints
 
Limited to test lab environment
⚠️ DOES disable signature verification - Images are not checked for GPG signatures
⚠️ Allows unsigned images - Any image from the registry is accepted without signature validation

psathyan:
IMHO, we shouldn't get this merged as it would prevent us for identifying any image missing signatures. I believe the issue is specific to upgrade and when we use the latest tag for Ceph 7.x

I am concerned with such merges being approved.

rakeshcn:
> IMHO, we shouldn't get this merged as it would prevent us for identifying any image missing signatures. I believe the issue is specific to upgrade and when we use the latest tag for Ceph 7.x
> 
> I am concerned with such merges being approved.

this is similar to https://github.com/red-hat-storage/cephci/pull/5679 , would you like to revert ?
https://github.com/red-hat-storage/cephci/pull/5679#issuecomment-3969989613

rakeshcn (approved):
Approve

PR #5901, opened 2026-06-05
interop: default BUILD_TYPE to rc for OpenStack certification runs

obannak:
The interop pipeline uses --skip-subscription, which is incompatible with --build released bootstrap on OpenStack. Use rc to install the same GA compose without enabling RH CDN tool repos via subscription-manager.


tintumathew10 (approved):
LGTM

PR #5903, opened 2026-06-08
Update interop script: remove --skip-subscription and set released as default

tintumathew10:
Update interop script: remove --skip-subscription and set released as default


obannak (approved):
LGTM

PR #5921, opened 2026-06-11
 Reverted --skip-subscription option and released tag from interop

tintumathew10:
Reverted --skip-subscription option and released tag from interop


obannak (approved):
LGTM

PR #5930, opened 2026-06-16
NFS Ganesha TLS wire encryption verification using fio I/O and tcpdump (CEPH-83629170)

SinghSamosa:
**Description**
This automation validates that NFS-over-TLS encrypts data on the network between the client and the Ganesha server. It uses a two-phase compare approach. 

In the first phase, a plain (non-TLS) Ganesha cluster is deployed, sustained fio I/O is run with a known buffer pattern, and traffic is captured with tcpdump. The test confirms that the fio pattern appears in cleartext on the wire, establishing that the capture and detection method works. 

In the second phase, a TLS-enabled Ganesha cluster and TLS export are deployed, the same fio workload is run under tcpdump, and the test confirms that the same pattern does not appear in the capture while TLS-framed traffic is present. Together, this proves the control baseline (plain NFS leaks the pattern) and that TLS actually protects payload data on the wire.

**Summary**

This change adds a new test operation tls_io_tcpdump_encryption to security.test_tls_feature.py and registers it in suites/tentacle/nfs/tier1-nfs-ganesha.yaml under Polarion ID CEPH-83629170. Supporting helpers were added in security_utils.py for deploying a plain NFS cluster without SSL, starting and stopping tcpdump capture on the client, running fio with a normalized hex buffer pattern, and analyzing pcaps on the client node.

The test deploys a plain NFS Ganesha cluster named cephfs-nfs-plain and a separate TLS cluster named cephfs-nfs-tls with inline certificates, TLSv1.3, and kTLS enabled. On the client, ktls-utils and tlshd are configured with the server CA for TLS mounts. Plain and TLS exports are created on CephFS, and NFS port 2049 traffic is captured with tcpdump during mount and fio I/O. The fio workload uses 100M file size, 4k block size, 300 second runtime for both write and read, and a buffer pattern of 0xcefacefe so that cleartext leakage can be detected by searching the raw pcap for those bytes.

Pcap verification runs on the client without transferring large capture files back to cephci. For the plain NFS control phase, the test passes when the fio pattern is found in the pcap, confirming unencrypted traffic. For the TLS phase, the test passes when the same pattern is not found in the pcap, indicating encrypted wire traffic. TLS record byte patterns are also counted as supporting evidence. The test performs full cleanup of mounts, exports, both NFS clusters, and CephFS subvolumes when complete.

**Test Steps**

The automation begins by installing tcpdump and fio on the NFS client. 
In **Phase 1**, it deploys a plain Ganesha cluster on the NFS node without TLS, verifies the service is running in ceph orch, creates a CephFS subvolume and plain export without xprtsec tls, and starts tcpdump on the client filtering traffic to the NFS server on port 2049. It then mounts the plain export on the client using NFSv4.2 without xprtsec tls, runs fio write for 300 seconds with the known buffer pattern, drops caches, runs fio read for 300 seconds, stops tcpdump, and unmounts. The plain pcap is analyzed and the test passes if the cleartext fio pattern appears in the capture. The plain cluster, export, and subvolume are then torn down before the TLS phase begins.

In **Phase 2**, a self-signed certificate is generated and a TLS-enabled Ganesha cluster is deployed with ssl enabled. The client is configured with tlshd and the server CA. A TLS export is created with xprtsec tls, tcpdump is started on the client, and the export is mounted with xprtsec tls so the TLS handshake is included in the capture. The same fio write and read workload runs for 300 seconds each. After stopping tcpdump and unmounting, the TLS pcap is analyzed. The test passes if the fio cleartext pattern is not found in the capture and TLS record traffic is present. Finally, all TLS resources are cleaned up including the cluster, exports, subvolumes, and mount paths.

**Pass Criteria**
The test passes when fio write and read complete successfully on both plain and TLS mounts, the plain NFS capture shows the fio buffer pattern in cleartext on the wire, and the TLS capture does not show that pattern while sustained I/O runs over the encrypted connection.

### DIFF
--- a/suites/tentacle/nfs/tier1-nfs-ganesha.yaml
+++ b/suites/tentacle/nfs/tier1-nfs-ganesha.yaml
@@ -515,3 +515,23 @@ tests:
         nfs_cluster_name: cephfs-nfs-tls
         operation: tls_logs_openssl_probe
         redeploy_cluster_min_tls12: false
+
+  - test:
+      name: TLS mount fio IO with tcpdump encryption verification
+      module: security.test_tls_feature.py
+      desc: >
+        Phase 1: plain NFS cluster/export, fio write/read, tcpdump- fio buffer
+        pattern must appear in cleartext on the wire. Phase 2: TLS cluster/export,
+        same fio pattern must not appear; TLS wire traffic verified encrypted.
+      abort-on-fail: false
+      polarion-id: CEPH-83629170
+      config:
+        nfs_version: 4.2
+        clients: 1
+        nfs_cluster_name: cephfs-nfs-tls
+        plain_nfs_cluster_name: cephfs-nfs-plain
+        operation: tls_io_tcpdump_encryption
+        fio_size: 100M
+        fio_bs: 4k
+        fio_runtime: 300
+        timeout: 7200
--- a/tests/nfs/security/security_utils.py
+++ b/tests/nfs/security/security_utils.py
@@ -1,4 +1,6 @@
 import json
+import shlex
+import time
 
 from ceph.ceph import CommandFailed
 from ceph.waiter import WaitUntil
@@ -16,6 +18,8 @@ TLS_TEST_MOUNT_CANDIDATES = [
     "/mnt/plain_tc2",
     "/mnt/tls_tc2",
     "/mnt/tls_tc1_0",
+    "/mnt/plain_tc4_0",
+    "/mnt/tls_tc4_0",
 ]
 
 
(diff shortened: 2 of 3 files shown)


SinghSamosa:
[manish-tls-enc12.zip](https://github.com/user-attachments/files/28997206/manish-tls-enc12.zip)
Logs uploaded

Manimaran-MM (approved):
LGTM..

Please fix DCO

aravindrrh (approved):
LGTM

Manimaran-MM:
@SinghSamosa Please squash commits to one

Manimaran-MM (approved):
LGTM

PR #5947, opened 2026-06-17
Adding sanity tests for cephadm agent

AmarnatReddy:
# Description
Adding sanity tests for cephadm agent

Logs : http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-G1L1FD/

<details>

</details>

### DIFF
--- a/tests/cephadm/cephadm_agent/test_agent_communication.py
+++ b/tests/cephadm/cephadm_agent/test_agent_communication.py
@@ -1,0 +1,342 @@
+"""Agent communication, MGR failover, and network partition tests."""
+
+import json
+import time
+
+from cephadm_agent.helpers import (
+    AGENT_HEALTH_WARNING,
+    DEFAULT_AGENT_DOWN_TIMEOUT,
+    DEFAULT_AGENT_REFRESH,
+    agent_service_name,
+    get_agent_daemons,
+    get_fsid,
+    get_node_for_host,
+    get_orch_ps,
+    log,
+    setup_run,
+    shell,
+    wait_for_agent_running,
+    wait_for_health_warning,
+)
+
+
+def run_communication_test(ceph_cluster, installer):
+    log.info("=== TEST: Communication - Block host → mgr ===")
+
+    agents = get_agent_daemons(installer)
+    assert len(agents) > 0, "No agent daemons found"
+    target = agents[0]
+    hostname = target["hostname"]
+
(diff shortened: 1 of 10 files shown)


tests/cephadm/cephadm_agent/test_agent_communication.py
```diff
@@ -0,0 +1,342 @@
...
+    DEFAULT_AGENT_DOWN_TIMEOUT,
+    DEFAULT_AGENT_REFRESH,
+    _agent_service_name,
+    _get_agent_daemons,
+    _get_fsid,
+    _get_node_for_host,
+    _get_orch_ps,
+    _shell,
+    _wait_for_agent_running,
+    _wait_for_health_warning,
```
psathyan: IMHO, methods starting with `_` are private methods and should only be referenced in that module. If it is required to be used by other modules, I would suggest not having the start with `_`.

psathyan (commented):
Thanks @AmarnatReddy for putting this together.

I wouldn't recommend having asserts in `def run(...):` method without a `try...catch` block. This kind of breaks the tests and execution without proper logging. Would it be possible to include them in `try...catch`?

Secondly, please avoid importing of methods starting with an `_` as they symbolize private methods.

tests/cephadm/cephadm_agent/test_agent_communication.py
AmarnatReddy: Done

AmarnatReddy:
> Thanks @AmarnatReddy for putting this together.
> 
> I wouldn't recommend having asserts in `def run(...):` method without a `try...catch` block. This kind of breaks the tests and execution without proper logging. Would it be possible to include them in `try...catch`?
> 
> Secondly, please avoid importing of methods starting with an `_` as they symbolize private methods.



> Thanks @AmarnatReddy for putting this together.
> 
> I wouldn't recommend having asserts in `def run(...):` method without a `try...catch` block. This kind of breaks the tests and execution without proper logging. Would it be possible to include them in `try...catch`?
> 
> Secondly, please avoid importing of methods starting with an `_` as they symbolize private methods.

Hi Pragadee,
The assert statements are not used directly inside the run() method. They are present in the individual sub-test functions, such as run_communication_test() and run_mgr_failover_test(). Each sub-test is invoked inside a try/except Exception block within run(), so any AssertionError should be caught and logged with the full traceback using log.error(..., exc_info=True).

psathyan (approved):

PR #5950, opened 2026-06-18
[RADOS] Enable monitoring stack and replace rc tag with GA

harshkumarRH:
Replaces RC tag with released/GA in RADOS test suites

Enables monitoring stack across the board


pdhiran (commented):
Looks good

pdhiran (approved):

PR #5953, opened 2026-06-22
NFS: TLS negative invalid certificate scenarios

SinghSamosa:
**Summary**
- Add `tls_negative_invalid_certs` operation to `test_tls_feature.py` for corrupted Ganesha server certs and invalid client trust anchor.
- Extend `security_utils.py` with cert corruption, mount-failure capture, and tlshd handshake failure checks.
- Wire the test into `suites/tentacle/nfs/tier1-nfs-ganesha.yaml`.

**What the test does**
1. Creates a TLS export on the existing TLS NFS cluster.
2. Confirms a valid TLS mount and small IO succeed as baseline.
3. **Server negative:** corrupts Ganesha key/cert/CA, redeploys NFS, and verifies TLS mount fails with tlshd handshake failure logged.
4. **Client negative:** installs a mismatched client trust anchor and verifies TLS mount fails with tlshd status captured.
5. Cleans up the mount point.

### DIFF
--- a/tests/nfs/security/security_utils.py
+++ b/tests/nfs/security/security_utils.py
@@ -480,12 +480,6 @@ def write_client_tls_ca_cert(client_node, ca_cert):
     client_node.exec_command(sudo=True, cmd="systemctl restart tlshd")
 
 
-def mount_path_is_active(client_node, mount_path):
-    """Return True if ``mount_path`` appears in the remote ``mount`` output."""
-    out, _ = client_node.exec_command(sudo=True, cmd="mount", check_ec=False)
-    return mount_path.rstrip("/") in str(out or "")
-
-
 def _write_remote_pem(client_node, file_path, pem_text):
     """Write PEM material to ``file_path`` on the remote client."""
     try:
--- a/tests/nfs/security/security_utils.py
+++ b/tests/nfs/security/security_utils.py
@@ -1,8 +1,10 @@
 import json
+import os
+import re
 import shlex
 import time
+from pathlib import Path
 
-from ceph.ceph import CommandFailed
 from ceph.waiter import WaitUntil
 from cli.ceph.ceph import Ceph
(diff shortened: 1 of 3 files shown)


SinghSamosa:
[manish-tls-neg8.zip](https://github.com/user-attachments/files/29233004/manish-tls-neg8.zip)
Logs can be accessed here.

tests/nfs/security/security_utils.py
```diff
@@ -279,14 +287,205 @@ def check_mount_fails(client_node, mount_cmd):
...
+        if line.startswith("-----") or not line.strip():
+            continue
+        if len(line) > 20:
+            mid = len(line) // 2
+            lines[idx] = line[:mid] + "XXX" + line[mid + 3 :]
+            break
+    return "\n".join(lines) + "\n"
+
+
+def read_client_tls_ca_cert(client_node):
```
Manimaran-MM: I see this function is unused. Can we remove if not used?

Manimaran-MM (commented):
Could you please review and move the funcitons to `tests/nfs/lib`, if you feel these are lib that can be consumed by other folks

tests/nfs/security/security_utils.py
SinghSamosa: Removed read_client_tls_ca_cert as it was unused.

SinghSamosa:
> Could you please review and move the funcitons to `tests/nfs/lib`, if you feel these are lib that can be consumed by other folks

Keeping negative-cert helpers in security_utils.py alongside existing TLS utilities used by the same test module. Happy to extract to tests/nfs/lib in a follow-up if we want a shared TLS lib for other NFS tests.

Manimaran-MM (dismissed):

aravindrrh (approved):
LGTM

PR #6009, opened 2026-07-13
fix: handle int sync_duration in validate_snapshot_sync_status

kdhaduk81:
Logs: http://10.64.24.74/logs/ceph-qe-logs//RH/9.1/rhel-10/regression/20.2.1-315/cephfs/99/logs/Test_Cephfs_Mirroring/


neha-gangadhar (approved):
lgtm

PR #6041, opened 2026-07-28
Remove period pull after realm pull in tentacle RGW MS suites [Extended regression only]

yuva-gayam:
BUg: https://ibm-ceph.atlassian.net/browse/IBMCEPH-16303 

Removed the secondary-side radosgw-admin period pull that runs immediately after realm pull in selected tentacle RGW multisite suites.

On 9.2, this extra period pull is terminating multisite setup with `(2) No such file or directory ` even though realm pull succeeds. The period is already pulled as part of realm pull, so the follow-up period pull is not necessary and is removed for now to unblock MS suite runs.

Multisite setup now continues from realm pull directly to zone create / period update and the remaining config steps.

pass logs: https://10.8.128.2/cephci-jenkins/ibm-cos//IBM/9.2/rhel-10/executor/20.2.2-124/rgw/2402/logs/tier_2_rgw_ms_archive/

### DIFF
--- a/suites/tentacle/rgw/migrate_default_single_to_multisite.yaml
+++ b/suites/tentacle/rgw/migrate_default_single_to_multisite.yaml
@@ -203,7 +203,6 @@ tests:
             commands:
               - "sleep 120"
               - "radosgw-admin realm pull --rgw-realm india --url http://{node_ip:ceph-pri#node6}:5000 --access-key a123 --secret s123 --default"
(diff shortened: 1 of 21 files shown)


viduship (dismissed):

ckulal (commented):
Looks good, however @yuva-gayam  just wanted to confirm,
removal of period pull will not cause any failures in older tentacle releases ryt?
i.e, 9.1/9.0

anrao19 (approved):
LGTM

ckulal (approved):

PR #6048, opened 2026-07-28
byok and cephfs fixes

Harishacharya-redhat:
what changed/added.

- Added small fix for byok tests (relogging to gklm if ceph command took more time, which logout the the gklm user )
- added comments for GRPC known failures.
- added fix for cephfs (8_1_features_suite.yaml ----- [test cephfs attributes - subvolume])

cephfs logs: https://10.8.128.2/cephci-jenkins/ibm-cos//IBM/9.2/rhel-10/executor/20.2.2-124/cephfs/2400/logs/8_1_features_suite/

byok logs: https://10.8.128.2/cephci-jenkins/ibm-cos//IBM/9.2/rhel-10/executor/20.2.2-124/nfs/2395/logs/tier1_nfs_ganesha_byok/

### DIFF
--- a/suites/tentacle/nfs/tier1-nfs-ganesha-byok.yaml
+++ b/suites/tentacle/nfs/tier1-nfs-ganesha-byok.yaml
@@ -238,6 +238,7 @@ tests:
            nfs_export: /export_byok
            nfs_instance_name: nfs_byok
            check_sighup: true
+           comments: bug - IBMCEPH-17211
 
      - test:
          name: Test SNI Default Secure Case
--- a/suites/tentacle/nfs/tier1-nfs-ganesha-grpc.yaml
+++ b/suites/tentacle/nfs/tier1-nfs-ganesha-grpc.yaml
@@ -134,6 +134,7 @@ tests:
         clients: 1
         nfs_version: 4.1
         port: 2049
+        comments: bug - IBMCEPH-16155
 
   # =============================================================================
   # Test Scenario 2: Verify gRPC Service Discovery
--- a/tests/cephfs/cephfs_case_sensitivity/case_sensitivity_subvolume.py
+++ b/tests/cephfs/cephfs_case_sensitivity/case_sensitivity_subvolume.py
@@ -243,6 +243,7 @@ def sv1_mount(mount_type="fuse"):
             smb_user_password="passwd",
             custom_dns="",
             clustering="default",
+            client_compat=None,
         )
 
         # Check smb cluster
--- a/tests/cephfs/lib/cephfs_common_lib.py
+++ b/tests/cephfs/lib/cephfs_common_lib.py
@@ -9,6 +9,7 @@ import os
 import random
 import re
 import secrets
+import shlex
 import string
 import time
 
@@ -283,6 +284,8 @@ class CephFSCommonUtils(FsUtils):
                     mnt_client.exec_command(
                         sudo=True,
                         cmd=cmd,
+                        check_ec=False,
+                        timeout=90,
                     )
                     if mnt_type == "nfs":
                         nfs_export = mount_details[sv_name][mnt_type]["nfs_export"]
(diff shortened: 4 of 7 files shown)


Harishacharya-redhat:
BOB review
tests/cephfs/lib/cephfs_common_lib.py — 3 issues
C1 — Unquoted mp in umount command Must Fix

Now: cmd=f"timeout 60 umount -l {mp}"
Risk: Mount paths with spaces split into multiple shell tokens; unmount silently skips the real target. A crafted path with $() gets subshell execution under sudo.
Fix: cmd=f"timeout 60 umount -l {shlex.quote(mp)}" + prefix guard before the loop.
C2 — rm -rf glob*/ still runs even if a umount silently failed Should Improve

Now: rm -rf {mount_path_prefix}*/ runs unconditionally after the per-mount umount loop; each umount has check_ec=False so failures are invisible.
Risk: A still-active or dead FUSE mount causes rm to hang until the 120s timeout fires; the entire cleanup path returns 1 with no indication of which mount caused the stall.
Fix: Re-query mounted paths after the loop; skip rm and warn if any are still mounted.
C3 — No aggregate log after unmount loop Should Improve

Now: log.info("Successfully unmounted on …") was removed and not replaced.
Risk: No way in CI logs to tell whether 0 or 10 mounts were attempted/processed.
Fix: Add log.info("Unmount pass complete: attempted %d mount(s) on %s", len(mount_points), …) after the loop.
tests/cephfs/lib/fscrypt_utils.py — 4 issues
C4 — except BaseException swallows KeyboardInterrupt/SystemExit Must Fix

Now: Two new except BaseException as ex: blocks inside name_content_read().
Risk: CI SIGINT abort is caught, logs a spurious "cannot list files" error, returns 1 — job appears as a test failure instead of a clean abort. Triggers unnecessary retry runs.
Fix: Replace both with except Exception as ex:.
C5 — except (CommandFailed, Exception) is redundant Should Improve

Now: except (CommandFailed, Exception) as ex: in get_file_list.
Risk: CommandFailed inherits from Exception; listing both is dead code that misleads readers and may leave an unused import (flake8 F401).
Fix: except Exception as ex: only; remove the CommandFailed import if unused elsewhere.
C6 — Case-sensitive error-string match in get_file_list Should Improve

Now: if "No such file" in combined or "Not a directory" in combined:
Risk: Lowercase or translated locale messages bypass the guard; code continues with empty/garbage output.
Fix: combined = f"…\n…".lower() then if "no such file" in combined or "not a directory" in combined:.
C7 — Missing KeyError guard for unknown op name in ops_func Must Fix

Now: test_status = ops_func[file_op]() with no key existence check.
Risk: A YAML typo or new op name not yet in ops_func raises an unhandled KeyError — no failed_ops recorded, raw traceback in CI, cleanup may not run.
Fix: Add if file_op not in ops_func: log.error(…); failed_ops.append(file_op); continue before the call.
tests/nfs/byok/test_byok_single_multi_restart.py — 1 issue
C8 — gklm_api_call() return value iterated directly; None → TypeError in cleanup Must Fix

Now: for x in gklm_api_call(gklm_rest_client, gklm_rest_client.certificates.list_system_certificates): with no null guard.
Risk: If the API returns None (empty response, post-retry failure), TypeError: 'NoneType' object is not iterable fires inside the finally block, gklm_cert_alias is never refreshed, clean_up_gklm runs with a stale alias → GKLM certs/keys leak across CI runs.
Fix: sys_certs = gklm_api_call(…) or [] then for x in sys_certs:.
tests/nfs/security/test_tls_feature.py — 1 issue
C9 — plain_nfs_name, nfs_node, client_node may be unbound in finally Must Fix

Now: clusters_to_parse.append(plain_nfs_name) and nfs_log_parser(client=client_node, nfs_node=nfs_node, …) with no initialisation guard.
Risk: If an exception fires before these variables are assigned, UnboundLocalError inside finally replaces the original exception — the real failure reason is permanently lost from CI logs.
Fix: Initialise plain_nfs_name = nfs_node = client_node = None at the top of run(); guard with if plain_nfs_name is not None and if nfs_node is None or client_node is None: continue.
tests/nfs/byok/byok_tools.py — 1 issue
C10 — Two unconditional time.sleep(30) magic-number waits Must Fix

Now: time.sleep(30) added to both create_nfs_instance_for_byok and create_multiple_nfs_instance_for_byok.
Risk: 30s of unconditional wait added per call, not tied to any observable condition. In a 3-instance BYOK test = 90s wasted. On slow runners the 30s may still not be enough, giving false confidence without actual readiness verification.
Fix: Replace with a poll loop that checks list_certificates() until the new alias appears (with _GKLM_CERT_TIMEOUT = 60 guard). At minimum, extract to a named constant _GKLM_CERT_SETTLE_SEC = 30 with a rationale comment.

Harishacharya-redhat:
Latest results


Screenshot 2026-07-30 at 2 03 23 PM

Manimaran-MM (approved):
LGTM

PR #6060, opened 2026-07-30
adding check_crash_status and log_cluster_health, in nfs upgrade tests.

Harishacharya-redhat:
Enhancing the upgrade tests to have health and crash checks for nfs upgrades.

logs: http://magna002.ceph.redhat.com/ceph-qe-logs/hacharya/upgrades/checks/nfs/tier1_nfs_ganesha_upgrade_8x_to_9x_latest/

### DIFF
--- a/tests/nfs/nfs_operations.py
+++ b/tests/nfs/nfs_operations.py
@@ -1979,3 +1979,67 @@ def get_ceph_version(client, prefix_cephadm=False):
         raise RuntimeError("Failed to get ceph version from cluster")
     ceph_version_installed = match.group(1)
     return ceph_version_installed
+
+
+def init_cluster_health_check(ceph_cluster, config=None):
+    """
+    Create RadosOrchestrator and capture start timestamp for crash checks.
+
+    Returns:
+        tuple: (rados_obj, start_time) or (None, None) on failure.
+    """
+    try:
+        from ceph.ceph_admin import CephAdmin
+        from ceph.rados.core_workflows import RadosOrchestrator
+        from ceph.rados.utils import get_cluster_timestamp
+
+        # Only pass keys CephAdmin actually needs; NFS suite config must not
+        # be spread into CephAdmin (would poison / shadow attributes).
+        cephadm_config = {
+            k: config[k] for k in ("build", "rhbuild") if config and k in config
+        }
+        cephadm = CephAdmin(cluster=ceph_cluster, **cephadm_config)
+        rados_obj = RadosOrchestrator(node=cephadm)
+        start_time = get_cluster_timestamp(rados_obj.node)
+        log.debug(f"Crash-check window started at {start_time}")
+        return rados_obj, start_time
+    except Exception as e:
+        log.warning(f"Failed to initialize cluster health check: {e}")
+        return None, None
+
+
+def log_cluster_health_and_check_crashes(rados_obj, start_time, check_nfs=True):
+    """
+    Log cluster health and check for crashes since start_time.
+
+    Returns:
+        True if a crash was detected, False otherwise.
+    """
+    from ceph.rados.utils import get_cluster_timestamp
+
+    if not rados_obj or not start_time:
+        log.warning("Skipping health/crash check: rados_obj or start_time missing")
+        return False
+
+    # Best-effort health dump — must not skip the crash check on failure.
+    try:
+        rados_obj.log_cluster_health()
+    except Exception as e:
+        log.error(f"log_cluster_health() failed: {e}")
+
+    try:
+        end_time = get_cluster_timestamp(rados_obj.node)
+        log.debug(
+            f"Crash-check window completed. Start time: {start_time}, "
+            f"End time: {end_time}"
+        )
+        if rados_obj.check_crash_status(
+            start_time=start_time,
+            end_time=end_time,
+            check_nfs=check_nfs,
+        ):
+            log.error("Crash detected during test window")
+            return True
+    except Exception as e:
+        log.error(f"Crash status check failed: {e}")
+    return False
--- a/tests/nfs/nfs_verify_multiple_parallel_io_and_lookups.py
+++ b/tests/nfs/nfs_verify_multiple_parallel_io_and_lookups.py
@@ -1,7 +1,12 @@
 import time
 from threading import Thread
 
-from nfs_operations import cleanup_cluster, setup_nfs_cluster
+from nfs_operations import (
+    cleanup_cluster,
+    init_cluster_health_check,
+    log_cluster_health_and_check_crashes,
+    setup_nfs_cluster,
+)
 
 from cli.exceptions import ConfigError, OperationFailedError
 from cli.io.io import linux_untar
@@ -40,6 +45,7 @@ def run(ceph_cluster, **kw):
     nfs_mount = "/mnt/nfs"
     fs = "cephfs"
     nfs_server_name = nfs_node.hostname
+    rados_obj, start_time = init_cluster_health_check(ceph_cluster, config)
 
     try:
         if setup:
@@ -130,6 +136,7 @@ def run(ceph_cluster, **kw):
         log.error(f"Failed to validate multiple ios and lookups : {e}")
         return 1
     finally:
+        crash_detected = log_cluster_health_and_check_crashes(rados_obj, start_time)
         if cleanup:
             log.info("Cleaning up")
             # Wait for NFS operations to stabilize before cleanup
@@ -139,3 +146,5 @@ def run(ceph_cluster, **kw):
                 clients, nfs_mount, nfs_name, nfs_export, nfs_nodes=nfs_nodes[0]
             )
             log.info("Cleaning up successful")
+        if crash_detected:
+            return 1
--- a/tests/nfs/nfs_verify_parallel_rm_write_lookup.py
+++ b/tests/nfs/nfs_verify_parallel_rm_write_lookup.py
@@ -1,7 +1,12 @@
 import time
 from threading import Thread
 
-from nfs_operations import cleanup_cluster, setup_nfs_cluster
+from nfs_operations import (
+    cleanup_cluster,
+    init_cluster_health_check,
+    log_cluster_health_and_check_crashes,
+    setup_nfs_cluster,
+)
 
 from ceph.ceph import CommandFailed
 from cli.exceptions import ConfigError, OperationFailedError
@@ -63,6 +68,7 @@ def run(ceph_cluster, **kw):
     fs = "cephfs"
     nfs_server_name = nfs_node.hostname
     file_count = 100
+    rados_obj, start_time = init_cluster_health_check(ceph_cluster, config)
     try:
         # Setup nfs cluster
         setup_nfs_cluster(
@@ -105,11 +111,11 @@ def run(ceph_cluster, **kw):
 
         # Perform rm while file creation and lookup is in progress
         max_total_time = 300
-        start_time = time.time()
+        wall_start = time.time()
         iteration = 0
         log.info("Performing rm while file creation and lookup in progress")
         while operations[0].is_alive() or operations[1].is_alive():
-            if time.time() - start_time > max_total_time:
+            if time.time() - wall_start > max_total_time:
                 log.error(f"Operations exceeded {max_total_time}s timeout")
                 raise OperationFailedError("Parallel operations timed out")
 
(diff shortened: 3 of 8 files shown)


Harishacharya-redhat:
PR #6060 — Inline Review Comments
tests/nfs/nfs_operations.py — init_cluster_health_check passes entire config dict as **kwargs to CephAdmin
What's there now

cephadm = CephAdmin(cluster=ceph_cluster, **(config or {}))

Why this is a blocker / risk

config in every test file is a raw YAML-derived dict containing NFS-specific keys like port, nfs_version, clients, operation, enable_rdma, rdma_port, ha, vip, sudo, file_count, dd_command_size_in_M, exports_number, chown_cephuser, etc. None of these are CephAdmin.__init__ parameters.
CephAdmin.__init__(self, cluster, **config) stores the entire dict. Stray unknown keys can silently poison the CephAdmin state, cause attribute collisions, or cause downstream cephadm shell / bootstrap operations to fail in confusing ways.
Any test that passes a key that happens to shadow an internal attribute of CephAdmin (e.g. ha, version) will silently override it — no error, just wrong behaviour in unrelated code paths.
How to fix (exact replacement)

# pass only the keys CephAdmin actually needs (none beyond cluster for health checks)
cephadm = CephAdmin(cluster=ceph_cluster)

If specific config keys are intentionally needed by CephAdmin, list them explicitly:

cephadm_config = {k: config[k] for k in ("build", "rhbuild") if k in (config or {})}
cephadm = CephAdmin(cluster=ceph_cluster, **cephadm_config)

Severity

Must Fix
tests/nfs/nfs_operations.py — init_cluster_health_check has no exception handling; an init failure crashes the entire test with an unhandled traceback
What's there now

def init_cluster_health_check(ceph_cluster, config=None):
    from ceph.ceph_admin import CephAdmin
    from ceph.rados.core_workflows import RadosOrchestrator
    from ceph.rados.utils import get_cluster_timestamp

    cephadm = CephAdmin(cluster=ceph_cluster, **(config or {}))
    rados_obj = RadosOrchestrator(node=cephadm)
    start_time = get_cluster_timestamp(rados_obj.node)
    log.debug("Crash-check window started at %s", start_time)
    return rados_obj, start_time

Why this is a blocker / risk

This function is called before the try block in every test (rados_obj, start_time = init_cluster_health_check(...) sits outside try:). If CephAdmin, RadosOrchestrator, or get_cluster_timestamp raises any exception, the test function exits with an unhandled traceback — bypassing the entire finally cleanup block in that test. This means NFS clusters, exports, and mounts are left mounted and leaked across test runs.
CI would report a Python traceback rather than a clean return 1, breaking result parsing.
How to fix
Wrap the call site in all tests, or — better — make the function return safe fallback values on failure:

def init_cluster_health_check(ceph_cluster, config=None):
    try:
        from ceph.ceph_admin import CephAdmin
        from ceph.rados.core_workflows import RadosOrchestrator
        from ceph.rados.utils import get_cluster_timestamp

        cephadm = CephAdmin(cluster=ceph_cluster)
        rados_obj = RadosOrchestrator(node=cephadm)
        start_time = get_cluster_timestamp(rados_obj.node)
        log.debug("Crash-check window started at %s", start_time)
        return rados_obj, start_time
    except Exception as e:
        log.warning("Failed to initialize cluster health check: %s", e)
        return None, None

The existing if not rados_obj or not start_time guard in log_cluster_health_and_check_crashes already handles the None, None case safely.

Severity

Must Fix
tests/nfs/nfs_verify_parallel_rm_write_lookup.py — start_time variable is silently overwritten, breaking the crash-detection window
What's there now

# line 71 — crash-check start time
rados_obj, start_time = init_cluster_health_check(ceph_cluster, config)
...
# line 114 — overwrites start_time with a wall-clock float
start_time = time.time()
...
# line 148 — passes the float, not the cluster timestamp, to crash checker
crash_detected = log_cluster_health_and_check_crashes(rados_obj, start_time)

Why this is a blocker / risk

start_time set at line 71 is a cluster-side timestamp string used by check_crash_status to query ceph crash ls. At line 114 it is clobbered by time.time(), a float epoch value.
log_cluster_health_and_check_crashes → check_crash_status(start_time=<float>) will either crash with a type error or silently pass a nonsensical value to the crash query, meaning real crashes in the test window will be missed entirely.
This makes the entire crash-detection feature non-functional for this test file.
How to fix
Rename the wall-clock variable:

# line 114 — use a distinct name
wall_start = time.time()
...
# line 118
if time.time() - wall_start > max_total_time:

Keep start_time (the cluster timestamp) untouched.

Severity

Must Fix
tests/nfs/nfs_verify_stress.py — Double cleanup_cluster call on the error path; resources cleaned up twice
What's there now

    except Exception as e:
        log.error(f"Failed to setup nfs cluster {e}")
        cleanup_cluster(clients, nfs_mount, nfs_name, nfs_export)       # ← first call
        return 1
    finally:
        crash_detected = log_cluster_health_and_check_crashes(rados_obj, start_time)
        cleanup_cluster(clients, nfs_mount, nfs_name, nfs_export, nfs_nodes=servers)  # ← second call
        if crash_detected:
            return 1

Why this is a blocker / risk

When the except branch is taken, cleanup_cluster runs twice — once in except (without nfs_nodes) and again in finally (with nfs_nodes=servers). The second call will attempt to delete/unmount already-deleted resources and may log spurious errors or raise, masking the real failure.
Also note the two calls use different arguments (nfs_nodes omitted in except), so partial teardown state differs between paths.
How to fix
Remove the duplicate call from the except block and rely solely on finally:

    except Exception as e:
        log.error(f"Failed to setup nfs cluster {e}")
        return 1
    finally:
        crash_detected = log_cluster_health_and_check_crashes(rados_obj, start_time)
        cleanup_cluster(clients, nfs_mount, nfs_name, nfs_export, nfs_nodes=servers)
        if crash_detected:
            return 1

Severity

Must Fix
tests/nfs/test_nfs_io_operations_during_upgrade.py — return 1 inside finally silently suppresses the OperationFailedError raised in except
What's there now

    except Exception as e:
        log.error(f"An error occurred during NFS IO operations: {e}")
        raise OperationFailedError(f"NFS IO operations failed: {e}")
    finally:
        if log_cluster_health_and_check_crashes(rados_obj, start_time):
            return 1

Why this is a blocker / risk

In Python, a return statement inside a finally block swallows any active exception. If the except block raises OperationFailedError and a crash is simultaneously detected, the finally block's return 1 silently discards OperationFailedError. The CI pipeline receives return 1 with no exception traceback, losing the diagnostic information about what actually failed.
This also affects the success path: if the test passes (return 0) and a crash is detected, finally: return 1 correctly overrides — but the mechanism of suppressing exceptions here is fragile and misleading.
How to fix
Store the crash result and raise or return after the finally completes, or capture the exception explicitly:

    except Exception as e:
        log.error(f"An error occurred during NFS IO operations: {e}")
        raise OperationFailedError(f"NFS IO operations failed: {e}")
    finally:
        crash_detected = log_cluster_health_and_check_crashes(rados_obj, start_time)
        if crash_detected:
            log.error("Crash detected; overriding exit status to 1")
            # Note: a return here suppresses any in-flight exception.
            # Prefer structuring this at the call-site or using a flag.

The cleanest pattern used correctly in other files in this PR is:

    except Exception as e:
        log.error(f"An error occurred during NFS IO operations: {e}")
        return 1
    finally:
        crash_detected = log_cluster_health_and_check_crashes(rados_obj, start_time)
        if crash_detected:
            return 1

(Replace raise with return 1 to remain consistent with how all other test files in this PR signal failure. The framework reads the integer return code, not the exception.)

Severity

Must Fix
tests/nfs/test_nfs_multiple_operations_for_upgrade.py — Crash detected during before_upgrade phase silently returns 1 and skips after_upgrade cleanup, leaking NFS cluster state
What's there now

    finally:
        crash_detected = log_cluster_health_and_check_crashes(rados_obj, start_time)
        if operation == "after_upgrade":
            # Cleanup NFS Cluster
            cleanup_cluster(
                clients, nfs_mount, nfs_name, nfs_export, nfs_nodes=nfs_node
            )
        if crash_detected:
            return 1

Why this is a blocker / risk

When operation == "before_upgrade" and a crash is detected, return 1 fires without any cleanup. The NFS cluster created in before_upgrade is intentionally left alive for after_upgrade, but if the test is aborted here, subsequent pipeline runs may find a stale NFS cluster and fail with cluster already exists errors.
Additionally, when operation == "after_upgrade" and crash_detected is True, the cleanup runs (correct), but then return 1 fires — this is actually fine for after_upgrade, but the code is ambiguous about intent.
How to fix
Add a log statement to make the intentional no-cleanup explicit, and document the assumption:

    finally:
        crash_detected = log_cluster_health_and_check_crashes(rados_obj, start_time)
        if operation == "after_upgrade":
            cleanup_cluster(
                clients, nfs_mount, nfs_name, nfs_export, nfs_nodes=nfs_node
            )
        elif crash_detected:
            log.warning(
                "Crash detected during 'before_upgrade' phase. "
                "NFS cluster intentionally left alive for after_upgrade step. "
                "Manual cleanup may be required if the upgrade pipeline is aborted."
            )
        if crash_detected:
            return 1

Severity

Should Improve
tests/nfs/nfs_operations.py — log_cluster_health_and_check_crashes silently swallows all exceptions; a broken log_cluster_health hides real cluster problems
What's there now

    try:
        rados_obj.log_cluster_health()
        end_time = get_cluster_timestamp(rados_obj.node)
        ...
        if rados_obj.check_crash_status(...):
            log.error("Crash detected during test window")
            return True
    except Exception as e:
        log.warning("Health/crash check failed: %s", e)
    return False

Why this is a blocker / risk

If log_cluster_health() raises (e.g., connectivity failure, ceph health returns non-zero), the entire block is skipped including check_crash_status. The function returns False — "no crash detected" — even though the cluster may be completely unreachable. This produces a false-negative: the test is reported as passing despite potential crashes.
log.warning at WARNING level is likely to be filtered out in CI log aggregation. A health check failure should be log.error.
How to fix
Separate the health logging (best-effort) from the crash check (must succeed):

    try:
        rados_obj.log_cluster_health()
    except Exception as e:
        log.error("log_cluster_health() failed: %s", e)

    try:
        end_time = get_cluster_timestamp(rados_obj.node)
        log.debug(
            "Crash-check window completed. Start time: %s, End time: %s",
            start_time,
            end_time,
        )
        if rados_obj.check_crash_status(
            start_time=start_time,
            end_time=end_time,
            check_nfs=check_nfs,
        ):
            log.error("Crash detected during test window")
            return True
    except Exception as e:
        log.error("Crash status check failed: %s", e)
    return False

Severity

Should Improve
tests/nfs/nfs_operations.py — f-string mixed with %-style logging in the same module (PEP-8 / project style inconsistency)
What's there now

# new functions use %-style (correct for logging)
log.debug("Crash-check window started at %s", start_time)
log.warning("Health/crash check failed: %s", e)

# existing and surrounding code in all 7 test files uses f-strings in log calls
log.error(f"Failed to validate multiple ios and lookups : {e}")
log.error(f"Failed to setup nfs cluster {e}")

Why this is a blocker / risk

No direct runtime failure, but the project's established convention (used uniformly across all test files in this PR) is f-strings in log calls. The two new utility functions switch to %-style formatting without justification, creating an inconsistency that will confuse future contributors.
The %-style lazy evaluation advantage only matters when the logger level would skip the message — for log.error / log.warning that always emit in CI, it provides no benefit.
How to fix
Align with project convention and use f-strings throughout the two new functions:

log.debug(f"Crash-check window started at {start_time}")
log.warning(f"Skipping health/crash check: rados_obj or start_time missing")
log.debug(f"Crash-check window completed. Start time: {start_time}, End time: {end_time}")
log.error(f"Crash detected during test window")
log.warning(f"Health/crash check failed: {e}")

Severity

Should Improve
tests/nfs/test_file_ops_copy.py — try block has no except clause; exceptions propagate uncaught past the finally block
What's there now

    try:
        # Setup nfs cluster
        setup_nfs_cluster(...)
        ...
        return 0

    finally:
        crash_detected = log_cluster_health_and_check_crashes(rados_obj, start_time)
        log.info("Cleaning up")
        cleanup_cluster(...)
        log.info("Cleaning up successful")
        if crash_detected:
            return 1

Why this is a blocker / risk

There is no except block. Any OperationFailedError or other exception raised inside try will propagate through finally (cleanup still runs — that part is fine), but the exception will be unhandled at the caller. The CI framework expects run() to return an integer (0 or 1); an unhandled exception may cause the test to be marked as errored rather than failed, breaking result tracking.
This pattern was present before this PR, but the PR introduces init_cluster_health_check before try:, so a failure there also bypasses finally. (Covered by the init-exception comment above.)
How to fix
Add an except block consistent with all other test files in this PR:

    except (ConfigError, OperationFailedError) as e:
        log.error(f"Failed to complete file copy operations: {e}")
        return 1
    except Exception as e:
        log.error(f"Unexpected error during file copy test: {e}")
        return 1
    finally:
        crash_detected = log_cluster_health_and_check_crashes(rados_obj, start_time)
        log.info("Cleaning up")
        cleanup_cluster(clients, nfs_mount, nfs_name, nfs_export, nfs_nodes=nfs_nodes[0])
        log.info("Cleaning up successful")
        if crash_detected:
            return 1

Severity

Should Improve

Harishacharya-redhat:
BOB review addressed.
Screenshot 2026-07-30 at 6 16 53 PM

aravindrrh (dismissed):
LGTM

Manimaran-MM (approved):

PR #6082, opened 2026-08-10
adjusting grace period during nfs upgrade

Harishacharya-redhat:
fix for upgrade issue, adjusting grace period + added known bugs in the comments for cephfs

initially in the automation. we did mount retry multiple times with longer pulling time, So it was working good.
Now we reduce the retry time, to fix indefinite mounting time if something go down in CI, So instead of blinding retrying every time for every tests. I have introduced grace period sleep time.

logs : https://10.8.128.2/cephci-jenkins/ibm-cos//IBM/9.2/rhel-9/executor/20.2.2-214/nfs/2474/logs/tier1_nfs_ganesha_upgrade_8x_to_9x_latest/

updated logs: after poll logic : https://10.8.128.2/cephci-jenkins/ibm-cos//IBM/9.2/rhel-9/executor/20.2.2-220/nfs/2514/logs/
image

PR #6083, opened 2026-08-10
Fix upgrades cephfs

Harishacharya-redhat:
THis PR fixes

CG Quiesce (crash_check false fail)
Suite: suites/tentacle/cephfs/tier-1_cephfs_cg_quiesce_systemic.yaml
Conf: conf/tentacle/cephfs/tier-1_cephfs_cg_quiesce.yaml
Should fix: cephfs-crash-check
old Log: http://10.64.24.74/logs/ceph-qe-logs/IBM/9.2/rhel-10/extended_regression/20.2.2-214/cephfs/39/logs/Test_Cephfs_CG_Quiesce/ flag qe-test_coverage “+” and link Polarion test
</details>
new logs: https://10.8.128.2/cephci-jenkins/ibm-cos//IBM/9.1/rhel-10/executor/20.2.1-367/cephfs/2477/logs/tier_1_cephfs_cg_quiesce_systemic/
image

Crash check 
suites/tentacle/cephfs/tier-1_cephfs_fscrypt.yaml + mirroring
old logs : http://10.64.24.74/logs/ceph-qe-logs/IBM/9.2/rhel-10/extended_regression/20.2.2-214/cephfs/39/logs/
new logs: https://10.8.128.2/cephci-jenkins/ibm-cos//IBM/9.1/rhel-10/executor/20.2.1-367/cephfs/2478/logs/
image

### DIFF
--- a/suites/squid/cephfs/tier-0_cephfs_upgrade_6x_to_8x.yaml
+++ b/suites/squid/cephfs/tier-0_cephfs_upgrade_6x_to_8x.yaml
@@ -140,6 +140,8 @@ tests:
   - test:
       name: Upgrade along with IOs
       module: test_parallel.py
+      config:
+        max_time: 10800
       parallel:
         - test:
             abort-on-fail: false
@@ -160,6 +162,7 @@ tests:
               service: upgrade
               base_cmd_args:
                 verbose: true
+              upgrade_timeout: 10800
               benchmark:
                 type: rados # future-use
                 pool_per_client: true
@@ -173,6 +176,9 @@ tests:
             module: cephfs_upgrade.cephfs_mds_failover.py
             name: "Fail MDS while Upgrade is in Progress"
             polarion-id: CEPH-83575628
+            config:
+              min_active_mds: 2
+              mds_recovery_timeout: 1200
       desc: Running upgrade, mds Failure and i/o's parallelly
       abort-on-fail: true
   - test:
--- a/suites/squid/cephfs/tier-0_cephfs_upgrade_8x_latest_to_81x_latest.yaml
+++ b/suites/squid/cephfs/tier-0_cephfs_upgrade_8x_latest_to_81x_latest.yaml
@@ -29,7 +29,7 @@ tests:
                 mon-ip: node1
                 orphan-initial-daemons: true
                 registry-url: registry.redhat.io
-                skip-monitoring-stack: false
+                skip-monitoring-stack: true
               base_cmd_args:
                 verbose: true
               command: bootstrap
@@ -140,6 +140,8 @@ tests:
   - test:
       name: Upgrade along with IOs
       module: test_parallel.py
+      config:
+        max_time: 10800
       parallel:
         - test:
             abort-on-fail: false
@@ -161,6 +163,7 @@ tests:
               service: upgrade
               base_cmd_args:
                 verbose: true
+              upgrade_timeout: 10800
               benchmark:
                 type: rados # future-use
                 pool_per_client: true
@@ -174,6 +177,9 @@ tests:
             module: cephfs_upgrade.cephfs_mds_failover.py
             name: "Fail MDS while Upgrade is in Progress"
             polarion-id: CEPH-83575628
+            config:
+              min_active_mds: 2
+              mds_recovery_timeout: 1200
       desc: Running upgrade, mds Failure and i/o's parallelly
       abort-on-fail: true
   - test:
--- a/suites/tentacle/cephfs/tier-0_cephfs_upgrade_7x_to_9x.yaml
+++ b/suites/tentacle/cephfs/tier-0_cephfs_upgrade_7x_to_9x.yaml
@@ -148,6 +148,8 @@ tests:
   - test:
       name: Upgrade along with IOs
       module: test_parallel.py
+      config:
+        max_time: 10800
       parallel:
         - test:
             abort-on-fail: false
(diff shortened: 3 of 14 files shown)


Harishacharya-redhat:
bob- review 
Summary by severity
🔴 Must Fix (4)
1. tests/cephfs/cephfs_mirroring/snapdiff_perf_improvements_across_releases.py — return inside finally silently swallows the failure exit code

The two new bare return statements added in the finally block:

if not config.get("cleanup", True):
    return   # ← overwrites return 1 from except
if not (source_clients and ...):
    return   # ← same problem

In Python, any return inside a finally block unconditionally overrides any return value already in flight from try/except — including return 1. A test that fails setup will enter the except, execute return 1, but then the finally block fires, hits source_clients is None, and return overrides it → CI receives exit code 0 / None and marks the job PASSED. Verified with a reproducer. Fix: use if/elif/else and let the block fall through naturally — never use bare return inside finally.

2. tests/cephfs/cephfs_mirroring/snapdiff_perf_improvements_across_releases.py — fs_util_ceph1.wait_for_nfs_process() doesn't exist

if not fs_util_ceph1.wait_for_nfs_process(
    source_clients[0], nfs_name, timeout=60, desired_state="running",
):

There is no wait_for_nfs_process method on FsUtils (confirmed by grepping the main branch). This raises AttributeError at runtime and kills every CI run of this test. Fix: either add the method in a prior reviewed commit, or replace with a retry loop around the existing get_ganesha_pid helper already in the file.

3. tests/cephfs/cephfs_mirroring/snapdiff_perf_improvements_across_releases.py — os.mkdir(log_dir) runs at module import time

log_dir = f"{log_base_dir}/SnapDiff_Results/"
os.mkdir(log_dir)   # module-level, not inside run()

Two parallel workers importing this module race on os.mkdir → second raises FileExistsError and aborts import. The new nfs_debug_logs(… log_dir …) call added in this PR creates a second dependency on this. Fix: remove the os.mkdir at module level; call os.makedirs(log_dir, exist_ok=True) at the top of run().

4. suites/tentacle/cephfs/tier-1_cephfs_cg_quiesce_systemic.yaml — comment: → comments: is an undocumented schema change

- comment: "IBMCEPH-17181"
+ comments: IBMCEPH-17181 in 9.2

If the suite executor only recognises comment (singular), the test annotated for bug IBMCEPH-17181 will no longer be skipped — it will run unconditionally, removing a deliberate CI gate with no warning. The PR does not update the schema parser or add a test. Other files in the repo use comments: (plural) so either is possible — but the intent must be verified and the parser updated in the same PR.

🟡 Should Improve (4)
5. tests/cephfs/cephfs_mirroring/validate_snapdiff_perf_results.py — compare_sync_duration never returns non-zero — Perf regressions are only log.warning; the function returns None and run() always returns 0. The new empty-result guards are correct, but the actual comparison is a no-op gate.

6. tests/cephfs/cephfs_mirroring/snapdiff_perf_improvements_across_releases.py — log.error("nfs files: {out}") wrong severity — Using ERROR for a diagnostic listing of leftover files pollutes CI log scanners with false-positive error lines on every clean run. Use log.info.

7. tests/cephfs/cephfs_system/cephfs_system_utils.py — node.node.exit_status is a private/racy attribute — The exit status should be captured from the return value or via try/except check_ec=True, not from node.node.exit_status which can reflect the previous command's code in async SSH sessions.

8. tests/cephfs/cephfs_mirroring/snapdiff_perf_improvements_across_releases.py — PEP-8: f-strings in log.* calls — New code mixes f-string and %s-style interpolation in log calls. f-strings are evaluated eagerly even when the log level is inactive; use %s-style lazy interpolation throughout (run flake8 --select=G to catch all).

Harishacharya-redhat:
addressed bob-review

tests/cephfs/cephfs_mirroring/snapdiff_perf_improvements_across_releases.py
```diff
@@ -121,9 +139,37 @@ def get_ganesha_pid(nfs_server_node):
...
+                pid = get_ganesha_pid(nfs_node)
+                if pid:
+                    return pid
+                time.sleep(interval)
+            return None
+
+        ganesha_pid = wait_for_ganesha_pid(nfs_server_node, timeout=60)
         if not ganesha_pid:
-            log.error("Failed to get ganesha process PID")
+            log.error(
```
Manimaran-MM: In VMs sometimes it takes additional time to bring up the daemon. Can we do wait for nfs process first and then get the pid should always work. And if that fails, we can fail the TC
```    
if not fs_util_ceph1.wait_for_nfs_process(
        source_clients[0],
        nfs_name,
        timeout=60,
        desired_state="running",
    ):

Harishacharya-redhat:
@Manimaran-MM in this PR: https://github.com/red-hat-storage/cephci/pull/6082, we have polling logic. Ill wil resolve this after merging this PR

Harishacharya-redhat:
this have too many files to review, SO spliting this PR into multiple PRs


1.fix/cephfs-upgrade-automation
[#6096](https://github.com/red-hat-storage/cephci/pull/6096)
2.fix/cephfs-post-upgrade-clone
[#6097](https://github.com/red-hat-storage/cephci/pull/6097)
3.fix/cephfs-crash-check-annotations
[#6098](https://github.com/red-hat-storage/cephci/pull/6098)
4.fix/cephfs-mirror-sync-perf
[#6099](https://github.com/red-hat-storage/cephci/pull/6099)

PR #6099, opened 2026-08-17
Improve CephFS mirror snapshot sync perf test robustness

Harishacharya-redhat:
## Summary

Hardens CephFS mirror snapshot sync performance tests to parse and compare sync durations reliably across daemon output formats.

### Problem
Mirror sync perf tests compared raw `sync_duration` strings from `cephfs-mirror` daemon status. Values vary in format (`49s`, `1m 16`, `76.5`, etc.), which led to brittle parsing and inconsistent pass/fail in tier-2 mirror sync perf runs.

### Changes
**`cephfs_mirroring_utils.py`**
- Add `parse_sync_duration_to_seconds()` for numeric and human-readable duration strings
- Add helpers to collect and compare sync durations across mirror daemons more robustly

**Perf test modules**
- Refactor `test_cephfs_mirror_snapshot_sync_perf_{small,mixed,large}_files.py` to use the shared duration parsing/comparison logic

**Suite**
- `tier-2_cephfs_mirror_snapshot_sync_perf.yaml`: minor suite metadata update

### Scope
- Mirror sync perf only; no upgrade or NFS suite changes

## Test plan
- [ ] `suites/tentacle/cephfs/tier-2_cephfs_mirror_snapshot_sync_perf.yaml`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/suites/tentacle/cephfs/tier-2_cephfs_mirror_snapshot_sync_perf.yaml
+++ b/suites/tentacle/cephfs/tier-2_cephfs_mirror_snapshot_sync_perf.yaml
@@ -220,6 +220,7 @@ tests:
       module: cephfs_mirroring.test_cephfs_mirror_snapshot_sync_perf_large_files.py
       name: CephFS Mirror snapshot sync performance - large files
       polarion-id: "CEPH-83574100"
+      comments: IBMCEPH-13886
   - test:
       abort-on-fail: false
       desc: "Snapshot Sync Performance Test - Mixed dataset (100 small + 5x1GiB large across dirs) with varying datasync thread counts on kernel/fuse/NFS"
@@ -244,4 +245,5 @@ tests:
       module: cephfs_mirroring.test_cephfs_mirror_snapshot_sync_perf_mixed_files.py
       name: CephFS Mirror snapshot sync performance - mixed files
       polarion-id: "CEPH-83574101"
+      comments: IBMCEPH-13886
 
--- a/tests/cephfs/cephfs_mirroring/cephfs_mirroring_utils.py
+++ b/tests/cephfs/cephfs_mirroring/cephfs_mirroring_utils.py
@@ -7,6 +7,7 @@ It contains all the re-useable functions related to cephfs mirroring feature
 import csv
 import json
 import random
+import re
 import secrets
 import string
 import time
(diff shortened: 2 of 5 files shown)


tests/cephfs/cephfs_mirroring/cephfs_mirroring_utils.py
```diff
@@ -23,6 +24,170 @@
...
+    log.info("Rule: multi-thread must be equal or faster than 1-thread")
+    log.info("=" * 70)
+
+    failed = False
+    for mtype in mount_types:
+        single = durations[mtype].get(1)
+        if single is None or single <= 0:
+            log.warning(
+                "No valid 1-thread baseline for mount=%s; skipping validation",
+                mtype,
```
Manimaran-MM, line 120: Skipping the mount entirely when the 1-thread baseline is missing or `<= 0` can make this function return `True` even when no mount was actually validated (all skipped → PASSED). That is a false pass for the check this PR is adding.

This is especially likely on the small-file case: tests round `sync_duration_sec` to 1 decimal before calling this helper, so a real `0.04s` baseline becomes `0.0` and is skipped.

Please fail (or at least fail when **every** mount is skipped) instead of treating "no baseline" as success. Zero is also documented as a valid duration in `parse_sync_duration_to_seconds`, so treating it as invalid here is inconsistent.

suites/tentacle/cephfs/tier-2_cephfs_mirror_snapshot_sync_perf.yaml
```diff
@@ -220,6 +220,7 @@ tests:
       module: cephfs_mirroring.test_cephfs_mirror_snapshot_sync_perf_large_files.py
       name: CephFS Mirror snapshot sync performance - large files
       polarion-id: "CEPH-83574100"
+      comments: IBMCEPH-13886
```
Manimaran-MM, line 223: `IBMCEPH-13886` is an RGW CORS bug ("Methods set using rgw_gcors_allow_methods not recognised"), not a CephFS mirror sync-duration issue. Same value is also attached to the mixed-files test below.

Please point `comments` at the actual tracker for this perf/parse failure (or drop it). Small-files in this suite has no `comments` key at all — if the known-issue tag is intentional, it should be consistent across the three tests.

tests/cephfs/cephfs_mirroring/test_cephfs_mirror_snapshot_sync_perf_small_files.py
```diff
@@ -563,39 +568,73 @@ def run(ceph_cluster, **kw):
...
+                    source_fs,
+                    fsid,
+                    asok_file,
+                    filesystem_id,
+                    peer_uuid,
+                    mirroring_paths,
+                )
+            except CommandFailed as e:
+                idle_ok = False
+                log.error(
```
Manimaran-MM, line 587: Accidental implicit string concat (`"...for " "thread_count=%d"`). Same leftover is in the large and mixed modules. Please make it one string.

tests/cephfs/cephfs_mirroring/cephfs_mirroring_utils.py
```diff
@@ -982,15 +1147,23 @@ def validate_snapshot_sync_status(
...
-                    )
+                    sync_duration = parse_sync_duration_to_seconds(raw_duration)
+                    if sync_duration is None:
+                        log.error(
+                            "Snapshot %s matched but sync_duration is "
+                            "missing/unparseable (raw=%r)",
+                            snapshot_name,
+                            raw_duration,
+                        )
+                        raise CommandFailed(
```
Manimaran-MM, line 1158: This `CommandFailed` is the same exception `@retry` already uses to mean "snapshot not synced yet". When the snap **is** present but `sync_duration` is missing/unparseable, this helper (and every caller, including `_poll_snapshot_sync` with 120 tries) will keep retrying as if sync is still pending, then fail with "did not sync within timeout".

Please raise a distinct error (or return the matched snap and let the caller fail immediately) so a parse problem is not confused with a sync timeout. This method is shared by non-perf mirroring tests too.

tests/cephfs/cephfs_mirroring/cephfs_mirroring_utils.py
```diff
@@ -2452,6 +2629,17 @@ def validate_daemon_status_for_fs(
         log.info("FSID '%s' validated for %s", remote_fsid, fs_name)
 
 
+def _ensure_ceph_common(asok_file):
```
Manimaran-MM, line 2632: `_ensure_ceph_common()` is new but never called. `wait_for_sync_idle()` used to `dnf install ceph-common` on every daemon node before querying ASOK; that install was removed and not wired through this helper.

Please call this from `wait_for_sync_idle()`, or drop the unused helper so the install is not silently lost for standalone callers.

tests/cephfs/cephfs_mirroring/test_cephfs_mirror_snapshot_sync_perf_small_files.py
```diff
@@ -622,6 +661,9 @@ def run(ceph_cluster, **kw):
             )
         log.info("CSV results written to %s", csv_file)
 
+        if not compare_and_validate_sync_durations(results):
```
Manimaran-MM, line 664: `results` stores `sync_duration_sec` rounded to 1 decimal (same in large/mixed). That value is what this helper compares, so a real `0.04s` 1-thread baseline becomes `0.0` and hits the `single <= 0` skip path — the new multi-thread check is then skipped.

Please keep full precision in `results` for comparison (round only for CSV/log display). Same change needed in the large and mixed modules.

tests/cephfs/cephfs_mirroring/cephfs_mirroring_utils.py
```diff
@@ -2468,14 +2656,12 @@ def wait_for_sync_idle(fs_name, fsid, asok_file, filesystem_id, peer_uuid, paths
...
-    for node, asok in asok_file.items():
-        try:
-            asok[0].exec_command(
-                sudo=True, cmd="dnf install -y ceph-common --nogpgcheck"
-            )
-        except Exception as e:
-            log.warning("dnf install ceph-common failed (non-fatal): %s", e)
+    if not asok_file:
+        raise RuntimeError("asok_file dict is empty; no mirror daemon to query")
+
```
Manimaran-MM, line 2663: The install loop was replaced with `next(iter(asok_file))` and `_ensure_ceph_common()` is never invoked. For this suite there is only one mirror daemon, so querying the first ASOK is probably fine — but `ceph --admin-daemon` still needs `ceph-common` on that node.

Please call `_ensure_ceph_common(asok_file)` before this query (or keep the previous per-node install).

Manimaran-MM (changes requested):
Thanks for centralizing duration parsing — `parse_sync_duration_to_seconds()` and the comparison table are the right direction. A few issues should be fixed before merge:

1. **False pass:** `compare_and_validate_sync_durations()` skips a mount when the 1-thread baseline is missing or `<= 0`, then returns PASSED if every mount was skipped. Combined with rounding `sync_duration_sec` to 1 decimal, short syncs become `0.0` and the new check never runs.
2. **Misleading retries:** `validate_snapshot_sync_status()` raises `CommandFailed` for an unparseable duration. That is the same signal `@retry` / `_poll_snapshot_sync` use for "not synced yet", so a parse problem looks like a sync timeout and is retried for minutes. This helper is shared beyond these perf tests.
3. **Dead helper:** `_ensure_ceph_common()` is unused; `wait_for_sync_idle()` no longer installs `ceph-common`.
4. **Wrong Jira:** suite `comments: IBMCEPH-13886` is an RGW CORS bug, not a CephFS mirror sync-duration issue.

Nit: leftover `"for " "thread_count=%d"` concat in all three perf modules.

Please also add unit tests for `parse_sync_duration_to_seconds()` covering `49s`, `1m 16`, `1m 16s`, `76.5`, `0`, and `None`.

tests/cephfs/cephfs_mirroring/cephfs_mirroring_utils.py
```diff
@@ -2468,14 +2656,12 @@ def wait_for_sync_idle(fs_name, fsid, asok_file, filesystem_id, peer_uuid, paths
...
+    node, asok = next(iter(asok_file.items()))
     asok_basename = asok[1].rsplit("/", 1)[-1]
     asok_dir = f"/var/run/ceph/{fsid}"
```
Manimaran-MM, line 2666: **Severity: High**

`_ensure_ceph_common()` was extracted from `wait_for_sync_idle()` but is never called. The old path ran `dnf install -y ceph-common` on every mirror-daemon node before the admin-socket query; now nothing installs `ceph-common`, so the first `ceph --admin-daemon` call can fail on a fresh mirror node while `wait_for_sync_idle()` still retries and the perf test may false-pass cleanup or hit flaky idle waits.

```python
    if not asok_file:
        raise RuntimeError("asok_file dict is empty; no mirror daemon to query")

    _ensure_ceph_common(asok_file)
    node, asok = next(iter(asok_file.items()))
```

tests/cephfs/cephfs_mirroring/cephfs_mirroring_utils.py
```diff
@@ -23,6 +24,170 @@
...
+    log.info("SYNC DURATION IMPROVEMENT VS 1-THREAD")
+    log.info("Rule: multi-thread must be equal or faster than 1-thread")
+    log.info("=" * 70)
+
+    failed = False
+    for mtype in mount_types:
+        single = durations[mtype].get(1)
+        if single is None or single <= 0:
+            log.warning(
+                "No valid 1-thread baseline for mount=%s; skipping validation",
```
Manimaran-MM, line 119: **Severity: High**

When the 1-thread baseline is missing or `<= 0`, this helper logs a warning and **skips** the multi-thread comparison for that mount type, then returns `True`. A slower multi-thread run can therefore pass even though the core perf assertion (multi-thread must be ≤ 1-thread) never ran.

```python
        if single is None or single <= 0:
            log.error(
                "No valid 1-thread baseline for mount=%s (value=%r); failing validation",
                mtype,
                single,
            )
            failed = True
            continue
```

Treat a missing/zero baseline as a failure (or `return 1` from the perf test) unless mapped to a filed tracker.

tests/cephfs/cephfs_mirroring/test_cephfs_mirror_snapshot_sync_perf_small_files.py
```diff
@@ -563,39 +568,73 @@ def run(ceph_cluster, **kw):
...
+                            _remount_nfs(
+                                source_clients[0],
+                                mount_dirs["nfs"],
+                                nfs_server,
+                                nfs_export_name,
+                                fs_util_ceph1,
+                            )
+
+            if not idle_ok:
+                return 1
```
Manimaran-MM, line 637: **Severity: Medium**

This PR changes mirror sync perf parsing/validation but the PR body has no Jenkins/magna link or pasted log showing `RESULT |` lines and the new `Single vs multi-thread validation PASSED` summary from `tier-2_cephfs_mirror_snapshot_sync_perf.yaml`. Please attach a run log before merge.

Manimaran-MM (changes requested):

PR #6108, opened 2026-08-20
Add NFS-Ganesha CephX key rotation coverage for IBMCEPH-16471

mananxkhanna:
# Description

<details>

</details>


tests/nfs/test_nfs_cephx_key_rotation.py
```diff
@@ -0,0 +1,635 @@
...
+    found = {}
+    for pattern in patterns:
+        try:
+            nfs_node.exec_command(
+                sudo=True,
+                cmd=f'grep "{pattern}" /tmp/nfs_ganesha_log',
+            )
+            found[pattern] = True
+            log.info("  [FOUND] pattern %r in %s logs", pattern, daemon_name)
+        except Exception:
```
Manimaran-MM: **Severity: High**

`except Exception: found[pattern] = False` treats *any* failure of the `grep` command identically to "pattern not found" — including infra-level failures like the log file being empty/missing because the preceding `cephadm logs --name {daemon_name} > /tmp/nfs_ganesha_log` silently captured nothing (e.g. wrong `daemon_name` from the fallback at line 249, or the file simply doesn't exist on `nfs_node`).

This directly threatens Scenario B's core assertion (`expect_present=False`, the actual 9.1z2 fix verification): if log capture quietly fails, every pattern reports "absent" and the test reports a clean PASS even though the check never ran — a false pass on the PR's main claim.

Distinguish "grep found nothing" (exit code 1) from other failures, and fail loudly on the latter:

```python
for pattern in patterns:
    rc, out, err = nfs_node.exec_command(
        sudo=True,
        cmd=f'grep "{pattern}" /tmp/nfs_ganesha_log',
        check_ec=False,
    )
    if rc not in (0, 1):
        raise OperationFailedError(
            f"grep failed unexpectedly (rc={rc}) while checking Ganesha logs for {pattern!r}: {err}"
        )
    found[pattern] = rc == 0
```

(Adjust to match whatever `exec_command`'s actual return shape/`check_ec` semantics are in this codebase — the point is to only treat "no match" as absence, not every exception.)

tests/nfs/test_nfs_cephx_key_rotation.py
```diff
@@ -0,0 +1,635 @@
...
+
+    # Step 6 — recover by re-applying the export config.
+    # Fetch the current export config from the client node (which has ceph in PATH)
+    # and re-apply it so Ganesha picks up the new key reference.
+    # Uses the same pattern as nfs_operations.permission().
+    log.info("Step 6: recovering via ceph nfs export apply ...")
+    export_conf_raw = Ceph(client).nfs.export.get(nfs_name, export)
+    client.exec_command(
+        sudo=True,
+        cmd=f"echo '{export_conf_raw}' > /tmp/nfs_export_recovery.conf",
```
Manimaran-MM: **Severity: Medium**

`export_conf_raw` (the live JSON/text export config fetched from the cluster) is interpolated directly into a single-quoted shell string. If the export config ever contains a single quote or other shell-special character, this breaks the command — at best a spurious test failure, at worst unintended shell behavior.

Quote it safely, or write it without a shell round-trip:

```python
import shlex

client.exec_command(
    sudo=True,
    cmd=f"echo {shlex.quote(export_conf_raw)} > /tmp/nfs_export_recovery.conf",
)
```

tests/nfs/test_nfs_cephx_key_rotation.py
```diff
@@ -0,0 +1,635 @@
...
+    try:
+        # ── setup — use the standard helper (skip_mount=True so we control
+        #   mount/unmount explicitly per rotation step).
+        # setup_nfs_cluster already verifies the export exists before returning.
+        setup_nfs_cluster(
+            clients,
+            nfs_server_hostnames,
+            NFS_PORT,
+            NFS_VERSION,
+            nfs_name,
```
Manimaran-MM, line 464: **Severity: Medium**

Both suite entries in `suites/tentacle/nfs/tier1-nfs-ganesha-cephx-key-rotation.yaml` set `config: {nfs_version: "4.1", ...}`, but `run()` never reads `config.get("nfs_version")` — it passes the module-level `NFS_VERSION` constant here (and as the default for `_mount_and_verify`'s `version` param) unconditionally. Changing `nfs_version` in the suite YAML silently has no effect, which will confuse whoever tries to run this against a different NFS version later.

```python
# in run(), alongside the other config.get(...) reads:
nfs_version = config.get("nfs_version", NFS_VERSION)
...
setup_nfs_cluster(
    clients,
    nfs_server_hostnames,
    NFS_PORT,
    nfs_version,
    ...
)
```

and thread `nfs_version` through to `_mount_and_verify(...)` calls instead of relying on its default.

tests/nfs/test_nfs_cephx_key_rotation.py
```diff
@@ -0,0 +1,635 @@
...
+                               if False, raise if ANY pattern IS found in logs
+                               (use when verifying the fix).
+
+    Returns:
+        dict mapping pattern -> bool (True = found in logs)
+    """
+    # Get the daemon name that is running on nfs_node from ceph orch ps.
+    # ceph orch ps output: <daemon_name> <host> <status> ...
+    out, _ = client.exec_command(
+        sudo=True, cmd=f"ceph orch ps | grep {nfs_name}"
```
Manimaran-MM: **Severity: Low**

`_check_ganesha_logs` re-runs `ceph orch ps | grep {nfs_name}` and re-parses the host/daemon columns — the same remote call and parsing `_get_active_nfs_node` (above) just did to produce `nfs_node`. Since `nfs_node` is already known here, consider having `_get_active_nfs_node` return `(node, daemon_name)` together so this function doesn't need to redo the lookup:

```python
def _get_active_nfs_node(client, nfs_name, nfs_nodes):
    ...
    return node, daemon_name  # instead of just node
```

Then `_check_ganesha_logs(client, nfs_name, nfs_node, daemon_name, patterns, expect_present)` skips straight to fetching logs.

tests/nfs/test_nfs_cephx_key_rotation.py
```diff
@@ -0,0 +1,635 @@
...
+# ── helpers ───────────────────────────────────────────────────────────────────
+
+def _get_nfs_entities(installer, nfs_name, retries=10, delay=6):
+    """Return all CephX entities belonging to the NFS cluster as a list.
+
+    Polls up to retries*delay seconds to allow the NFS daemon time to
+    register its CephX keys after cluster creation.
+    """
+    prefix = f"client.nfs.{nfs_name}"
+    for attempt in range(1, retries + 1):
```
Manimaran-MM: **Severity: Low**

This hand-rolls a bounded poll loop (`for attempt in range(1, retries + 1): ... sleep(delay)`) where the repo already has `ceph.waiter.WaitUntil` for this exact pattern elsewhere in the codebase. Not required, but worth switching to for consistency:

```python
from ceph.waiter import WaitUntil

for w in WaitUntil(timeout=retries * delay, interval=delay):
    entities = [...]
    if entities:
        return entities
if w.expired:
    raise OperationFailedError(...)
```

Manimaran-MM (changes requested):
Reviewed the new NFS-Ganesha CephX key-rotation test and CLI helpers. The overall structure (setup/cleanup via `nfs_operations`, two clearly-separated scenarios, JSON response validation) is solid. One High-severity finding: the Ganesha-log check in `_check_ganesha_logs` swallows all exceptions from the `grep` call as "pattern absent," which risks a false PASS on Scenario B's core fix-verification assertion if log capture fails for an unrelated reason. Also flagged two Medium issues (unsafe shell quoting when re-applying the export config, and a `nfs_version` suite config key that's silently ignored by `run()`) and two Low suggestions (de-duplicating the `ceph orch ps` daemon lookup, and reusing the existing `WaitUntil` helper instead of a hand-rolled poll loop).

conf/tentacle/nfs/1admin-3node-1client.yaml
Manimaran-MM: Can we re-use the existing available config for nfs and remove this?

tests/nfs/test_nfs_cephx_key_rotation.py
mananxkhanna: This matches the existing NFS export get → apply pattern in this repo, including nfs_operations.permission():

```
out = Ceph(client).nfs.export.get(nfs_name, f"{nfs_export}_0")
client.exec_command(sudo=True, cmd=f"echo '{out}' > export.conf")
Ceph(client).nfs.export.apply(nfs_name, "export.conf")
```
The same echo `'{out}' > export.conf` is used in several other NFS tests (export edit / rootsquash / HA failover). Export JSON from ceph nfs export get is double-quoted, so a single quote in the payload is unlikely. Scenario A recovery follows that established helper rather than introducing a different file-write path.

Happy to switch to shlex.quote if you would rather this test not follow that pattern.

tests/nfs/test_nfs_cephx_key_rotation.py
mananxkhanna: exec_command raises on any non-zero exit (check_ec defaults to True). grep exits 1 when the pattern is missing, so “not found” was an exception. The bare except Exception also swallowed infra failures (empty/missing log file, wrong daemon, grep cannot open the file). Scenario B uses expect_present=False, so every pattern marked absent is a PASS — including when the log check never actually ran.

Fixed in _check_ganesha_logs:

After cephadm logs, fail if /tmp/nfs_ganesha_log is empty or missing (test -s).
Run grep with check_ec=False and use nfs_node.exit_status (this helper returns (stdout, stderr), not rc, out, err).
Exit 0 = found, 1 = not found, anything else raises.
Scenario B can now pass only when grep searched a real log file and the bug strings are absent.

tests/nfs/test_nfs_cephx_key_rotation.py
mananxkhanna: Agreed — the suite key was unused.

run() documented nfs_version and the YAML set nfs_version: "4.1", but setup/mount always used the module constant NFS_VERSION. Both were "4.1", so current runs were correct; changing only the YAML would have had no effect.

Fixed: run() now does nfs_version = config.get("nfs_version", NFS_VERSION), passes it into setup_nfs_cluster, and threads it through both scenarios into _mount_and_verify(..., version=nfs_version). Default remains "4.1" if the suite omits the key.

tests/nfs/test_nfs_cephx_key_rotation.py
mananxkhanna: Agreed this is duplicate work, not a functional bug.

_get_active_nfs_node already runs ceph orch ps | grep to pick the host. _check_ganesha_logs runs the same command again to get daemon_name. Returning (node, daemon_name) from the first helper would avoid the second lookup.

Leaving it for now to keep this PR focused on the High false-pass in Scenario B and the unused nfs_version config. Happy to fold the cleanup in if you prefer it in this change.

tests/nfs/test_nfs_cephx_key_rotation.py
mananxkhanna: Switched _get_nfs_entities to ceph.waiter.WaitUntil for consistency with nfs_operations.py and other NFS tests. Same timeout/interval (60s / 6s); parsing of ceph auth ls is unchanged.

tests/nfs/test_nfs_cephx_key_rotation.py
```diff
@@ -0,0 +1,686 @@
...
+    Config keys (all optional):
+        nfs_version (str): NFS mount version, default "4.1"
+        clients (int): number of client nodes to use, default 1
+        nfs_name (str): NFS cluster name, default "cephfs-nfs"
+        fs_name (str): CephFS filesystem name, default "cephfs"
+        nfs_mount (str): client mount point, default "/mnt/nfs_cephx"
+        operation (str): "manual_rotate_key", "cluster_rotate_key", or "all" (default "all")
+
+    Returns:
+        0 on success, 1 on failure.
```
Manimaran-MM, line 421: **Severity: High**

`operation` is read from suite config but never validated. If the YAML has a typo (`manual_rotate_keys`) or an unknown value, both scenario `if` blocks are skipped and `run()` still returns `0` after `setup_nfs_cluster` — a false pass with no rotation exercised.

```python
    valid_ops = {"manual_rotate_key", "cluster_rotate_key", "all"}
    if scenario not in valid_ops:
        log.error(
            "Invalid operation %r; expected one of %s",
            scenario,
            sorted(valid_ops),
        )
        return 1
```

Add this immediately after `scenario = config.get("operation", "all").lower()`.

tests/nfs/test_nfs_cephx_key_rotation.py
```diff
@@ -0,0 +1,686 @@
...
+                nfs_name,
+                export,
+                nfs_server_ip,
+                mount_point,
+                nfs_version,
+            )
+
+        # ── cluster_rotate_key ─────────────────────────────────────────────
+        if scenario in ("cluster_rotate_key", "all"):
+            _scenario_b_rotate_key_command(
```
Manimaran-MM: **Severity: Medium**

`run()` returns `0` without an unmistakable terminal pass line. Scenario helpers log `=== Scenario A/B PASSED ===`, but log triage greps for a single `TEST PASSED` / `Test passed:` before `return 0` in `run()`.

```python
        log.info(
            "TEST PASSED - NFS CephX key rotation completed (operation=%s)",
            scenario,
        )
        return 0
```

tests/nfs/test_nfs_cephx_key_rotation.py
```diff
@@ -0,0 +1,686 @@
...
+
+
+def _find_export_entity(installer, nfs_name):
+    """Return the export CephFS entity (client.nfs.<name>.<fs>.<hash>)."""
+    entities = _get_nfs_entities(installer, nfs_name)
+    for e in entities:
+        parts = e.split(".")
+        # Pattern: client.nfs.<name>.<fs_name>.<8-char-hash>
+        if len(parts) >= 5 and not parts[-1].endswith("-rgw") and len(parts[-1]) == 8:
+            return e
```
Manimaran-MM: **Severity: Medium**

This PR adds new NFS test coverage and CLI helpers but the PR body has no Jenkins/magna run log or pasted snippet showing a terminal pass line from this suite. Please attach a link or fenced log excerpt from `suites/tentacle/nfs/tier1-nfs-ganesha-cephx-key-rotation.yaml` showing `TEST PASSED` / scenario pass lines before merge.

Manimaran-MM (changes requested):

tests/nfs/test_nfs_cephx_key_rotation.py
mananxkhanna: Agreed — a typo in operation would skip both scenarios and still return 0 after setup.

run() now validates against {"manual_rotate_key", "cluster_rotate_key", "all"} immediately after config.get("operation", "all"). Unknown values log and return 1 before setup_nfs_cluster.

tests/nfs/test_nfs_cephx_key_rotation.py
mananxkhanna: Agreed. Pass lines now start with TEST PASSED so log triage can grep them:

Scenario A: TEST PASSED - Scenario A: manual ceph auth rotate reproduced the IBMCEPH-16471 bug ...; export apply recovered the share and IO succeeded
Scenario B: TEST PASSED - Scenario B: ceph nfs cluster rotate-key rotated keys and updated exports with no Ganesha errors; mount and IO succeeded without a manual export apply
run(): TEST PASSED - NFS CephX key rotation completed (operation=%s) immediately before return 0

tests/nfs/test_nfs_cephx_key_rotation.py
mananxkhanna: 2026-08-20 17:37:29,473 - cephci - run:1391 - INFO - Ceph cluster Version: ceph version 20.2.1-391.el9cp (014e8d16be066af1356f3551eb860d1018a3eae7) tentacle (stable - RelWithDebInfo) release 9.9.1.2
2026-08-20 17:37:29,475 - cephci - run:1164 - INFO - Test <module 'test_nfs_cephx_key_rotation' from '/Users/manankhanna/Desktop/cephci/tests/nfs/test_nfs_cephx_key_rotation.py'> passed
Test <module 'test_nfs_cephx_key_rotation' from '/Users/manankhanna/Desktop/cephci/tests/nfs/test_nfs_cephx_key_rotation.py'> passed
2026-08-20 17:37:29,475 - cephci - run:1225 - INFO - 
All test logs located here: /tmp/cephci-run-LNR3RL

All test logs located here: /tmp/cephci-run-LNR3RL

TEST NAME                        TEST DESCRIPTION                                               DURATION                         STATUS                    COMMENTS
setup pre-requisites             Install software pre-requisites for cluster deployment.        0:15:43.036080                   Pass                             
Deploy cluster using cephadm     Bootstrap and deploy services.                                 0:17:51.301365                   Pass                             
configure client                 Configure the NFS client node                                  0:02:11.206889                   Pass                             
NFS CephX key rotation — Scena   Rotate all NFS CephX keys individually via ceph auth rotate    0:05:54.846988                   Pass                             
NFS CephX key rotation — Scena   Verify the 9.1z2 fix: run ceph nfs cluster rotate-key and as   0:04:22.073147                   Pass         
2026-08-20 17:37:29,549 - cephci - run:1401 - INFO - final rc of test run 0
                    
[configure_client_0.log](https://github.com/user-attachments/files/31265475/configure_client_0.log)
[Deploy_cluster_using_cephadm_0.log](https://github.com/user-attachments/files/31265478/Deploy_cluster_using_cephadm_0.log)
[index.html](https://github.com/user-attachments/files/31265479/index.html)
[NFS_CephX_key_rotation_—_Scenario_A_(manual_rotate_path)_0.log](https://github.com/user-attachments/files/31265480/NFS_CephX_key_rotation_._Scenario_A_.manual_rotate_path._0.log)
[NFS_CephX_key_rotation_—_Scenario_B_(ceph_nfs_cluster_rotate-key)_0.log](https://github.com/user-attachments/files/31265481/NFS_CephX_key_rotation_._Scenario_B_.ceph_nfs_cluster_rotate-key._0.log)
[setup_pre-requisites_0.log](https://github.com/user-attachments/files/31265482/setup_pre-requisites_0.log)
[startup.log](https://github.com/user-attachments/files/31265483/startup.log)

Manimaran-MM (commented):
Cursor Review Completed

tests/nfs/test_nfs_cephx_key_rotation.py
```diff
@@ -0,0 +1,462 @@
...
+        _run_rotate_key_live_mount(
+            installer,
+            nfs_nodes,
+            clients,
+            nfs_name,
+            export,
+            mount_point,
+        )
+
+        log.info("TEST PASSED - NFS CephX key rotation completed")
```
Manimaran-MM, line 481: **Severity: Medium**

The PR body has no Jenkins/magna link or pasted log snippet, and the only run evidence in the thread is from the earlier two-scenario version (Scenario A manual rotate + Scenario B cluster rotate-key). This revision is a single live-mount `rotate-key` path only, so those logs no longer exercise the code under review.

Please attach a fresh run of `suites/tentacle/nfs/tier1-nfs-ganesha-cephx-key-rotation.yaml` showing the terminal pass lines from this module, e.g.:

```
TEST PASSED - ceph nfs cluster rotate-key rotated keys (values changed) ...
TEST PASSED - NFS CephX key rotation completed
Test  passed
final rc of test run 0
```

tests/nfs/test_nfs_cephx_key_rotation.py
```diff
@@ -0,0 +1,462 @@
...
+        )
+    log.info("All expected exports listed: %s", all_exports)
+
+
+def _write_test_file(client, mount_point, filename="cephx_testfile"):
+    """Write a small test file to the mounted share."""
+    client.exec_command(
+        sudo=True, cmd=f"dd if=/dev/urandom of={mount_point}/{filename} bs=1M count=10"
+    )
+    log.info("Wrote test file %s/%s", mount_point, filename)
```
Manimaran-MM, line 191: **Severity: Medium**

`_write_test_file()` runs a 10 MB `dd` over the live NFS mount with no `timeout=` on `exec_command`. If Ganesha or the mount stalls mid-rotation, the helper can block until the default remote timeout (600s) instead of failing fast.

`test_nfs_rdma_basic.py` already bounds NFS `dd` this way — mirror that here (and add a timeout on the follow-up `ls` in `_verify_test_file` too):

```python
client.exec_command(
    sudo=True,
    cmd=(
        f"dd if=/dev/urandom of={mount_point}/{filename} bs=1M count=10 "
        f"conv=fsync status=none"
    ),
    timeout=300,
)
```

Manimaran-MM (commented):
Cursor review of the refactored single-scenario rotate-key test. Two Medium findings: attach a fresh suite run log for this revision, and bound NFS `dd`/`ls` with `timeout=` on `exec_command`. Prior High items (Ganesha log grep false-pass, unused `nfs_version`) appear addressed.

tests/nfs/test_nfs_cephx_key_rotation.py
mananxkhanna: 2026-08-20 22:37:13,521 - cephci - run:1164 - INFO - Test <module 'test_nfs_cephx_key_rotation' from '/Users/manankhanna/Desktop/cephci/tests/nfs/test_nfs_cephx_key_rotation.py'> passed
Test <module 'test_nfs_cephx_key_rotation' from '/Users/manankhanna/Desktop/cephci/tests/nfs/test_nfs_cephx_key_rotation.py'> passed
2026-08-20 22:37:13,522 - cephci - run:1225 - INFO - 
All test logs located here: /tmp/cephci-run-8UWLF7

All test logs located here: /tmp/cephci-run-8UWLF7

TEST NAME                        TEST DESCRIPTION                                               DURATION                         STATUS                    COMMENTS
setup pre-requisites             Install software pre-requisites for cluster deployment.        0:15:52.006205                   Pass                             
Deploy cluster using cephadm     Bootstrap and deploy services.                                 0:21:41.422232                   Pass                             
configure client                 Configure the NFS client node                                  0:02:23.044891                   Pass                             
NFS CephX key rotation (ceph n   Verify IBMCEPH-16471 / 9.1z2: run ceph nfs cluster rotate-ke   0:04:42.609588                   Pass                             
2026-08-20 22:37:13,622 - cephci - run:1401 - INFO - final rc of test run 0

tests/nfs/test_nfs_cephx_key_rotation.py
mananxkhanna: Thanks — the 10MB dd here is not running while Ganesha is restarting.

_write_test_file is only used:

before rotate-key (pre_rotate_file), mount is up
after _wait_for_live_mount_io has already succeeded (post_rotate_file)
Rotate/redeploy is in between. If the mount is wedged after restart, we fail in _wait_for_live_mount_io (timeout 20 ls + exec_command timeout=30), well under 600s. We never start dd on a hung mount.

test_nfs_rdma_basic.py bounds dd because that IO is the path under test. Here it’s only a smoke check after we already know the mount is back.

Happy to add timeout= on dd/ls as extra safety if you still want it.

mananxkhanna:
Uploading latest pass logs.
[configure_client_0.log](https://github.com/user-attachments/files/31363901/configure_client_0.log)
[Deploy_cluster_using_cephadm_0.log](https://github.com/user-attachments/files/31363902/Deploy_cluster_using_cephadm_0.log)
[index.html](https://github.com/user-attachments/files/31363903/index.html)
[NFS_CephX_key_rotation_(ceph_nfs_cluster_rotate-key)_0.log](https://github.com/user-attachments/files/31363904/NFS_CephX_key_rotation_.ceph_nfs_cluster_rotate-key._0.log)
[setup_pre-requisites_0.log](https://github.com/user-attachments/files/31363905/setup_pre-requisites_0.log)
[startup.log](https://github.com/user-attachments/files/31363906/startup.log)

aravindrrh:
While the PR looks good, I have a few nits:
1. Need to add version check or check if the command "ceph nfs cluster rotate-key " exists using  --help and skip if it is a pre CephX version
2. Delete the config Yaml file, use the common yaml used for all NFS tests
3. We need to scale NFS clusters and exports and this can be takes as another PR

mananxkhanna:
Addressed Arun's review comment:

1. Need to add version check or check if the command "ceph nfs cluster rotate-key " exists using --help and skip if it is a pre CephX version - **Done**
2. Delete the config Yaml file, use the common yaml used for all NFS tests - **Done**
3. We need to scale NFS clusters and exports and this can be takes as another PR - **To be taken up as a separate PR**

aravindrrh (approved):
LGTM

PR #6113, opened 2026-08-21
Fixing testcases after the new GRPC changes with mTLS

aravindrrh:
# Description

A Major Overhaul on the GRPC tests

1. Now GRPC uses mTLS and will not work with Plain-text so we have to give the certs with the grpcurl command. Old method was failing

- Now grpc command uses the `grpcurl -cacert ca.crt -cert client.crt -key client.key  <NFS-Node-IP>:50051 list`
- also we copy the certs required by above command from `/var/lib/ceph/<fsid>/nfs_grpc-client-certs/nfs.<cluster-name>/` on the admin node

2. Also, I found a few false positivies with Log analysis where the earlier grace period messages were found and validated as PASS. This is now fixed

3. For 9.1.x it uses plaintext and only for 9.2 and above certs are used.

here are the logs for 
1. 9.1 :  https://10.8.128.2/ceph-qe-logs/aravindranath/grpc_logs/tier1_nfs_ganesha_grpc_91/
2. 9.2 : https://10.8.128.2/ceph-qe-logs/aravindranath/grpc_logs/tier1_nfs_ganesha_grpc_92/

### DIFF
--- a/tests/nfs/nfs_state_lifecycle/nfs_state_lifecycle_workflows.py
+++ b/tests/nfs/nfs_state_lifecycle/nfs_state_lifecycle_workflows.py
@@ -482,6 +482,34 @@ class NfsStateLifecycleWorkflows:
             self.install_grpcurl()
         except Exception as e:
             log.warning("grpcurl install failed (gRPC checks will be skipped): %s", e)
+
+        try:
+            from test_nfs_grpc import (
+                copy_nfs_grpc_certs,
+                grpc_requires_mtls,
+                resolve_ceph_version,
+            )
+
+            ceph_version = resolve_ceph_version(self.clients[0], self.installer)
+            self._grpc_use_mtls = grpc_requires_mtls(ceph_version)
+            log.info(
+                "Ceph version %s -> NFS gRPC mode: %s",
+                ceph_version,
+                "mTLS" if self._grpc_use_mtls else "plaintext",
+            )
+
+            if self._grpc_use_mtls:
+                try:
+                    self._grpc_cert_dir = copy_nfs_grpc_certs(
+                        self.installer, self.clients[0], nfs_name
+                    )
+                except Exception as e:
+                    log.error("NFS gRPC mTLS cert copy failed: %s", e)
+                    raise
+        except Exception as e:
+            self._grpc_cert_dir = None
+            self._grpc_use_mtls = False
+            log.warning("Unable to configure NFS gRPC cert/version helpers: %s", e)
         log.info("All cluster configs applied for %s", nfs_name)
 
     # ------------------------------------------------------------------
@@ -505,28 +533,44 @@ class NfsStateLifecycleWorkflows:
     def get_client_ids(self):
         """Query active NFS client IDs via gRPC GetClientIds.
 
-        Delegates to ``test_nfs_grpc.get_client_ids()``.
+        Delegates to ``test_nfs_grpc.get_client_ids()`` (mTLS or plaintext
+        based on Ceph version).
 
         Returns:
             tuple: (success: bool, client_ids: list of str).
             Returns (False, []) if grpcurl is not installed or gRPC fails.
         """
-        from test_nfs_grpc import get_client_ids
+        from test_nfs_grpc import GRPC_CERT_DIR_CLIENT, get_client_ids
 
-        return get_client_ids(self.clients[0], self.get_nfs_server_ip())
+        cert_dir = getattr(self, "_grpc_cert_dir", GRPC_CERT_DIR_CLIENT)
+        use_mtls = getattr(self, "_grpc_use_mtls", False)
+        return get_client_ids(
+            self.clients[0],
+            self.get_nfs_server_ip(),
+            cert_dir=cert_dir,
+            use_mtls=use_mtls,
+        )
 
     def get_session_ids(self):
         """Query active NFS session IDs via gRPC GetSessionIds.
 
-        Delegates to ``test_nfs_grpc.get_session_ids()``.
+        Delegates to ``test_nfs_grpc.get_session_ids()`` (mTLS or plaintext
+        based on Ceph version).
 
         Returns:
             tuple: (success: bool, session_ids: list of str).
             Returns (False, []) if grpcurl is not installed or gRPC fails.
         """
-        from test_nfs_grpc import get_session_ids
+        from test_nfs_grpc import GRPC_CERT_DIR_CLIENT, get_session_ids
 
-        return get_session_ids(self.clients[0], self.get_nfs_server_ip())
+        cert_dir = getattr(self, "_grpc_cert_dir", GRPC_CERT_DIR_CLIENT)
+        use_mtls = getattr(self, "_grpc_use_mtls", False)
+        return get_session_ids(
+            self.clients[0],
+            self.get_nfs_server_ip(),
+            cert_dir=cert_dir,
+            use_mtls=use_mtls,
+        )
 
     # ------------------------------------------------------------------
     #  State Tracking via Log Parsing
--- a/tests/nfs/nfs_state_lifecycle/nfs_state_lifecycle_workflows.py
+++ b/tests/nfs/nfs_state_lifecycle/nfs_state_lifecycle_workflows.py
@@ -482,6 +482,22 @@ class NfsStateLifecycleWorkflows:
             self.install_grpcurl()
         except Exception as e:
             log.warning("grpcurl install failed (gRPC checks will be skipped): %s", e)
+
+        try:
+            from test_nfs_grpc import GRPC_CERT_DIR_CLIENT, copy_nfs_grpc_certs
+
+            try:
+                self._grpc_cert_dir = copy_nfs_grpc_certs(
+                    self.installer, self.clients[0], nfs_name
+                )
+            except Exception as e:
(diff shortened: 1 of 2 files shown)


tests/nfs/test_nfs_grpc.py
```diff
@@ -39,284 +45,513 @@
...
+        log.error(f"No expected patterns defined for Event ID {event_id}")
+        return False
+
+    log_after = _fetch_nfs_daemon_log(client, nfs_node, nfs_name)
+    if not log_after.strip():
+        log.error("NFS daemon log is empty after grace event")
+        return False
+
+    for pattern in patterns:
+        before_count = log_before.count(pattern) if log_before is not None else 0
```
Manimaran-MM: **Severity: High**

`log_before` defaults to `""` and `_fetch_nfs_daemon_log()` can return an empty string when `cephadm logs`/`cat` fails. Here `before_count` becomes `0` for every pattern, so any pre-existing ganesha line in `log_after` satisfies `after_count > before_count` and the grace TC passes without a new event line.

Reject empty snapshots before the count loop:

```python
if log_before is not None and not str(log_before).strip():
    log.error("Empty log_before snapshot; cannot verify new grace log lines")
    return False
```

tests/nfs/test_nfs_grpc.py
```diff
@@ -39,284 +45,513 @@
...
+                f"Found new log occurrences of {pattern!r} for Event ID {event_id} "
+                f"({before_count} -> {after_count})"
+            )
+            return True
+        log.info(
+            f"Pattern {pattern!r} did not increase ({before_count} -> {after_count})"
+        )
 
+    # Fallback for callers that did not snapshot logs: full-log search via nfs_log_parser
+    if not log_before:
```
Manimaran-MM: **Severity: High**

When `log_before` is falsy (including `""` from a failed snapshot), this falls back to `nfs_log_parser` on the full daemon log—the same stale-line false pass this PR fixes for the main path. Callers that pass an empty snapshot still return `True` on startup grace lines.

Remove this fallback, or gate it behind an explicit opt-in flag. With the before/after snapshot approach, failure to snapshot should fail the TC:

```python
if not str(log_before or "").strip():
    log.error("Missing non-empty log_before snapshot")
    return False
```

tests/nfs/test_nfs_grpc.py
```diff
@@ -555,85 +764,109 @@ def run(ceph_cluster, **kw):
...
+                raise OperationFailedError(
+                    "Expected client or session count to decrease after unmount; "
+                    f"clients {initial_clients}->{updated_clients}, "
+                    f"sessions {initial_sessions}->{updated_sessions}"
+                )
+
+            log.info("PASS: Client/session count decreased after unmount")
 
         else:
             raise ConfigError(f"Unknown operation: {operation}")
```
Manimaran-MM, line 981: **Severity: Medium**

Success returns `0` without a final terminal pass line. Each branch logs `PASS: ...`, but log triage greps for a single unmistakable summary before `return 0` on the exercised operation.

Add before `return 0`:

```python
log.info(
    "TEST PASSED - operation=%s nfs=%s mTLS grpc checks completed",
    operation,
    nfs_name,
)
return 0
```

Manimaran-MM (changes requested):
Cursor review found stale-log false-pass paths remaining in `verify_grace_logs` when the before-snapshot is empty, plus a missing aggregate pass log before `return 0`. See inline comments. Attached magna log link was not reachable from this environment to cross-check pass lines.

tests/nfs/test_nfs_grpc.py
aravindrrh: Fixed

tests/nfs/test_nfs_grpc.py
aravindrrh: fixed

tests/nfs/test_nfs_grpc.py
```diff
@@ -39,284 +45,503 @@
...
-    cmd = f"grpcurl -plaintext {nfs_ip}:{GRPC_PORT} list"
+    return True
+
+
+def probe_grpc_mtls(client_node, nfs_ip, cert_dir):
+    """
+    Lightweight mTLS connectivity probe: `grpcurl list` must return >=1 service.
+
+    Unlike discovery, this does not assert the full expected service set.
+    """
```
Manimaran-MM, line 387: **Severity: Medium**

`grpcurl` is invoked without `timeout=` on `exec_command`. If the NFS gRPC endpoint stalls (TLS handshake hang, firewall drop, daemon wedged), this TC can block the entire suite indefinitely — the same risk applies to `start_grace_period`, `get_client_ids`, and `get_session_ids`.

Other NFS tests in this repo pass explicit timeouts (e.g. `timeout=60`). Bound grpcurl calls:

```python
out, err = client_node.exec_command(
    sudo=True, cmd=cmd, check_ec=False, timeout=120
)
```

tests/nfs/nfs_state_lifecycle/nfs_state_lifecycle_workflows.py
```diff
@@ -482,6 +482,22 @@ def apply_all_configs(self, nfs_name, lease_lifetime=None):
...
+            from test_nfs_grpc import GRPC_CERT_DIR_CLIENT, copy_nfs_grpc_certs
+
+            try:
+                self._grpc_cert_dir = copy_nfs_grpc_certs(
+                    self.installer, self.clients[0], nfs_name
+                )
+            except Exception as e:
+                self._grpc_cert_dir = GRPC_CERT_DIR_CLIENT
+                log.warning(
+                    "NFS gRPC mTLS cert copy failed (gRPC queries may fail): %s", e
```
Manimaran-MM: **Severity: Medium**

On cert-copy failure this silently falls back to `GRPC_CERT_DIR_CLIENT` and only logs a warning. If that directory is empty or holds stale certs from a prior run, lifecycle gRPC snapshots can return misleading counts (`-1` / empty) instead of failing fast during `apply_all_configs`.

Prefer leaving `_grpc_cert_dir` unset and letting callers surface the copy error, or re-raise after the warning:

```python
except Exception as e:
    log.error("NFS gRPC mTLS cert copy failed: %s", e)
    raise
```

tests/nfs/test_nfs_grpc.py
```diff
@@ -555,89 +758,118 @@ def run(ceph_cluster, **kw):
...
+
+            if not (
+                updated_clients < initial_clients or updated_sessions < initial_sessions
+            ):
+                raise OperationFailedError(
+                    "Expected client or session count to decrease after unmount; "
+                    f"clients {initial_clients}->{updated_clients}, "
+                    f"sessions {initial_sessions}->{updated_sessions}"
+                )
+
```
Manimaran-MM, line 977: **Severity: Medium**

The PR links pass logs at `https://10.8.128.2/ceph-qe-logs/.../tier1_nfs_ganesha_grpc/` but that host is not reachable from automated review. Please paste a short log snippet in the PR body (or a comment) showing the terminal pass line for the exercised operations, e.g.:

```
TEST PASSED - operation=verify_discovery nfs=cephfs-nfs mTLS grpc checks completed
```

This confirms the new summary line is emitted in the actual tier1 run, not only in code.

Manimaran-MM (commented):
Cursor review: prior High findings (empty `log_before` false pass, stale-log fallback) look addressed. Three Medium items remain — grpcurl hang risk, lifecycle cert-copy fallback, and unreachable pass-log link.

tests/nfs/nfs_state_lifecycle/nfs_state_lifecycle_workflows.py
```diff
@@ -482,6 +482,36 @@ def apply_all_configs(self, nfs_name, lease_lifetime=None):
...
+                except Exception as e:
+                    log.error("NFS gRPC mTLS cert copy failed: %s", e)
+                    raise
+            else:
+                self._grpc_cert_dir = None
+        except Exception as e:
+            self._grpc_cert_dir = None
+            self._grpc_use_mtls = False
+            log.warning("Unable to configure NFS gRPC cert/version helpers: %s", e)
         log.info("All cluster configs applied for %s", nfs_name)
```
pdhiran, line 513: The outer except catches the inner raise from cert copy failure and falls back to plaintext:

```
if self._grpc_use_mtls:
    try:
        self._grpc_cert_dir = copy_nfs_grpc_certs(...)
    except Exception as e:
        log.error("NFS gRPC mTLS cert copy failed: %s", e)
        raise  # caught by outer except below
except Exception as e:
    self._grpc_cert_dir = None
    self._grpc_use_mtls = False
    log.warning("Unable to configure NFS gRPC cert/version helpers: %s", e)
```
On 9.2+, a cert problem becomes a misleading plaintext grpcurl TLS failure later instead of failing fast in apply_all_configs(). 

Suggested fix: Split the try blocks — only catch import/version-resolution errors in the outer handler; let cert-copy failures propagate (or re-raise when _grpc_use_mtls was True).

pdhiran (changes requested):
1 minor comment. with that fixed, looks good to merge.

tests/nfs/nfs_state_lifecycle/nfs_state_lifecycle_workflows.py
aravindrrh: removed the else block

pdhiran (approved):
LGTM

PR #6115, opened 2026-08-22
Include public ssh keys that are fips compliant

psathyan:
# Description

Adding ssh public keys that are FIPS compliant so that we end users switch their systems to be compliant are still able to access them with compliant keys.

Other minor changes included as part of this PR are

- Adding nodes on the default SSH public key fingerprint
- Confs are also modified to point to the compliant keys.
- Removing unsupported inventory files which are EOL

- [x] Unit test the code


rakeshcn (approved):

rakeshcn:
LGTM

PR #6124, opened 2026-08-27
Adding preprod.icr.io registry support alongside cp.stg.icr.io

vidya-lokesh:
Add preprod.icr.io registry support alongside cp.stg.icr.io

IBM is migrating staging container images from cp.stg.icr.io to preprod.icr.io with a hard cutover on September 1st, 2026. This PR adds the minimum changes needed to support preprod.icr.io in cephci while retaining full cp.stg.icr.io support during the transition period.

Changes:

ceph/ceph_admin/bootstrap.py — Added explicit preprod tier detection in _detect_registry_tier(). Without this, cephadm bootstrap called with a preprod.icr.io image would fail to look up the correct credentials from .cephci.yaml since "preprod" matches neither "stg", "stage" nor "quay".

conf/inventory/ibm-eu-rhel-{9.2,9.4,9.6,9.8,9.8-rgw,10.0,10.2,10.2-rgw}.yaml — Added preprod.icr.io mirror redirect block pointing to nginx-vm-01.qe.ceph.lab:5000/ibm-stage in the containers registry config written to test VMs. Without this, VMs would attempt to pull preprod.icr.io images directly from the internet instead of the local nginx mirror, causing pull failures in the lab environment.

No existing cp.stg.icr.io behaviour is modified. Both registries coexist and route to the same local ibm-stage/ mirror path.

PR #6156, opened 2026-09-15
Fix RH registry login target in construct_registry / bootstrap

vidya-lokesh:
PR #6124 changed registry-url to always prefer the image pull host (_reg), which caused RH nightly builds to log in to quay.io using stage credentials. Since quay.io is shared via qa@redhat.com, it returns HTTP 429 (rate-limit) under load. The correct login target for RH builds is the stage registry defined in .cephci.yaml (typically registry.stage.redhat.io), not the image pull host.

Changes
construct_registry()

IBM vendor with a known image host → use _reg (preprod.icr.io / cp.stg.icr.io) as before.
RH vendor → prefer cdn_cred.get('registry') so the credential file drives the login URL, not the pull host.
bootstrap() caller

IBM → registry_url = image_registry (unchanged behaviour).
RH → registry_url = '' so construct_registry() falls back to the cred file entry. The image is pulled from quay.io via the lab mirror without a direct login.
Guard updated to product in ('ibm', 'redhat') so the RH path still calls construct_registry() even when registry_url is empty.


vidya-lokesh:
Test logs:
All bvt and sanity runs done on RH builds post the preprod PR merge are using the changes in this PR and all of them have passed.
Eg:
RH 8.1z9 19.2.1-414
BVT: https://10.8.128.2/cephci-jenkins/ibm-cos//RH/8.1/rhel-9/bvt/19.2.1-414/484/logs/
Sanity: https://10.8.128.2/cephci-jenkins/ibm-cos//RH/8.1/rhel-9/sanity/19.2.1-414/250/logs/

rakeshcn:
**Minor nits (non-blocking)**

1. **Follow-up:** `get_registry_details()` in `cli/utilities/configs.py` still prefers `registry_host` from the image for RH (`registry_host or creds.get("registry")`). Non-bootstrap paths (`configure.py`, `cephadm_ansible.py`) may still login to `quay.io` if an image is passed. Worth a small follow-up PR for parity with this bootstrap fix.

2. **Community builds:** For `product == "community"`, `registry_url` is cleared and `construct_registry()` is no longer called. Assuming community pulls from public quay without login — please confirm that's intentional.

3. **`registry-json` on RH:** `construct_registry()` now prefers `cdn_cred.registry` over suite YAML `registry.redhat.io` for RH. That seems correct for login, but worth noting as a behavior change for `registry-json` consumers.

4. **Tests:** Consider adding unit tests for `construct_registry()` RH vs IBM paths (similar to `tests/cli/test_registry_config.py`). Not required to merge given integration results.

vidya-lokesh:
Hey @rakeshcn 
Thanks for the thorough review! Addressing each point:

1. get_registry_details() in cli/utilities/configs.py
Will raise a separate follow-up PR to bring get_registry_details() and the non-bootstrap paths (configure.py, cephadm_ansible.py) in line with this fix.

2. Community builds
Yes, intentional. registry_url is set to "" for all non-IBM products, and community doesn't match product in ("ibm", "redhat"), so construct_registry() is never called — no registry flags are passed to cephadm bootstrap, which is correct for a public upstream image. This is also an improvement over post-#6124 behaviour where registry_url was always the image host, which would have incorrectly triggered construct_registry() for community builds too.

3. registry-json on RH
Noted. Login driven by cdn_cred.registry rather than the suite YAML value is intentional, will call this out explicitly in the follow-up PR.

4. Unit tests
Both BVT and full sanity pipelines have passed with this branch, so I'd like to defer unit tests to a follow-up PR rather than block the merge.

Thank you

vidya-lokesh:
Manual interop verification — PASSED 

Ran suites/tentacle/interop/test-ceph-sanity.yaml on branch fix/rh-registry-login-url-construct-registry via the test executor — all tests passed.

Why this suite covers interop:
The ODF interop pipeline depends on a healthy standalone Ceph cluster as its first step (external mode). This suite validates exactly that — it bootstraps a cluster via cephadm using registry.redhat.io, deploys MON/OSD/RGW/CephFS/MDS, configures a client, and runs parallel RGW/RBD/CephFS I/O. The construct_registry() fix in this PR is directly exercised at the bootstrap step — previously this would fail with Failed to login to custom registry @ quay.io because the registry-url was being set to the image pull host instead of the credential registry from .cephci.yaml. Bootstrap now correctly authenticates against registry.stage.redhat.io, cluster came up healthy, and all I/O tests passed. 
Logs: http://9.11.121.102/logs/RH/9.2/rhel-10/executor/20.2.2-338/interop/3102/logs/test_ceph_sanity/

Additionally, verified that the fix works for upstream Ceph as well — logs here: http://magna002.ceph.redhat.com/ceph-qe-logs/adarsha/smoke-udn-l2-036KQV/logs/

rakeshcn (dismissed):
Further to the discussion - Method to select container registry in bootstrap.py
**Short term plan:**
_Sanity and interop fix:_ 
- Sanity PR fix : https://github.com/red-hat-storage/cephci/pull/6156
- Verify the interop manually via CLI using this branch to confirm no-breakage
- Based on the result fix / review and land the PR

rakeshcn:
lgtm

PR #6162, opened 2026-09-15
[OneCloud/OSD]: Attach extra disks on OneCloud VMs at cluster create as per global conf.

mananxkhanna:
## **Summary**

OneCloud images no longer ship additional data disks. OSD and other device-backed tests need extra volumes on the guest. This restores extras in the OneCloud provider at `POST /clusters`, using the same cluster-conf knobs as IBM/OSP: per-node `no-of-volumes` and `disk-size`. Create fails if the deployed VM does not match what was posted.

Skip is role-only: `win-iscsi-clients` even if that node lists volumes. There is no Linux/Windows OS check. Typical cephci clusters use a RHEL image, so those VMs are Linux in practice.

## **Why this is a framework change**

This is an infra image regression, not a test-specific disk campaign. Every OneCloud cephci cluster is affected. 

Count and size limits are left to the OneCloud API. cephci does not clamp or reject a count/size that the API accepts.

## **Conf rules**

| Conf | OneCloud create |
| --- | --- |
| `no-of-volumes` missing or `0` | No extra disks (`disks` omitted on POST) |
| `no-of-volumes` set, `disk-size` missing | `NodeError`, create aborts |
| `disk-size` set, `no-of-volumes` missing/`0` | No extra disks (`disk-size` ignored) |
| Both set | That many extras, each `{"size_gb": <disk-size>}` |
| Role `win-iscsi-clients` | Always no extras |

Example: `conf/tentacle/nfs/1admin-7node-3client.yaml` — node1–3 `no-of-volumes: 6`, `disk-size: 15`; node4–7 clients have neither key, so no extras.

Nodes in the same cluster can differ (some extras, some none, different counts/sizes). That is per `node` entry, not a cluster-wide constant.

## **What create does**

1. For each node, build extras from that node’s conf (rules above).
2. `POST /clusters` `virtualMachines[].disks` only when extras are non-empty.
3. After VMs are `on` with an IP, `GET /vm/{vmid}` using the list row’s `vmid` (the key the API returns).
4. Compare non-zero `resources.disk[1:]` to what was posted for that VM. Index 0 is the OS disk. Zeros are empty slots (`[100, 0, 0]` → no extras).
5. Mismatch or missing `vmid` → `NodeError`, suite never starts.
6. `CephVMNodeOneCloud.no_of_volumes` is the extra count from that GET. `volumes()` stays `[]`: the API returns sizes, not `/dev/sdX`.

`Host.add` is unchanged: dedicated `[client]` is still excluded from `ceph orch`. Extra disks on dedicated clients are unused by `all-available-devices` unless that node is an orch host. Colocated `osd+client` nodes are orch hosts and pick up the new disks.

On create `NodeError`, OneCloud VMs are left in place (same as other post-POST create failures). Cleanup is existing `--cleanup` / cluster delete, not this path.

## **API contract (live)**

`GET /vm/{id}` returns `{ "status": "success", "data": [ { VM } ] }`. `vm_get_details()` takes `data[0]`. Confirmed on `GET /api/v3/vm/119278`.

List `GET /vm` uses the same VM object (`data` is an array of many nodes). New images without this POST look like `[100, 0, 0]`. With extras posted, extras appear as non-zero `disk[1:]` (for example NFS 7-node conf → `[100, 15, 15, 15, 15, 15, 15]` on nodes that requested 6×15).

## **Files**

- `compute/onecloud.py` — conf → POST payload, skip `win-iscsi-clients`, GET `/vm/{vmid}` check, `no_of_volumes`
- `ceph/utils.py` — attach `disks` on create, call verify after poll

## **Out of scope**

- Changing orch `all-available-devices` or `Host.add`
- Extra disks on `win-iscsi-clients`
- Auto-delete of VMs when verify fails
- cephci-side min/max size or count caps

## **Test plan**

- [x] OneCloud create with tentacle NFS conf (`1admin-7node-3client.yaml`): `Done creating OneCloud nodes` (`ci-addisk-XR441F`)
- [ ] `GET /vm/{vmid}` on a node with `no-of-volumes: 6` / `disk-size: 15`: `[100, 15, 15, 15, 15, 15, 15]`
- [x] Nodes without those keys: extras empty on GET
- [x] OSD apply: extra devices consumed on orch hosts only
- [x] Dedicated clients: disks only if conf lists volumes; still not in orch inventory
- [ ] `win-iscsi-clients`: no `disks` on POST; extras empty on GET
- [ ] Create abort if GET extras ≠ POST (suite does not start)

## **Example**

**If invalid/not supported disk size [other than OneCloud threshold] is passed**
```
2026-09-16 09:23:32,073 - cephci - run:863 - ERROR - OneCloud deploy failed: 400 {"errors":[{"msg":"Each additional disk must be between 10-50 GB","path":"virtualMachines[0].disks[0].size_gb"},**{"msg":"Each additional disk must be between 10-50 GB",**"path":"virtualMachines[0].disks[1].size_gb"}]}. Verify image_id, project_id, site, groupID, VLAN in osp-cred or inventory.
2026-09-16 09:23:32,167 - cephci - run:1427 - INFO - final rc of test run 1

```
**If invalid/not supported disk count [other than OneCloud threshold] is passed**

```
2026-09-16 09:38:28,934 - cephci - run:863 - ERROR - OneCloud deploy failed: 400 {"errors":[{"msg":"1-8 disks available per deployed Virtual Machine","path":"virtualMachines[1].disks"}]}. Verify image_id, project_id, site, groupID, VLAN in osp-cred or inventory.
2026-09-16 09:38:29,035 - cephci - run:1427 - INFO - final rc of test run 1
```

manankhanna@Manans-MacBook-Pro ~ % cat /tmp/onecloud-extra-disk-experiment.yaml 
```
# node1–5: extras, each a different count/size
# node6: no extras
# node7: extras again, different from 1–5
globals:
  - ceph-cluster:
      name: ceph
      node1:
        role:
        no-of-volumes: 2
        disk-size: 20
      node2:
        role:
        no-of-volumes: 4
        disk-size: 25
      node3:
        role:
        no-of-volumes: 6
        disk-size: 15
      node4:
        role:
          - client
        no-of-volumes: 3
        disk-size: 30
      node5:
        role:
          - client
        no-of-volumes: 1
        disk-size: 40
      node6:
        role:
          - client
      node7:
        role:
          - client
        no-of-volumes: 5
        disk-size: 10
```
image

image

image

image

image

image

image

```
[ceph: root@ci-addisk-9lcit7-node1 /]# ceph osd tree
ID  CLASS  WEIGHT   TYPE NAME                        STATUS  REWEIGHT  PRI-AFF
-1         0.22409  root default                                              
-3         0.03897      host ci-addisk-9lcit7-node1                           
 0    hdd  0.01949          osd.0                        up   1.00000  1.00000
 3    hdd  0.01949          osd.3                        up   1.00000  1.00000
-5         0.09760      host ci-addisk-9lcit7-node2                           
 1    hdd  0.02440          osd.1                        up   1.00000  1.00000
 4    hdd  0.02440          osd.4                        up   1.00000  1.00000
 6    hdd  0.02440          osd.6                        up   1.00000  1.00000
 9    hdd  0.02440          osd.9                        up   1.00000  1.00000
-7         0.08752      host ci-addisk-9lcit7-node3                           
 2    hdd  0.01459          osd.2                        up   1.00000  1.00000
 5    hdd  0.01459          osd.5                        up   1.00000  1.00000
 7    hdd  0.01459          osd.7                        up   1.00000  1.00000
 8    hdd  0.01459          osd.8                        up   1.00000  1.00000
10    hdd  0.01459          osd.10                       up   1.00000  1.00000
11    hdd  0.01459          osd.11                       up   1.00000  1.00000
```


compute/onecloud.py
```diff
@@ -38,6 +38,113 @@
...
 }
 
+# Extra data disks at POST /clusters (not the image OS disk).
+# Same layout as the old OneCloud images (100GB OS + 6x30GB). API allows 1–8 disks, 10–50 GB each.
+# Linux clients (including dedicated and osd+client) get disks. Windows iSCSI clients are skipped.
+ONECLOUD_EXTRA_DISK_SIZES_GB = (30, 30, 30, 30, 30, 30)
+ONECLOUD_DISK_SIZE_GB_MIN = 10
+ONECLOUD_DISK_SIZE_GB_MAX = 50
+ONECLOUD_DISK_COUNT_MAX = 8
+ONECLOUD_SKIP_EXTRA_DISK_ROLES = ("win-iscsi-clients",)
```
psathyan: This should be driven through the global conf and not be hardcode

psathyan (commented):
Thanks for raising a pull request to address the new changes.

Please note the number of disks and sizes must come from the global conf file. For example: https://github.com/red-hat-storage/cephci/blob/main/conf/tentacle/nfs/1admin-12node-4client.yaml#L16-L17

We need to stitch the ability like we do in other compute drivers.

compute/onecloud.py
mananxkhanna: Addressed.

aravindrrh (approved):
LGTM

AmarnatReddy (approved):

PR #6164, opened 2026-09-16
fix: update RGW upgrade 9x sanity suites to bootstrap from 9.1 released build

vidya-lokesh:
The two RGW upgrade sanity suites were bootstrapping from RHCS 9.0 (release: stable) before upgrading to the latest 9.x build. 

Changes
Updated bootstrap args in both suites:

rhcs-version: 9.0 → 9.1
release: stable → "released"
Removed hardcoded registry-url: registry.redhat.io (picked up from credential file)

Suites Updated
suites/tentacle/rgw/tier-2_rgw_upgrade_9xsanity_to_9xlatest.yaml
suites/tentacle/rgw/tier-2_rgw_ms_upgrade_9xsanity_to_9xlatest.yaml


yuva-gayam:
Hi @vidya-lokesh , the main intent of these sanity-to-latest upgrade suites in the sanity pipeline is to check whether build-to-build upgrades are passing or not. Please do not change the release value.

Although we currently have an automation gap where it should be picking up the same-version build (e.g., 9.2 in the 9.2 pipeline) instead of 9.0, we still have a gap in that part.

For now, Nack on the PR.

rakeshcn:
The PR mixes two concerns:
Registry bootstrap fix (remove registry-url) — fine
Baseline version change (9.0 stable → 9.1 released) — changes scope and duplicates existing upgrade suites

vidya-lokesh:
Hi @yuva-gayam 

Understood the intent of the suite — these suites should dynamically pick up the same-version stable build as the baseline to check if build to build upgrades are passing (e.g., 9.1 stable when running against a 9.1 nightly build, 9.2 stable when running against a 9.2 nightly build). Clearly understood the automation gap and will discuss the same as well.

This change is temporary and the sole reason for bumping from 9.0 to 9.1 is that 9.0 RPM sync has issues, causing bootstrap failures before the upgrade even begins. I have changed it to 9.1 stable for now.The release: stable and registry-url are left untouched to preserve the build-to-build upgrade intent of the suite.

Once the automation gap is fixed to dynamically resolve the baseline version per nightly build, this hardcoded value can be removed entirely.

Thank you

yuva-gayam (approved):

mkasturi18 (approved):
@vidya-lokesh thank you for details on RPM sync issues.

Approving the PR, as its a temporary resolution until we have provision for build to build upgrade.

PR #3028, opened 2023-10-16
Addition of test suite for EC2+2@4

pdhiran:
1. Added new module to test EC pool for 2+2 config on a 4 node cluster.
2. Modified other test suites for BM pipeline

Pass logs for EC 2+2 suite on 4 node BM cluster: http://magna002.ceph.redhat.com/cephci-jenkins/cephci-run-QRQF69

PR #6166, opened 2026-09-17
cephfs: replace blind sleeps with polling in regression_suite_2

Harishacharya-redhat:
## Problem

`regression_suite_2.yaml` used fixed sleeps in cleanup and polling-sensitive tests, causing flakes or unnecessary wait time.

## Solution

Replace blind sleeps with state-based polling in cleanup, snap-schedule, CG quiesce, and client-eviction paths (including orch MDS list empty and snap count via `ls`).

## Changes

- CephFS regression suite 2 test modules and related helpers

## Test plan

- [ ] `suites/tentacle/cephfs/regression_suite_2.yaml` on `tentacle-test-executor` or cephfs regression CI


tests/cephfs/cephfs_clean_up.py
```diff
@@ -108,11 +130,14 @@ def run(ceph_cluster, **kw):
             log.error("Cluster health is not OK even after waiting for 20 mins.")
             return 1
         out, _ = clients[0].exec_command(sudo=True, cmd="ceph health detail")
-        log.info("Ceph health details After 3 min sleep:\n%s", out)
-        time.sleep(180)  # Waiting to remove all stale entries for cephfs
+        log.info("Ceph health details before MDS cleanup wait:\n%s", out)
+        log.info("Waiting for stale MDS daemons to stop after filesystem removal")
+        if not fs_util.wait_for_mds_process(
+            clients[0], process_name="", ispresent=False, timeout=180, interval=10
```
Manimaran-MM: **Severity: Medium**

`FsUtils.wait_for_mds_process(..., process_name="", ispresent=False)` cannot succeed when all MDS daemons are gone: with an empty `process_name`, `process_name not in mds["daemon_id"]` is never true (`""` is a substring of every daemon id), and an empty `ceph orch ps` list never hits a success path either. The call therefore always runs until `timeout` and returns `False`, so the warning below fires even when no MDS remain.

This PR also drops the previous fixed `time.sleep(180)` before that wait, so stale-MDS teardown may get less wall time than before (~180s vs ~360s).

**Fix:** poll for an empty MDS list (or a non-empty `process_name` that matches your daemons) instead of relying on `process_name=""`:

```python
end_time = time.time() + 180
while time.time() < end_time:
    out, _ = clients[0].exec_command(
        sudo=True,
        cmd="ceph orch ps --daemon_type=mds --format json",
        check_ec=False,
    )
    if not json.loads(out or "[]"):
        log.info("No MDS daemons reported by ceph orch ps")
        break
    time.sleep(10)
else:
    log.warning("MDS daemons still present after 180s; continuing cleanup")
```

tests/cephfs/snapshot_clone/cephfs_snap_utils.py
```diff
@@ -541,6 +541,23 @@ def add_snap_sched_data(self, client, io_path, run_time):
...
             time.sleep(30)
 
+    def wait_for_scheduled_snaps(
+        self, client, path, min_count=4, timeout=300, interval=15
+    ):
+        """Poll until scheduled snapshots appear under path/.snap/."""
+        end_time = time.time() + timeout
+        while time.time() < end_time:
+            out, _ = client.exec_command(sudo=True, cmd=f"ls -lrt {path}.snap/ | wc -l")
+            snap_count = int(out.strip())
```
Manimaran-MM, line 558: **Severity: Medium**

`wait_for_scheduled_snaps` uses `min_count=4` on `ls -lrt ... | wc -l`, matching `verify_snap_schedule`, but `wc -l` counts the `total` line from `ls -lrt` as well as snapshot entries. That is fine if the suite always expects the same threshold, yet a schedule that created only three snaps could still show `wc -l >= 4` and pass the poll while `verify_snap_schedule` would still fail—or the poll could time out while snaps are still accumulating.

**Fix:** align the poll with what you assert later (e.g. reuse the same helper as `verify_snap_schedule`, or count only `.snap` entries):

```python
out, _ = client.exec_command(sudo=True, cmd=f"ls -1 {path}.snap/ | wc -l")
snap_count = int(out.strip())
if snap_count >= min_count:
    return
```

Use `shlex.quote(path)` in the shell cmd if `path` is ever not a fixed mount prefix.

Manimaran-MM (commented):
Cursor review notes (Jenkins #3028 `consoleText` was not readable from this environment — login required — so pass/fail triage is from the PR description only).

The runtime table lists **66/69** tests passed with **3 failures** while Jenkins is marked **SUCCESS**. Please point to the three failing TC names in the suite log (or map them in-code via `KNOWN_ISSUES` / IBMCEPH tickets) so regressions from shorter polling are not masked by a green executor job.

Otherwise the sleep→poll direction in `client_evict`, CG quiesce helpers, and snap-schedule paths looks reasonable; see inline comments on MDS cleanup polling.

tests/cephfs/cephfs_clean_up.py
Harishacharya-redhat: Addressed in latest commit:
- Added `wait_for_no_mds_daemons()` that polls `ceph orch ps --daemon_type=mds --format json` until the list is empty (180s timeout), instead of `wait_for_mds_process(..., process_name="", ispresent=False)`.
Thanks for catching the empty-list edge case.

tests/cephfs/snapshot_clone/cephfs_snap_utils.py
Harishacharya-redhat: Addressed: `wait_for_scheduled_snaps` now uses `ls -1` on `{path}/.snap` with `shlex.quote`, and counts entries matching `scheduled` (same as `validate_snap_schedule`), avoiding the `ls -lrt | wc -l` total-line issue.

Harishacharya-redhat:
Re: **66/69 vs Jenkins SUCCESS on #3028**

I parsed the full `consoleText` for [tentacle-test-executor #3028](https://149.81.216.83/job/tentacle-test-executor/3028/consoleText): both suite invocations log `final rc of test run 0` and the job ends `Finished: SUCCESS`. I did not find any failed test cases in that executor log.

The **580 min / 66 passed / 3 failed** row is from the **shared CI test-summary dashboard** (screenshot in the PR discussion), which is a different reporting path than the personal executor job. The **531 min** number is Jenkins wall time for #3028 on branch `fix/regression-suite-2-polling`.

For per-test status on the validation run, see the COS suite index:
https://10.8.128.2/cephci-jenkins/ibm-cos/IBM/9.1/rhel-10/executor/20.2.1-405/cephfs/3028/logs/regression_suite_2/index.html

I will update the PR description to separate “shared summary baseline” from “executor #3028” so the SUCCESS vs 3-failed wording is not conflated. If you have the three TC names from the shared summary for that 580 min row, I can map them to KNOWN_ISSUES / open bugs.

Manimaran-MM (approved):

PR #6167, opened 2026-09-17
fix: remove Agent Auth Response Code Test from tier2-cephadm-agent suite

LahariMucchala5:
## Summary

Remove `Agent Auth Response Code Test` from the tier2-cephadm-agent
sanity suite as the underlying bug (IBMCEPH-16022) is currently unfixed.

## Problem

The test verifies that the cephadm agent detects auth rejection from MGR
when the keyring is invalid. However, MGR currently returns HTTP 200 even
on auth failure — the agent sees 200, assumes success, and continues
running normally. As a result, this test always fails with:

  BUG UNFIXED: Agent shows 'running' despite rejected keyring.

## Confirmation

Confirmed with @k.amarnath.reddy (ticket owner) that this test is not
required in sanity and can be removed until the upstream bug is fixed.

## Why it is safe to remove

- Self-contained: has a finally block that fully restores the keyring
  and restarts the agent regardless of pass/fail — no other test depends
  on it
- Listed as an additional test beyond the core test plan scope in the
  suite header (not part of TC-KI-* or any mandatory test plan ID)
- Only present in the sanity job — no regression or other job is affected

## Follow-up

The test can be re-added once IBMCEPH-16022 is resolved and MGR returns
401/403 on auth failure instead of HTTP 200.

Ref: https://ibm-ceph.atlassian.net/browse/IBMCEPH-16022?issueKey=IBMCEPH-16022&subProduct=jira-software


obannak (commented):
LGTM

rakeshcn:
Suggestions:

1. **Let the Test module itself is not touched.** `cephadm_agent/test_agent_defects.py` and the `defect_auth_response` implementation remain in the codebase — the test can be re-enabled by reverting this one block once IBMCEPH-16022 is fixed. Comment out the test section with references to bug . Don't delete the test code, just stop scheduling it.

2. **The suite header comment at line 37** still lists `defect_auth_response` in the "Additional tests" comment block:
   ```yaml
   # defect_zero_division, defect_cache_leak,
   # defect_auth_response (disbaled: IBMCEPH-16022), defect_silent_listener,  ← Update this line
   ```

rakeshcn (approved):

